PR #1, opened 2015-08-09
Check ListOfType results is an instance of ArrayAccess instead of ArrayObject

dmongeau:
To allow the use of a class implementing the ArrayAccess interface.

### DIFF
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -285,7 +285,7 @@ class Executor
     }
 
     /**
-     * Implements the logic to compute the key of a given field�s entry
+     * Implements the logic to compute the key of a given fields entry
      */
     private static function getFieldEntryKey(Field $node)
     {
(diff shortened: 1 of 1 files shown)


src/Executor/Executor.php
```diff
@@ -430,7 +430,7 @@ private static function completeField(ExecutionContext $exeContext, Type $fieldT
         if ($fieldType instanceof ListOfType) {
             $itemType = $fieldType->getWrappedType();
             Utils::invariant(
-                is_array($result) || $result instanceof \ArrayObject,
+                is_array($result) || $result instanceof \ArrayAccess,
```
vladar: Technically, we need to check for `Traversable` interface here, not `ArrayAccess`, because all we do later is iterating `$result` in foreach loop.

Will merge this pull request if you change it to `Traversable`.

dmongeau:
Yes, updated!

PR #10, opened 2015-11-02
Added possibility to throw a custom error from the resolver

dmongeau:
Currently it is not possible to use custom error class that extends `GraphQL\Error` to control the format of errors. The GraphQL specifications allow errors to have extra fields has stated here https://facebook.github.io/graphql/#sec-Errors

> GraphQL servers may provide additional entries to error as they choose to produce more helpful or machine‐readable errors, however future versions of the spec may describe additional entries to errors.

This pull request is a proposition that allow custom error to be thrown from the resolve method and use custom format. By this, I mean throwing a class that extends `GraphQL\Error` and modify the `formatError` method so it returns extra fields.

There is 2 modifications to allow this.

If you throw a custom error from the resolve method, it will go through the `Error::createLocatedError` method and it will be converted to a `GraphQL\Error`. The first modification is to reuse the same error class as the original (instead of using only `GraphQL\Error`) when creating the located error.

The second modification is to allow ExecutionResult to use the `formatError` method of the current `$error` class instead of always using `GraphQL\Error::formatError`

This feature would be really helpful to implements custom validation and error reporting.

### DIFF
--- a/src/Error.php
+++ b/src/Error.php
@@ -51,7 +51,7 @@ class Error extends \Exception
             $previous = null;
         }
         
-        $errorClass = $error instanceof Error ? get_class($error):'Error';
+        $errorClass = $error instanceof Error ? get_class($error):'\GraphQL\Error';
 
         return new $errorClass($message, $nodes, $previous);
     }
--- a/src/Error.php
+++ b/src/Error.php
@@ -50,8 +50,10 @@ class Error extends \Exception
             $message = (string) $error;
             $previous = null;
         }
+        
+        $errorClass = $error instanceof Error ? get_class($error):'Error';
 
-        return new Error($message, $nodes, $previous);
+        return new $errorClass($message, $nodes, $previous);
     }
 
     /**
--- a/src/Executor/ExecutionResult.php
+++ b/src/Executor/ExecutionResult.php
@@ -30,7 +30,11 @@ class ExecutionResult
         $result = ['data' => $this->data];
 
         if (!empty($this->errors)) {
-            $result['errors'] = array_map(['GraphQL\Error', 'formatError'], $this->errors);
+            $result['errors'] = array_map(function($e)
+            {
+                $errorClass = $e instanceof Error ? get_class($e):'\GraphQL\Error';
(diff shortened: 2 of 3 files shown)


vladar:
Hey, thanks for contributing!

However, I think it is possible to use custom formatting of output without code changes. Every `GraphQL\Error` references previous exception (say the one thrown in your `resolve` method).

And GraphQL Executor returns instance of ExecutionResult, which you can format as you wish in your code.

Here is an example of how this can be achieved:

```php
function myFormatError(GraphQL\Error $error) {
    $formatted = GraphQL\Error::formatError($error);
    if ($error->getPrevious() instanceof MyResolverException) {
        $formatted += $error->getPrevious()->myCustomErrorFields;
    }
    return $formatted;
}

function myGraphQLExecute(GraphQL\Schema $schema, $requestString, $rootValue = null, $variableValues = null, $operationName = null) {
    try {
        $source = new Source($requestString ?: '', 'GraphQL request');
        $documentAST = Parser::parse($source);
        $validationErrors = DocumentValidator::validate($schema, $documentAST);

        if (!empty($validationErrors)) {
            return ['errors' => array_map('myFormatError', $validationErrors)];
        } else {
            $result = Executor::execute($schema, $documentAST, $rootValue, $variableValues, $operationName);
            return ['data' => $result->data, 'errors' => array_map('myFormatError', $result->errors)]
        }
    } catch (Error $e) {
        return ['errors' => [myFormatError($e)]];
    }
}
```

Please let me know if that helped.

vladar:
Actually I just refactored library facade to simplify custom formatting. From now on you can use:

```php
$executionResult = GraphQL\GraphQL::executeAndReturnResult($query);

$data = [
    'data' => $executionResult->data,
    'errors' => array_map('myFormatError', $executionResult->errors)
];

```

dmongeau:
That was my problem, I didn't see how I could modify the output without overwriting one method. Everything is so encapsulated in GraphQL. I knew I could use the getPrevious method but there was no way "to reach" it.

Your solution works better, thanks a lot!

Issue #9, opened 2015-10-28
subscription implementation


arec:
Do you have any plan on implementing this:

https://github.com/graphql/graphql-js/pull/189
http://graphql.org/blog/subscriptions-in-graphql-and-relay/

arec:
Sorry, wrong place

vladar:
PHP used to be bad choice to handle subscriptions. Now things change slowly with some of new libs like ReactPHP. But most of the real-world projects still use node or other tools for websockets / comet.

So the solution for subscriptions in `graphql-php` will probably be different than in graphql-js. I will evaluate possible approaches, but probably not until next month.

smolinari:
If you ask me (and a little off topic), GraphQL with straight PHP is going to be a difficult mix from the start, as seen from the performance concerns already being found with this new project. I feel it is because of the sequential/ share nothing nature of PHP. There is a reason why Facebook created HHVM with Hack's async() function. It's to run requests to data services in parallel, which is also a need to allow GraphQL to perform well IMHO. I believe it is also why Facebook has turned to the NodeJS crowd for the server side. Why haven't they released a Hack version of a GraphQL type system? 

Hmmm....

Scott

vladar:
@smolinari Yeah, honestly I expected to see PHP / Hack version of GraphQL, but instead they released js version. So had to build this library to integrate with existing PHP project.

As for performance concerns - most of them equally apply to ORMs or custom solutions in PHP. Actually such library has some advantage over custom per-project solutions in that it could be re-implemented as PHP module (as drop-in replacement). 

And I think it is a direction to follow eventually. But I suspect that even pure PHP implementation may perform reasonably well if you replace recursive Executor with iterative one (as PHP is slow in handling deep recursion). So that is probably the next step.

As for async - if you're stuck with plain PHP it doesn't matter if you use GraphQL or other tools - you don't have async anyway %)

Targeting for HHVM only is not an option at this point, but I would like to have a version that can handle async features of HHVM for those who can afford to use it.

smolinari:
> So had to build this library to integrate with existing PHP project.

And thank you for doing so! :+1: 

> As for performance concerns - most of them equally apply to ORMs or custom solutions in PHP.

[...]

PR #17, opened 2015-12-20
Add subscriptionType

andytruong:
I was trying this lib with latest graphiql, found it's missing this type. I am not sure what it is, but no more error if I add it.

### DIFF
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -20,15 +20,18 @@ class Schema
 
     protected $mutationSchema;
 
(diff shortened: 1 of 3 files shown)


ooflorent:
Subscriptions are not implemented yet in the library.
See #9

vladar:
As @ooflorent mentioned subscriptions are not implemented in this library yet, but I still merged your PR to enable compatibility with Relay / GraphQL-js introspection tools.

Thanks for submitting!

PR #28, opened 2016-03-11
setDefaultResolveFn should be static

mcg-web:
To ease override of $defaultResolveFn the setter should be `static`


mcg-web:
Thanks, could you bump this version please?

vladar:
Done. And thanks for fixing this!

PR #33, opened 2016-04-19
Optimized implementation to interfaces loading

mcg-web:
* Optimized implementation to interfaces using lazy loader.
* Union types now accepting callback

### DIFF
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -33,6 +33,8 @@ class Schema
         $this->mutationSchema = $mutationSchema;
         $this->subscriptionSchema = $subscriptionSchema;
 
+        InterfaceType::loadImplementationToInterfaces();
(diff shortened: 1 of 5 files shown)


vladar:
I am ready yo merge this, but just curious - what is your use case that requires lazy initialization of interface implementations?

mcg-web:
The use case is simple let say we have this:

```
interface A {}
type B implement A {}
```

```php
new ObjectType([
      "name" => "B",
      "fields" => [/*...*/],

]);
```

vladar:
Yeah, looks like you are right. I will also rework test case for interface definition via Closure to check for this.

mcg-web:
Yes the test will help for no BC on later modifications :+1:

PR #37, opened 2016-04-25
Add ExecutionResult $extensions

mcg-web:
Add ExecutionResult $extensions to follow https://facebook.github.io/graphql/#sec-Response-Format

### DIFF
--- a/src/Executor/ExecutionResult.php
+++ b/src/Executor/ExecutionResult.php
@@ -14,15 +14,22 @@ class ExecutionResult
      * @var Error[]
      */
     public $errors;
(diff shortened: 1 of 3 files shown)


vladar:
What is the point of ExecutionResult extensions? They don't seem to be referenced anywhere in the codebase?

mcg-web:
The specs says that `extensions`:
> entry is reserved for implementors to extend the protocol however they see fit, and hence there are no additional restrictions on its contents...

In my use case it help me send some warnings without breaking graphql protocol.

PR #39, opened 2016-05-18
Fix `ResolveInfo::getFieldSelection()` when using multiple fragments

mcg-web:
When using multiple fragments on the same field, `ResolveInfo::getFieldSelection()` was taking account only of the selection on the first fragment. Here is an example:

```graphql
query UseFragment {
  human(id: "1003") {
    name
    ...fa
    ...fb
  }
}

fragment fa on Character {
  friends {
    id
  }
}
fragment fb on Character {
  friends {
    name
  }
}
```

results with after applying select fields projection:

```json
{
    "data": {
        "human": {
            "name": "Leia Organa",
            "friends": [
                {
                    "id": "1000",
                    "name": null
                },
                {
                    "id": "1002",
                    "name": null
                },
                {
                    "id": "2000",
                    "name": null
                },
                {
                    "id": "2001",
                    "name": null
                }
            ]
        }
    }
}
```

While waiting for:

```json
{
    "data": {
        "human": {
            "name": "Leia Organa",
            "friends": [
                {
                    "id": "1000",
                    "name": "Luke Skywalker"
                },
                {
                    "id": "1002",
                    "name": "Han Solo"
                },
                {
                    "id": "2000",
                    "name": "C-3PO"
                },
                {
                    "id": "2001",
                    "name": "R2-D2"
                }
            ]
        }
    }
}
```

### DIFF
--- a/tests/StarWarsSchema.php
+++ b/tests/StarWarsSchema.php
@@ -16,6 +16,7 @@ use GraphQL\Type\Definition\EnumType;
 use GraphQL\Type\Definition\InterfaceType;
 use GraphQL\Type\Definition\NonNull;
 use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\ResolveInfo;
 use GraphQL\Type\Definition\Type;
 
 /**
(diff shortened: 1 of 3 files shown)


tests/StarWarsSchema.php
```diff
@@ -161,8 +162,19 @@ public static function build()
...
+                        $fieldSelection['id'] = true;
+
+                        $friends = array_map(
+                            function($friend) use ($fieldSelection) {
+                                return array_intersect_key($friend, $fieldSelection);
+                            },
+                            StarWarsData::getFriends($human)
+                        );
+
+
```
ooflorent: Extra line 😁

PR #45, opened 2016-07-14
Change Lexer::readString to decode escaped characters to their litera…

konforce:
Given the partial query `foo(message: "new\nline")`, I would expect the following literal if I were to print `$args['message']`:

> new
> line

Instead, I get:

> new\nline

because the Lexer fails to decode the literal, and instead repeats it as-is.

Furthermore, the partial query `foo(message: "new\\nline")`, also results in `new\nline` creating an ambiguous situation where it's impossible to know what the source argument was.

However, `\"` and `\\` do correctly get transformed into `"` and `\` respectively.

This pull request corrects that behavior for `\/ \b \f \n \r \t` to make it work as `\"` and `\\` already are working. (PHP does not support `"\b"`, hence the `chr(8)`.) Related tests are updated.

### DIFF
--- a/src/Language/Lexer.php
+++ b/src/Language/Lexer.php
@@ -229,13 +229,13 @@ class Lexer
                 $code = Utils::charCodeAt($body, $position);
                 switch ($code) {
                     case 34: $value .= '"'; break;
-                    case 47: $value .= '\/'; break;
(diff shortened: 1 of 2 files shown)


vladar:
Sorry for long delay on that. You are correct with this change, so it is merged.

PR #51, opened 2016-09-15
Fix fatal error when GraphiQL tries to run introspection query

geshido:
Solves fatal error:

PHP Fatal error:  Cannot use object of type GraphQL\\Type\\Definition\\Directive as array in /var/www/html/sites/all/modules/med_stdlib/lib/vendor/webonyx/graphql-php/src/Type/Introspection.php on line 295

### DIFF
--- a/src/Type/Introspection.php
+++ b/src/Type/Introspection.php
@@ -282,25 +282,25 @@ EOD;
                         'deprecationReason' => 'Use `locations`.',
                         'type' => Type::nonNull(Type::boolean()),
                         'resolve' => function($d) {
-                            return in_array(Directive::$directiveLocations['QUERY'], $d['locations']) ||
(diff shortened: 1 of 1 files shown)


mcg-web:
@geshido I don't have this error when running introspection query with graphiQL, what version are you using?

geshido:
@mcg-web Oh, I see. I am using ChromeiQL extention and it runs outdated version of GraphiQL I suppose

vladar:
@geshido Try https://chrome.google.com/webstore/detail/graphiql-feen/mcbfdonlkfpbfdpimkjilhdneikhfklp - works nice for me and has more features than ChromeiQL

vladar:
The error you mentioned and fix are valid - introspection query with deprecated fields should still run correctly

Issue #60, opened 2016-10-21
Horizontal eval for lazy loading same-type objects


rudiedirkx:
When I first learned about GraphQL there was a performance chapter. If every relation loads every single relation, it requires many queries to load a few levels of data, even if the actual data is only on the last level. The solution was to lazy load objects: first collect ids (horizontal), and then load all objects at once, and then evaluate the next level.

Just in case. This query:

```
query {
    user {
        reservations {
            resource {
                sport {
                    name
                }
            }
        }
    }
}
```

would take 20 queries to find 10 reservations > 10x 1 resource > 10x 1 sport. If it would eval horizontally, it would do 2 queries: load 10 resources, then load 10 sports. (Maybe not 2 levels, but even 1 could be a big difference.)

Is that a thing? It seems like the query is evaled vertically first, so it loads the entire 1 reservation (1 resource + 1 sport), and then the entire 1 next (1 resource + 1 sport) etc.

rudiedirkx:
Some kind of preview would also be great: after parsing the query, without any eval, ask it which types are being requested. Or on any level, ask which types are on the next level(s). Any idea?

vladar:
Yes this is a well known N+1 problem. And you are correct - executor uses Depth-First algorithm vs Breadh-First (same as reference implementation). 

I posted this [comment](https://github.com/webonyx/graphql-php/issues/6#issuecomment-147172125) one day about how to workaround this in the user land.

As for solution in core - there is no such thing yet. But I do understand that we need one. The question is which approach will work best (and if I can find enough time to implement it).

One option is to use breadth-first execution as you suggest. But this is really major change and I never found enough time to experiment with it (I actually tracked some other implementations, like .NET which also had plans for breadth-first execution, but never started too).

Other option is to add rather minor modification to support Deferred (Lazy) values which will enable batching. 

Then your resolver would look like this:

```php
'resolve' => function($obj) {
    $this->someBuffer[] = $obj->someId;

    return new GraphQL\Deferred(function() use ($obj) {
        $values = DataLoader::loadAll($this->someBuffer); // must load stuff only once obviously
        return isset($values[$obj->someId]) ? $obj->someId : null;
    });
}
```
(see also [this thread](https://github.com/webonyx/graphql-php/issues/42) for related discussion)

This is actually how JS version works (Facebook created separate project called [DataLoader](https://github.com/facebook/dataloader) to help implementing this approach).

[...]

PR #61, opened 2016-10-22
+InlineFragment fields in ResolveInfo->getFieldSelection()

rudiedirkx:
Background starting at https://github.com/webonyx/graphql-php/issues/60#issuecomment-255546883


src/Type/Definition/ResolveInfo.php
```diff
@@ -130,6 +131,12 @@ private function foldSelectionSet(SelectionSet $selectionSet, $descend)
                     $fragment = $this->fragments[$spreadName];
                     $fields += $this->foldSelectionSet($fragment->selectionSet, $descend);
                 }
+            } else if ($selectionAST instanceof InlineFragment) {
+                $fragmentFields = $this->foldSelectionSet($selectionAST->selectionSet, $descend - 1);
+                $fragmentType = $selectionAST->typeCondition->name->value;
+                foreach ($fragmentFields as $name => $subFields) {
+                    $fields[$fragmentType . ':' . $name] = $subFields;
```
rudiedirkx: Not really. The type condition is important, because implementor A and B could both query field X, but have different children, because they mean something else:

```
... on A {
  some {
    type
  }
}
... on B {
  some {
    kind
  }
}
```

Would B win and overwrite A's `type`?

vladar: You are right. Need to think about it before merging

rudiedirkx: Maybe a new kind of nesting level instead of prefixing the key? Not very bc. Or returning a new, array accessible object with meta data like this? Could be bc, but might be overkill.

Also, I'm not sure this patch includes all cases: inline fragment, named fragment, inline fragment inside named fragment, etc. I've tried only 2 of many scenarios.

rudiedirkx:
Have any brilliant solutions come up? With this, or something like it, I can make very efficient getters.

vladar:
@rudiedirkx Not really. I guess good solution requires breaking change anyway.

I incline to deprecating this function and introducing new one wich will handle Inline fragments in more elegant way.

I guess I could merge this, but don't see much point since it will be deprecated anyway.

But technically it shouldn't block you. Just extract this piece of logic into your own utility function accepting `ResolveInfo` as argument and do the same.

Maybe eventually *you* will come up with brilliant solution and contribute it back %)

rudiedirkx:
Sure, I could use my own method, but I rather use your shiny new very awesome new method. I'm very curious to see this more elegant way. Sneak preview anywhere? Create issue to track progress? You can close this PR.

vladar:
@rudiedirkx See #65 to track this. Looking forward to hear your thoughts and ideas.

Issue #70, opened 2016-11-06
Introduce optional ConfigBuilder


vladar:
Currently type definitions require array with configurations. It will stay this way, yet it might be useful to introduce ConfigBuilder with chainable syntax for convenience to leverage the power of autocompletion in modern IDEs. 

Primary reason why it wasn't done originally is performance. Since all schema types had to be loaded on schema creation, such builder would add too much performance overhead. So real projects had to rely on error-prone string-based config definitions. 

But if #69 gets solved - our types will be loaded on demand and only needed types will be loaded for GQL query. Builder becomes more realistic in such case.

The envisioned syntax is something along these lines:

```php
$config = ObjectTypeConfig::create()
  ->name('TypeName')
  ->addField('field1', Type::string(), 'description', null, ArgsConfig::create()
      ->addArg('arg1', Type::boolean(), 'description', true)
      ->addArg('arg2', Type::string(), 'description', 'defaultVal')
  )
  ->addInterface($iface)
;

$type = new ObjectType($config->build());
```

Obviously such builder doesn't require any modifications to core project and is BC.

tpetry:
A ConfigBuilder like the OOP definition of [Youshido/GraphQL](https://github.com/Youshido/GraphQL) would be a nice option. Because it would allow useful encapsulation instead of one gigantic file with dozens/hundreds of configurations.

vladar:
@tpetry I do agree that it is convenient, but it should be optional. 

One of the reasons why array-based definitions were chosen as default is performance. Like it or not, but when you have hundreds of types and thousands of fields it means you get thousands of builder function calls on schema creation (likely tens of thousands). They are cheap, but when cheap is multiplied 10000x it gets expensive.

If it was other language (where schema could be created only once and re-used for every request) - it would made much more sense to prefer builder as a default. But in PHP you have to look for balance.

Now when we have some options for loading types/fields on demand (see #69) such builder makes more sense. 

As for gigantic file - you don't have to use single file with all definitions. See [example](https://github.com/webonyx/graphql-php/tree/master/examples/01-blog/Blog/Type) for type per file.

Also check out [this PR](https://github.com/webonyx/graphql-php/pull/74) for work on progress on this issue - maybe you will post your thoughts / comments on implementation.

tpetry:
> Now when we have some options for loading types/fields on demand (see #69) such builder makes more sense.

[...]

PR #74, opened 2016-11-12
Add ObjectTypeConfig and ArgConfig

mcg-web:
Introduce optional ConfigBuilder (see #70)

### DIFF
--- a/src/Type/Builder/ObjectTypeConfig.php
+++ b/src/Type/Builder/ObjectTypeConfig.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace GraphQL\Type\Builder;
+
+class ObjectTypeConfig extends Config
+{
+    public function name($name)
+    {
+        $this->addConfig('name', $name, false);
+        return $this;
+    }
+
+    public function description($description)
+    {
+        $this->addConfig('description', $description, false);
+        return $this;
+    }
+
+    public function addField($name, $type, callable $resolve = null, $description = null, ArgsConfig $args = null)
+    {
+        $field = [
+            'name' => $name,
+            'type' => $type,
+        ];
+
+        if (null !== $description) {
+            $field['description'] = $description;
+        }
+
+        if (null !== $resolve) {
+            $field['resolve'] = $resolve;
+        }
+
+        if (null !== $args) {
+            $field['args'] = $args->build();
+        }
+
+        $this->addConfig('fields', $field);
+
+        return $this;
+    }
+
+    public function addInterface($interface)
+    {
+        $this->addConfig('interfaces', $interface);
+        return $this;
+    }
+
+    public function isTypeOf(callable $isTypeOf = null)
+    {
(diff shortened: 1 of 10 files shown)


mcg-web:
@vladar can review this? if OK for you i'll add tests to cover remaining code

vladar:
Hey by the way - will be very glad to see this feature in our lib. Thanks for this work!

src/Type/Builder/ObjectTypeConfig.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public function description($description)
+    {
+        $this->addConfig('description', $description, false);
+        return $this;
+    }
+
+    public function addField($name, $type, callable $resolve = null, $description = null, ArgsConfig $args = null, callable $complexity = null, $deprecationReason = null)
+    {
```
mcg-web: ok no prob for the order of the addField Arguments, also +1 for the `addDeprecatedField.  we must be careful of reusing  fields definition in closure because same instance will be share in interface and object. I'll work on all this latter. Thanks for the review :)

vladar:
Feel free to merge it yourself when ready. 

One thing that is missing though is documentation. New docs are stored in /docs folder. Please add new file about builder when you have a chance. I guess it makes sense to explicitly mention that it is experimental API which will likely to change (docs are powered by [mkdocs](http://www.mkdocs.org/), so documentation site structure is defined in `mkdocs.yml`).

Also just a friendly request - can you avoid Traits? I know it is opinionated view, but if we can avoid them without major impact on codebase - I'd appreciate it. If not - I can still live with it.

mcg-web:
Ok i'll work on this later :+1:

enumag:
Can we get this merged? Even without documentation the builders are quite clear from the testcase and I'd like to use them.

mcg-web:
I completely forgot this PR, I will add documentation this week end so we can merge it :+1:

enumag:
@mcg-web Awesome! :+1:

enumag:
So, any progress? :-)

mcg-web:
I'm Improving the coverage to be sure to don't miss nothing... I'll submit my changes this morning.

mcg-web:
@vladar this can be merged in master or we must wait for next major release?

enumag:
Please wait for a bit. I dug a bit deeper into it and I think we might want to use a different approach. I'm writing an issue right now.

mcg-web:
OK here is the initial PR #70

vladar:
I think I can merge it before the next version release. So just keep it as is. Actually, I am not sure if we do need it at all (since many people introduce their own methods of building types or use framework-specific libs which do it for them).

Adding it to the main library may add some confusion about the right way to do things. @mcg-web what do you think about it? Will you guys use these builders for symfony bundle?

mcg-web:
No in the bundle we using the generator types so no need of this. We can close this. Less codes to maintain :D.

PR #75, opened 2016-11-12
Refactor parser

AndreasHeiberg:
*Warning*: massive brain dump here largely unproductive.

*Preface*: the lib has great compliance to the spec, but I personally find the code base could be cleaned up a bit. This is largely a superficial nudge in one direction, removing static, stop using exposing public fields, etc.

This is just a quick and dirty first pass trying to get the interfaces to be slightly nicer. Next pass would be to work on using type checking instead of relying on checking field existence.

*Problems*: it's not backward compliant in anyway. One could also argue that in the current state it's not much of an improvement. I just wanted to put a feeler out there for wether you guys wish to take the project in a direction that allows for DI, doesn't rely on statics and doesn't expose internal data structures.

Oh also 2 of the tests are still failing I will get on that.

Thoughts?

### DIFF
--- a/benchmarks/LexerBench.php
+++ b/benchmarks/LexerBench.php
@@ -30,7 +30,7 @@ class LexerBench
         $lexer = new Lexer($this->introQuery);
 
         do {
-            $token = $lexer->nextToken();
+            $token = $lexer->advance();
         } while ($token->kind !== Token::EOF);
     }
 }
--- a/benchmarks/LexerBench.php
+++ b/benchmarks/LexerBench.php
@@ -31,6 +31,6 @@ class LexerBench
 
         do {
             $token = $lexer->advance();
-        } while ($token->kind !== Token::EOF);
+        } while ($token->getKind() !== Token::EOF);
     }
 }
--- a/examples/01-blog/Blog/Type/Scalar/UrlType.php
+++ b/examples/01-blog/Blog/Type/Scalar/UrlType.php
@@ -53,11 +53,11 @@ class UrlType extends ScalarType
         // Note: throwing GraphQL\Error\Error vs \UnexpectedValueException to benefit from GraphQL
         // error location in query:
         if (!($ast instanceof StringValue)) {
-            throw new Error('Query error: Can only parse strings got: ' . $ast->kind, [$ast]);
+            throw new Error('Query error: Can only parse strings got: ' . $ast->getKind(), [$ast]);
         }
-        if (!is_string($ast->value) || !filter_var($ast->value, FILTER_VALIDATE_URL)) {
+        if (!is_string($ast->getValue()) || !filter_var($ast->getValue(), FILTER_VALIDATE_URL)) {
             throw new Error('Query error: Not a valid URL', [$ast]);
         }
-        return $ast->value;
+        return $ast->getValue();
     }
 }
--- a/src/Error/Error.php
+++ b/src/Error/Error.php
@@ -2,6 +2,8 @@
 
 namespace GraphQL\Error;
 
+use GraphQL\Language\AST\Location;
+use GraphQL\Language\AST\Node;
 use GraphQL\Language\Source;
 use GraphQL\Language\SourceLocation;
 use GraphQL\Utils;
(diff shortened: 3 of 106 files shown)


vladar:
I do agree with some of your statements and disagree with some others.

For example about using publics vs getters/setters - this is because of performance reasons primarily. PHP is not the best tool for parsers and huge queries already produce significant overhead. It is not uncommon to use public props for this. For example Doctrine ORM does the same for the most intensive cases: https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L46

As for other notes and changes - will get back to you as soon as I get a chance. Give me some time to digest it %)

AndreasHeiberg:
Yeah totally up to you :) I was just playing about, wanted to use the parser for something else and "needed" / wanted it to resemble what is was replacing slightly more.

vladar:
I reviewed your changes and will definitely cherry-pick some of them. Yet there are two main principles which I prefer to follow in this lib:

1. Backwards compatibility and deprecation strategy vs breaking. In other words - deprecated stuff stays in library long enough for all users to adapt (with optional E_USER_DEPRECATED notice) - for at least one medium release (better longer if possible).

It is not always possible, but very desireable. We only had one release with major BC break sine the inception of this lib (if my memory doesn't lie) and even then the only reason for break was that it was almost impossible to introduce needed change via deprecations.

2. Keeping surface area of library API small. It is hard to do in PHP, so we should probably document it instead (in new docs + maybe using @api dockblock as symfony used to do). 

So we do expose:
`parse() -> AST tree`
`validate() -> Error[]`
`execute() -> ExecutionResult`

+ type definitions and AST structures. (maybe there are couple other APIs which I forgot to mention now - we definitely need to document all of them).

Everything else is pretty much implementation detail and should not be relied upon by end-users. Every exposed API requires time and efforts to maintain (or irresponsibility in exposing it in the first place with the cost of breaks and errors in future).

As for DI. The trick with DI is that you only need it if you do want to expose something for extension or when you need polymorphism for your internal implementations. Same applies to `Interfaces` and other polyporphism-related tools in general.

But if you do not expose something - polyporphism-related tools often complicate things (at least when they introduce new entities like interfaces, factory classes, etc which you could avoid otherwise, hence breaking Occam's razor rule). Doing these things mechanically somehow became a rule of thumb and good practice, but I do not subscribe to this point of view %) 

For example changing `parse()` to `new Parser(new Lexer)->parse()` pretty much exposes implementation details of the Parser. Do we want to add other lexer in our implementation? No. Do we want to allow end-users define their own Lexers and are ready to maintain Lexer interface for that? No. Then why doing this?

So to summarize - I am very much in favor of polymorphism-enablement (DI, interfaces, etc) for cases where we decide to expose some functionality to end-users for extension or when we do need it for our own implementations. But not mechanically.

Obviously these are just principles. Sometimes they are hard to follow, but we try in this lib.

As for your changes - I do agree with some of them (style changes, changes of statics to objects where appropriate) - will cherry pick them when I find time to work on lib next time.

`publics` will stay because of performance and backwards compatibility. 

We can consider replacing them with getters/setters for type definitions, yet I do not see any particular advantage in doing this (more lines of code to maintain, worse read performance, bigger serialization size if we decide to allow definition serializations in the future). But even if we do - we'll add __isset, __get, __set for BC.

Anyway I am really glad to see your PR and will be happy to see other PRs as long as they follow two principles above. I'd like to keep this library open to contributions!

AndreasHeiberg:
Glad to hear back from you so swiftly. Here's my thoughts:

I don't think the public interface only extends to those 3 methods. When you do execution you might be looking at the AST in your resolve function to come up with a better strategy, so the entire AST is part of the public interface.

However I fully agree that injecting the lexer is a bit of an odd move, it was only done to make the initial refactor pass fast. This is far from how I would ideally want the code to look. I can come up with an argument for injecting in lexers though. You talk about overhead of lexing the query. Well part of this is that you have a lexer that both does Schema and Query lexing. Since these never occur in the same payload you could split them, here DI could become useful. Again this is very contrived.

But yeah fully agree with your point that you shouldn't make decisions mechanically. That being said I don't see the advantage to your static approach.

On backwards compatible changes thats very subjective obviously. I think it's very early days for graphql and your lib and I worry that your putting too many limitations on your self. But I'm very much in the build fast and break things camp :D

You've done an amazing job writing this library though, so I really shouldn't be complaining :D thanks so much for all the hard work and attention to detail. It was amazing to see that your test coverage is  extensive when doing these refactors.

vladar:
Hey, your PRs are welcome anyway! Even if they break stuff - we can negotiate %)

vladar:
Closing as I cherry-picked several commits from this PR.

PR #80, opened 2016-12-20
Nested async queries messes up the order of keys in arrays producing unwanted responses

lordthorzonus:
When batch loading multiple nested relations in an async environment the values of the arrays aren't resolved in a sync order anymore which produces php arrays whose indexes aren't ordered properly. This causes json_encode to evaluate them as associative arrays and add the keys to the json response.

For example a query:

```graphql
{
  eventParticipantsForAnEvent(eventId: 2006) {
    id
    contactInfo{
      id
      companyInfo{
        id
      }
    }
  }
}
```
Where all the relations are one -> one and batch loaded it produces following json:
```json
{
  "data": {
    "eventParticipantsForAnEvent": {
      "0": {
        "id": 89906,
        "contactInfo": {
          "id": 112218,
          "companyInfo": {
            "id": 25623
          }
        }
      },
      "1": {
        "id": 89919,
        "contactInfo": {
          "id": 112570,
          "companyInfo": {
            "id": 25696
          }
        }
      }
    }
  }
}
```
When the expected result would be: 

```json
{
  "data": {
    "eventParticipantsForAnEvent": [
      {
        "id": 89906,
        "contactInfo": {
          "id": 112218,
          "companyInfo": {
            "id": 25623
          }
        }
      },
      {
        "id": 89919,
        "contactInfo": {
          "id": 112570,
          "companyInfo": {
            "id": 25696
          }
        }
      }
    ]
  }
}
```
I'm not sure if this is the right approach to solve the problem. But at least it removes it :). If we want to add tests we probably need to add the event loop as dev dependency?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,9 @@ before_install:
 
 install:
     - composer install --dev --prefer-dist
+    - composer require react/promise:2.*
 
-script: if [ "$TRAVIS_PHP_VERSION" == "5.6" ]; then phpunit --coverage-clover build/logs/clover.xml; else phpunit; fi
+script: if [ "$TRAVIS_PHP_VERSION" == "5.6" ]; then phpunit --coverage-clover build/logs/clover.xml --group default,ReactPromise; else phpunit --group default,ReactPromise; fi
 
 after_success:
     - if [ "$TRAVIS_PHP_VERSION" == "5.6" ]; then composer require "satooshi/php-coveralls:^1.0" && travis_retry php bin/coveralls -v; fi
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -17,6 +17,12 @@
         </testsuite>
     </testsuites>
 
+    <groups>
+        <exclude>
+            <group>ReactPromise</group>
+        </exclude>
+    </groups>
+
     <filter>
         <whitelist>
             <directory suffix=".php">./src</directory>
--- a/src/Executor/Promise/Adapter/ReactPromiseAdapter.php
+++ b/src/Executor/Promise/Adapter/ReactPromiseAdapter.php
@@ -71,7 +71,16 @@ class ReactPromiseAdapter implements PromiseAdapter
         $promisesOrValues = Utils::map($promisesOrValues, function ($item) {
             return $item instanceof Promise ? $item->adoptedPromise : $item;
         });
-        $promise = \React\Promise\all($promisesOrValues);
+
+        $promise = \React\Promise\all($promisesOrValues)->then(function($values) use ($promisesOrValues) {
+            $orderedResults = [];
+
+            foreach ($promisesOrValues as $key => $value) {
+                $orderedResults[$key] = $values[$key];
+            }
+
+            return $orderedResults;
+        });
         return new Promise($promise, $this);
     }
 }
--- a/tests/Executor/Promise/ReactPromiseAdapterTest.php
+++ b/tests/Executor/Promise/ReactPromiseAdapterTest.php
@@ -19,8 +19,8 @@ class ReactPromiseAdapterTest extends \PHPUnit_Framework_TestCase
 {
     public function setUp()
     {
-        if(! class_exists(ReactPromise::class)) {
-            $this->markTestSkipped('react/promise package must be installed to run ' . self::class);
+        if(! class_exists('React\Promise\Promise')) {
+            $this->markTestSkipped('react/promise package must be installed to run GraphQL\Tests\Executor\Promise\ReactPromiseAdapterTest');
         }
     }
 
@@ -49,8 +49,8 @@ class ReactPromiseAdapterTest extends \PHPUnit_Framework_TestCase
 
         $promise = $reactAdapter->convertThenable($reactPromise);
 
-        $this->assertInstanceOf(Promise::class, $promise);
-        $this->assertInstanceOf(FulfilledPromise::class, $promise->adoptedPromise);
+        $this->assertInstanceOf('GraphQL\Executor\Promise\Promise', $promise);
+        $this->assertInstanceOf('React\Promise\FulfilledPromise', $promise->adoptedPromise);
     }
 
     public function testThen()
@@ -66,8 +66,8 @@ class ReactPromiseAdapterTest extends \PHPUnit_Framework_TestCase
         });
 
         $this->assertSame(1, $result);
-        $this->assertInstanceOf(Promise::class, $resultPromise);
-        $this->assertInstanceOf(FulfilledPromise::class, $resultPromise->adoptedPromise);
+        $this->assertInstanceOf('GraphQL\Executor\Promise\Promise', $resultPromise);
+        $this->assertInstanceOf('React\Promise\FulfilledPromise', $resultPromise->adoptedPromise);
     }
 
     public function testCreate()
@@ -77,8 +77,8 @@ class ReactPromiseAdapterTest extends \PHPUnit_Framework_TestCase
              $resolve(1);
         });
 
-        $this->assertInstanceOf(Promise::class, $resolvedPromise);
-        $this->assertInstanceOf(ReactPromise::class, $resolvedPromise->adoptedPromise);
+        $this->assertInstanceOf('GraphQL\Executor\Promise\Promise', $resolvedPromise);
+        $this->assertInstanceOf('React\Promise\Promise', $resolvedPromise->adoptedPromise);
 
         $result = null;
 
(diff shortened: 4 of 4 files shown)


mcg-web:
Hi, can you tell me what promise adapter are you using?

vladar:
Hey, thanks for reporting. Looks like it's not the order that is wrong - it's just that keys are set as strings vs int.

I don't think `ksort` is a proper fix for the problem, so need a bit more info on how to reproduce it:
1. Do you use ReactPHP or some other async environment? 
2. Also I am curious if you return "Promise for list" or "List of promises" in resolver?

As for adding event loop to tests - I wanted to leverage HHVM in Travis CI to test it, but couldn't find decent promise library for HHVM. Any hints how to integrate async stuff with Travis are welcome!

lordthorzonus:
I'm using ReactPromiseAdapter in a React event loop environment. 

The batch loader I'm using returns a promise which resolves into a value ( the relation between entities in this case is one eventParticipant has one contactInfo and contactInfo has one companyInfo).

I think the problem is the wrong order of index keys in array. For my examples I shortened the responses (the real one would contain 600 entities), Sorry for being unclear :). I tried vardumping the execution result before json_decoding and the keys are integers like they should.  

The print_r looks following where the unordering first occurs looks like this: 

```
Array
(
    [data] => Array
        (
            [eventParticipantsForAnEvent] => Array
                (
                    [0] => Array
                        (
                            [id] => 89906
                            [masterContact] => Array
                                (
                                    [id] => 112218
                                    [masterCompany] => Array
                                        (
                                            [id] => 25623
                                        )

                                )

                        )

                    [1] => Array
                        (
                            [id] => 89919
                            [masterContact] => Array
                                (
                                    [id] => 112570
                                    [masterCompany] => Array
                                        (
                                            [id] => 25696
                                        )

                                )

                        )

                    [2] => Array
                        (
                            [id] => 90531
                            [masterContact] => Array
                                (
                                    [id] => 112622
                                    [masterCompany] => Array
                                        (
                                            [id] => 25711
                                        )

                                )

                        )

                    [3] => Array
                        (
                            [id] => 90541
                            [masterContact] => Array
                                (
                                    [id] => 112869
                                    [masterCompany] => Array
                                        (
                                            [id] => 25803
                                        )

                                )

                        )

                    [4] => Array
                        (
                            [id] => 90547
                            [masterContact] => Array
                                (
                                    [id] => 113275
                                    [masterCompany] => Array
                                        (
                                            [id] => 25890
                                        )

                                )

                        )

                    [549] => Array
                        (
                            [id] => 106767
                            [masterContact] => Array
                                (
                                    [id] => 460759
                                    [masterCompany] => Array
                                        (
                                            [id] => 25890
                                        )

                                )

                        )

                )       
        ) 
)

```

As for the unit testing I think we just need to new up the event loop and run that explicitly before asserting the test results.

Please let me know if you need some more info! I can for example try to replicate this in a dummy repo with some mock data :).

vladar:
Sorry I am a bit slow today, so just to clarify...

The question is who actually creates this array with gaps? Is it produced by your resolver/promise or produced by graphql?

In the former case the result is expected. Here is a simple test which produces the same result as you describe:

```php
$a = [0 => 'a', 1 => 'b', 100 => 'z'];
echo json_encode($a);
```

If your app returns array with such keys - it should be *probably* fixed in the user-land (but we can think about forcing arrays with valid keys). But if you return array with ordered integer keys without gaps and graphql re-orders them - this is a bug.

So which is the case?

lordthorzonus:
Basically resolving promises async makes the array unordered. Here is a messy test to reproduce the problem.  We are expecting an array with of [1,2,3](keys: 0,1,2) for the test field but we are actually getting it in order [1,3,2](keys:0,2,1) as the second promise is resolved async later than the others. I think the problem happens somewhere when we are doing $promise->all(); The arrays keys are somehow preserved but the order changes as the values are resolved in different order.

```php
    public function testPromiseListsPreserveTheirArrayKeyOrder()
    {
        $loop = \React\EventLoop\Factory::create();
        Executor::setPromiseAdapter(new ReactPromiseAdapter());

        $testData = [
            new \React\Promise\Promise(
                function ($resolve) use ($loop) {
                    $loop->nextTick(
                        function () use ($resolve) {
                            $resolve(1);
                        }
                    );
                }
            ),
            new \React\Promise\Promise(
                function ($resolve) use ($loop) {
                    $loop->nextTick(
                        function () use ($resolve, $loop) {
                            \React\Promise\resolve()->then( function() use ($resolve, $loop) {
                                $loop->nextTick(function () use ($resolve) {
                                    $resolve(2);
                                });
                            });
                        }
                    );
                }
            ),
            new \React\Promise\Promise(
                function ($resolve) use ($loop) {
                    $loop->nextTick(
                        function () use ($resolve) {
                            $resolve(3);
                        }
                    );
                }
            )
        ];

        $expected =  [ 'data' => [ 'nest' => [ 'test' => [ 1, 2, 3 ] ] ] ];
        $expectedJson = json_encode($expected);


        $data = ['test' => $testData];
        $dataType = null;
        $testType = Type::nonNull(Type::listOf(Type::nonNull(Type::int())));

        $dataType = new ObjectType(
            [
                'name' => 'DataType',
                'fields' => function () use (&$testType, &$dataType, $data) {
                    return [
                        'test' => [
                            'type' => $testType
                        ],
                        'nest' => [
                            'type' => $dataType,
                            'resolve' => function () use ($data) {
                                return $data;
                            }
                        ]
                    ];
                }
            ]
        );

        $schema = new Schema(['query' => $dataType]);

        $ast = Parser::parse('{ nest { test } }');
        $result = Executor::execute($schema, $ast, $data);

        $response = null;
        $loop->run();

        $result->then(function($value) use (&$response) {
            $response = $value;
        });

        $this->assertEquals($expectedJson, json_encode($response->toArray()));
    }
```

vladar:
Thanks a lot for the test case! Will check it as soon as I have a chance.

lordthorzonus:
I investigated bit further and this could be a problem (or indented behavior) in ReactPHP itself. I opened a issue there : https://github.com/reactphp/promise/issues/74 .

If it's indented the question is should we ever expect a unordered array as a list? As it should be a list by GraphQl spec. Javascript doesn't have this problem as arrays are always in order and arrays with gaps are converted to json arrays correctly.

Thanks a thousand for help! :)

vladar:
I suspected that it could be caused by ReactPHP `all`, but you are right - we should probably force proper keys in array when expected return type is `List`. I'll create separate issue to track this.

vladar:
@lordthorzonus I suspect that we need to add `ksort` to `ReactPromiseAdapter::all` anyway. Can you change this PR accordingly (move `ksort` from Executor to ReactPromiseAdapter)? I'll merge it then.

mcg-web:
Hi @vladar, I think that `ksort` is not a reliable solution, what not implement something similar to [this](https://github.com/webonyx/graphql-php/blob/master/src/Executor/Executor.php#L396-L410) ?

vladar:
@mcg-web You are right. `ksort` will mess keys of associative arrays. I guess what we need in `ReactPromiseAdapter::all` is something like this:

```php
function all(array $promisesOrValues)
{
    $result = \React\Promise\all($promisesOrValues);
    return $result->then(function($values) use ($promisesOrValues) {
        $newResult = [];
        foreach ($promisesOrValues as $key => $value) {
            $newResult[$key] = $values[$key];
        }
        return $newResult;
    });
}
```
Is it what you mean?

mcg-web:
Yes @vladar, you got it :+1:

lordthorzonus:
@mcg-web @vladar  :+1: I was thinking the same. I'll update the PR. Do we want tests as well for this? Then we probably need to add `react/event-loop` and `react/promise` as a dev-dependency. Is it acceptable?

vladar:
@lordthorzonus I don't think we should add it as dev dependency. I don't want to create impression that ReactPHP is somwhow required for development when using this lib. It is 100% optional stuff and composer `suggest` is appropriate for it.

What we could do with reagrds to tests: 
1. Add tests for ReactPHP promise adapter, but exclude them in default `phpunit.xml.dist`
2. Add `composer require react/event-loop` to .travis.yml (similar to how we do it with coveralls) + add ReactPHP tests for Travis builds only.

If you do not want to bother with this - just add tests for ReactPHP that you find appropriate - and exclude them from default `phpunit.xml.dist`. We'll manage Travis stuff when we find time.

P.S. Glad that you are using thins async stuff! %)

lordthorzonus:
@vladar PR updated. Added tests and fixed the `all()` behaviour. 

`ReactPromiseAdapterTest` is now excluded with phpunit's group feature from the default run (and travis runs them by using the group flag `--group default,RectPromise`. It's also marked as skipped if somebody tries to run it and hasn't got the `react/promise` package installed.

vladar:
@lordthorzonus Great work. Thanks!

PR #84, opened 2017-01-13
Added compatibility with php7 error handling

OwlyCode:
Hello !

I recently stumbled upon an issue with php7 and error handling. In php7 a `\Error` class was introduced (to convert warning and fatal errors to exceptions). The problem is that `\Error` is not inheriting `\Exception` and, as a consequence, is not catched correctly.

Here is a fix for that, I checked on older php versions, it should not breaking anything (as catching on an unknown class will not generate any error : https://3v4l.org/Hb5HH ).


vladar:
Thanks for PR!

PR #88, opened 2017-02-06
Schema element descriptions can be set through comments

jane-olszewska:
## Reason for change:

GraphQL schema shorthand parsing currently doesn't allow for adding descriptions to the type system.

## Solution description:

Following the Javascript implementations, comment blocks immediately preceding a definition will be treated as a description.

Some examples:

```
# This is a description of MyType type
type FriendlyType {
  # This is a field description
  friendlyField: String
}

# This is a floating comment, and is therefore not treated as a description of UnfriendlyType

type UnfriendlyType {
  unfriendlyField: String # not a description, just an aside
}
```

### DIFF
--- a/src/Language/AST/EnumTypeDefinitionNode.php
+++ b/src/Language/AST/EnumTypeDefinitionNode.php
@@ -22,4 +22,9 @@ class EnumTypeDefinitionNode extends Node implements TypeDefinitionNode
      * @var EnumValueDefinitionNode[]
      */
     public $values;
+
+    /**
+     * @var string
+     */
+    public $description;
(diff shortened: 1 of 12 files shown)


vladar:
Awesome! Thanks.

Torsten85:
Am I missing something or do comments disappear when calling `AST::fromArray`? Calling `AST::toArray` correctly writes the comments in the returned array.

vladar:
Looks like it's the same as #201

Torsten85:
You are right! I'll keep an eye on that issue. Thank you!

Issue #90, opened 2017-02-15
buildSchema


petecoop:
Hey, is there anything similar to `buildSchema()` from `graphql-js`? Which allows you to build the schema from the GraphQL Schema language.

I've found it to be a much nicer/quicker way when building schema's in node and was hoping to find something similar in PHP.

If not would it be possible based on the parser already available in the project? I'd be happy to look into it if I could be pointed in the right direction, I'm planning on using this on a few projects.

vladar:
Unfortunately not yet. PR is welcome!

As for direction...

Current parser is capable of parsing schema definition language. It produces AST document for schema definition. The one step which is missing is to actually traverse this AST and create type definitions along the way.

So here is an example of Schema language parsing: [SchemaParserTest.php](https://github.com/webonyx/graphql-php/blob/master/tests/Language/SchemaParserTest.php). We tried to keep this project as close as possible to `graphql-js` so it maps directly to [schema-parser-test.js](https://github.com/graphql/graphql-js/blob/master/src/language/__tests__/schema-parser-test.js)

Next you can look into [buildASTsSchema.js](https://github.com/graphql/graphql-js/blob/master/src/utilities/buildASTSchema.js) which traverses this AST. Most of the methods and variable names are the same or similar in this lib, so make sure to search through the codebase to avoid unnecesarry work %)

Differencies are mainly in dropping prefixes and porting js functions as static methods:
e.g. `use GraphQL\Type\Definition\ObjectType` vs `import GraphQLObjectType from '../type/definition'` or `GraphQL\Utils\AST::valueFromAST` `import { valueFromAST } from './valueFromAST';`

And feel free to ask any questions along the way.

petecoop:
Thanks for the swift reply and some direction, I will have a look into it when I can!

petecoop:
I've opened a PR #91 for this, not done just yet but passing the initial tests.

It's pretty much identical to the js code, although at one point the js just started using a scoped var so had to turn it into a instanced class rather than just all static methods

PR #91, opened 2017-02-18
Build Schema & Schema Printer

petecoop:
I've made some progress on #90 and thought I'd open a PR so tests run etc.

So far I've only done 2 tests and need to do Docblocks, it looks like I also need to write a Schema Printer too to match the js tests.

### DIFF
--- a/src/Executor/Values.php
+++ b/src/Executor/Values.php
@@ -94,7 +94,7 @@ class Values
      * @return array
      * @throws Error
      */
-    public static function getArgumentValues($def, $node, $variableValues)
+    public static function getArgumentValues($def, $node, $variableValues = null)
     {
         $argDefs = $def->args;
         $argNodes = $node->arguments;
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -1,0 +1,304 @@
+<?php
+namespace GraphQL\Utils;
+
+use GraphQL\Language\Printer;
+use GraphQL\Schema;
+use GraphQL\Type\Definition\EnumType;
+use GraphQL\Type\Definition\InputObjectType;
+use GraphQL\Type\Definition\InterfaceType;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\ScalarType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Definition\UnionType;
+use GraphQL\Type\Definition\Directive;
+use GraphQL\Utils;
+
+/**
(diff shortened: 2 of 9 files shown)


vladar:
Looks ok to me, except for tests as coverage has dropped significantly.

petecoop:
Hey @vladar yeah not finished just yet, tests are not done for `BuildSchema` as I had to write `SchemaPrinter` first, working on it!

petecoop:
Hey @vladar, this is now finished! All the tests are identical to `graphql-js`. And everything follows `graphql-js` almost 1:1.

I did have some trouble with something I found in the lexer - where you're using `substr()` put it's not technically identical to the js equivalent of `slice()`, I tried to fix it in the lexer but it caused other breaking tests elsewhere, so worked around it instead.

src/Utils/SchemaPrinter.php
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    private static function isIntrospectionType($typename)
+    {
+        return strpos($typename, '__') === 0;
+    }
+
+    private static function isBuiltInScalar($typename)
+    {
+        return (
```
mcg-web: Use [constants](https://github.com/webonyx/graphql-php/blob/master/src/Type/Definition/Type.php#L20-L24) please :)

petecoop: done!

vladar:
Thanks for your work! I will publish 0.9.4 release soon with this feature included.

src/Executor/Values.php
```diff
@@ -94,7 +94,7 @@ public static function getVariableValues(Schema $schema, $definitionNodes, array
      * @return array
      * @throws Error
      */
-    public static function getArgumentValues($def, $node, $variableValues)
+    public static function getArgumentValues($def, $node, $variableValues = null)
```
mcg-web: @vladar here is a BC this should not be released in a minor version?

vladar: It is a relaxed requirement, all existing calls to this method will work as they do now. I don't see how it is a BC break?

mcg-web: its a BC because the signature change, this class is not final then it can be override :)

vladar: In theory this is possible, but we use this class as a container for pure stateless functions and reference these static methods directly by class name, so it doesn't imply any polimorphism. It is implicitly final. But yeah - we should probably mark such classes as final explicitly.

Anyway it's too late: I released 0.9.4 even before your first comment %) But I am sure it won't break anyone, so let's cross the fingers %)

PR #92, opened 2017-02-23
Fixed SyncPromiseAdapter::all() to not change the order of arrays

webmozart:
Right now, `SyncPromiseAdapter::all()` changes the order of arrays when resolving promises. For example, an array that should be

```
Array &0 (
    0 => '1'
    1 => 'value1'
    2 => 'value2'
    3 => 3
    4 => 'value2-value3'
    5 => Array &1 ()
)
```

is instead

```
Array &0 (
    0 => '1'
    3 => 3
    5 => Array &1 ()
    1 => 'value1'
    2 => 'value2'
    4 => 'value2-value3'
)
```

As a consequence, the array will be converted to a JSON object with numeric keys when encoding as JSON. `json_encode()` only converts arrays with incremental integer indices to JSON arrays.

This problem is actually detected simply by changing `assertEquals()` to `assertSame()` in the tests. I recommend to generally use `assertSame()` much more often than is done in this project (169 uses of `assertSame()` vs. 562 use of `assertEquals()`) to more easily discover similar flaws.

### DIFF
--- a/src/Executor/Promise/Adapter/SyncPromiseAdapter.php
+++ b/src/Executor/Promise/Adapter/SyncPromiseAdapter.php
@@ -96,6 +96,7 @@ class SyncPromiseAdapter implements PromiseAdapter
 
         foreach ($promisesOrValues as $index => $promiseOrValue) {
             if ($promiseOrValue instanceof Promise) {
+                $result[$index] = null;
                 $promiseOrValue->then(
                     function($value) use ($index, &$count, $total, &$result, $all) {
                         $result[$index] = $value;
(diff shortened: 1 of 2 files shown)


webmozart:
Thank you for the great work on this library by the way, I really love it! :ok_hand:

vladar:
Thanks for the fix!

I do agree that `assertSame` is usually more appropriate, but my IDE shows diff of expected / actual result when `assertEquals` fails and shows nothing useful  for `assertSame` (apart from error message). Streamlines writing tests a lot.

PR #98, opened 2017-03-16
php merge for associative array non-integer intexed

decebal:
`+=` is useful as an array merge just in case the array is numerically indexed, which is not the case here at all.

Problems solved by this:
```
fragment Fb on User {
  id
  _posts2R94ZT: posts(first: $first_1) {
    totalCount
  }
}

fragment Fc on User {
  _posts2R94ZT: posts(first: $first_1) {
    edges {
      node {
        id
        ...F5
      }
      cursor
    }
    pageInfo {
      hasNextPage
      hasPreviousPage
    }
  }
  id
  ...Fb
}
```

fragment selection would be merged as opposed to first fragment selection being ignored before.

### DIFF
--- a/src/Type/Definition/ResolveInfo.php
+++ b/src/Type/Definition/ResolveInfo.php
@@ -136,10 +136,10 @@ class ResolveInfo
                 if (isset($this->fragments[$spreadName])) {
                     /** @var FragmentDefinitionNode $fragment */
                     $fragment = $this->fragments[$spreadName];
-                    $fields += $this->foldSelectionSet($fragment->selectionSet, $descend);
+                    $fields = array_merge_recursive($this->foldSelectionSet($fragment->selectionSet, $descend), $fields);
                 }
             } else if ($selectionNode instanceof InlineFragmentNode) {
-                $fields += $this->foldSelectionSet($selectionNode->selectionSet, $descend);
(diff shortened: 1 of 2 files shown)


vladar:
Actually `+=` is mostly useful for associative arrays vs int-indexed arrays. Can you please update the PR with PHPUnit test case which you are trying to fix? As it is not clear from this PR what bug it actually tries to fix?

decebal:
@vladar I am not sure what you have in mind when stating `+=` is mostly useful. The effects of it are reflected better on [this thread](http://stackoverflow.com/questions/7059721/array-merge-versus) . In my app, useful means no loss of selected fields as opposed to maybe array merging performance. One other way of improving the merge performance and keep the same merging method would be to change the structure of selected fields by using as keys global integers to identify types as opposed to what currently is defined as `[type_name] => true`, but that can prove ineffective for other purposes.

vladar:
@decebal, now I see what you mean. The issue occurs when two fragments request the same field with different sub-selection. Current solution would only consider selection of first fragment visited, while we actually need to merge it.

Thanks for noticing and fixing! Please let me know when you are done with PR, I'll merge it.

decebal:
@vladar it's done now, unless there is something else I can relate to this, thanks

PR #100, opened 2017-03-20
Simplify install instructions

PowerKiKi:
Composer can create `composer.json` if missing, and letting 
composer pick the latest version by itself is more future proof than
hardcoding a version that need to be changed on each release.


vladar:
Thanks!

PR #104, opened 2017-03-31
make Schema serializable

andheiberg:
When using BuildSchema it becomes important to cache the output schema.

This can be achieved trivially with something like APC but only if you don’t have closures in your objects and/or you wrap them in SerializableClosures.

I’ve transformed everything so that it’s now serializable and will fit into APC if users are writing their resolvers without closures.

# TODO
- [ ] Fix test

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
   ],
   "require": {
     "php": ">=5.4,<8.0-DEV",
-    "ext-mbstring": "*"
+    "ext-mbstring": "*",
+    "opis/closure": "^2.4.0"
   },
   "require-dev": {
     "phpunit/phpunit": "^4.8"
--- a/src/Schema.php
+++ b/src/Schema.php
@@ -133,6 +133,11 @@ class Schema
         $this->typeResolutionStrategy = $config['typeResolution'] ?: $this->getEagerTypeResolutionStrategy();
     }
 
+    public function __wakeup()
+    {
+        $this->eagerTypeResolutionStrategy->addType(Introspection::_schema());
+    }
+
     /**
      * @return ObjectType
      */
--- a/src/Type/EagerResolution.php
+++ b/src/Type/EagerResolution.php
@@ -53,6 +53,15 @@ class EagerResolution implements Resolution
     }
 
     /**
+     * @param Type $type
+     */
+    public function addType($type)
+    {
+        $typeMap = Utils\TypeInfo::extractTypes($type);
+        $this->typeMap = array_merge($this->typeMap, $typeMap);
+    }
+
+    /**
      * @inheritdoc
      */
     public function resolveType($name)
--- a/src/Type/Introspection.php
+++ b/src/Type/Introspection.php
@@ -22,6 +22,7 @@ use GraphQL\Type\Definition\UnionType;
 use GraphQL\Type\Definition\WrappingType;
 use GraphQL\Utils;
 use GraphQL\Utils\AST;
+use Opis\Closure\SerializableClosure;
 
 class TypeKind {
     const SCALAR = 0;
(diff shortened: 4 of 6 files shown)


vladar:
While this feature makes sense I must think about it. The problem is that we'll have to mantain it and always make sure that schema is serializable in future versions. 

As for implementation:
1. I don't think Introspection even needs to be changed - it is not a part of schema - it is called during query execution in runtime. If it is not the case - then we should investigate why and probably fix (vs adding serialization to these closures)

2. I would really like to avoid external library dependencies. One option how we could do this is to add separate classes for object types created by `buildSchema` (which extend base ObjectType and InterfaceType) and can `sleep` and `wakeup`. This way we don't need to serialize closures.

AndreasHeiberg:
> I don't think Introspection even needs to be changed - it is not a part of schema - it is called during query execution in runtime. If it is not the case - then we should investigate why and probably fix (vs adding serialization to these closures)

Well I only made these changes as they were breaking serialization of the output schema from BuildSchema. So it would appear you need to do some fixing. I don't really care about the internals of introspection, so I will leave that to you as for how you wanna solve it. But I can confirm that this PR is indeed *a* solution to the problem.

> I would really like to avoid external library dependencies. One option how we could do this is to add separate classes for object types created by buildSchema (which extend base ObjectType and InterfaceType) and can sleep and wakeup. This way we don't need to serialize closures.

I have many things I would like to say about your internal class structure as I made clear in another PR trying to refactor your parser and AST. Yes in a sensible world there would be away for an AST to be written to array 👍 I did a practical approach to the problem and worked with what you have as I figured this would increase the likelihood of this being merged.

Again feel free to come up with a more sensible solution that requires more refactoring and/or fixing problems in existing code.

jane-olszewska:
> I don't think Introspection even needs to be changed - it is not a part of schema 

Introspection types are currently part of the schema. They are included in initial types of `EagerResolution`.

Since the `Resolution` interface is marked as highly experimental, it's hard to know how to modify it. I have therefore included a further workaround which means that after unserialising Schema, introspection queries continue to work.

AndreasHeiberg:
@vladar thoughts? Is this something you would want to merge?

vladar:
@AndreasHeiberg not in current form.

I agree that we need a way to avoid unnecessary parsing for those using `buildSchema`. But I do not agree with implementation using serializable closures (and internal comparision of types changing from === to ==).

I'll add separate task to track this feature request. Will include it in next major version of the lib, but it depends on some other work in the backlog.

In the mean-time I am curious about performance difference of caching AST + calling buildSchema on it vs caching whole schema. Have you tried benchmarking it?

As for AST as array - that is also something we do consider. Mostly because of possible [libgraphqlparser](https://github.com/graphql/libgraphqlparser) integration. So if that would work for you - this is the PR I am ready to accept. Will open separate issue for this as well.

vladar:
FYI I just made a quick benchmark which shows that unserializing AST and then calling `BuildSchema` is only ~10% slower than unserializing schema itself in PHP5.6 and actually a bit faster on PHP7.0+. So this might be a best way to do it for now.

Benchmark code in case if I missed something:

```php
<?php
namespace GraphQL\Benchmarks;

use GraphQL\Language\Parser;
use GraphQL\Utils\BuildSchema;

/**
 * @BeforeMethods({"setUp"})
 * @OutputTimeUnit("milliseconds", precision=3)
 */
class BuildSchemaBench
{
    private $astSerialized;
    private $schemaSerialized;

    public function setUp()
    {
        $ast = Parser::parse(file_get_contents(__DIR__ . '/schema.graphql'));
        $this->astSerialized = serialize($ast);

        $schema = BuildSchema::buildAST($ast);
        $this->schemaSerialized = serialize($schema);
    }

    /**
     * @Warmup(2)
     * @Revs(10)
     * @Iterations(3)
     */
    public function benchCachedSchema()
    {
        $schema = unserialize($this->schemaSerialized);

        // Try this to make sure all types are loaded
        // $schema->getTypeMap();
    }

    /**
     * @Warmup(2)
     * @Revs(10)
     * @Iterations(3)
     */
    public function benchCachedAst()
    {
        $ast = unserialize($this->astSerialized);
        $schema = BuildSchema::buildAST($ast);

        // Try this to make sure all types are loaded
        // $schema->getTypeMap();
    }
}
```

AndreasHeiberg:
Funny you should mention it I was playing with that earlier https://github.com/AndreasHeiberg/graphql-parser-php-extension

Almost finished building the `experimental` branch which instead of outputting json outputs proper PHP objects but I never got around to finishing it. Currently there's a memory bug in there of any C++ devs are are super eager to help out I can create a reduced test case :) That being said improving parsing performance doesn't matter much as you generally can avoid doing that for actual prod requests. I was mainly entertaining the idea to get a consistent AST foundation that will not have to be maintained that other stuff can be build on top of.

> unserializing AST and then calling BuildSchema is only ~10% slower than unserializing schema itself

Interesting approach hadn't thought about that at all. 🤔  we could switch over to doing something like that. But we would still have to attach resolvers to that schema so yeah it would involve some overhead which is fine.

> But I do not agree with implementation using serializable closures

That's fair it wouldn't have been my first choice either but I tried converting [this](https://github.com/webonyx/graphql-php/pull/104/files#diff-6f9b96c905acf9bade29996c9a3639dbR425) into just a regular array as the closure is entirely unnecessary, but it's seems like the framework internals is dependent on it. 

>  (and internal comparison of types changing from === to ==).

We don't like this either but we frankly couldn't dedicate more time to getting the framework to play nicely with unserialised schemas. Would love to see at better implementation.

> Will include it in next major version of the lib

👍

jrichard0725:
If I have built a schema, is there a way to get the AST back out of it for caching?

Looking for something like `$schema->getAST()` that I can then cache via memcached. Then load from memcached, deserialize then buildschema from.

Or if there is a better solution to not having to run extractTypes every time that would be great too. It's killing our response times right now though and caching seemed like an obvious strategy.

vladar:
@jrichard0725 Not yet, but in new version 0.10.0 it is possible. Check out these docs from future branch: https://github.com/webonyx/graphql-php/blob/v0.10/docs/type-system/type-language.md#performance-considerations

Will be released soon (I hope).

vladar:
Instead of serializing the schema, the recommended way (as of v0.10.0) is to serialize/cache the AST. `BuildSchema::build()` is now lazy by default, so it should be more efficient than any serialized schema.

See related [entry in docs](http://webonyx.github.io/graphql-php/type-system/type-language/#performance-considerations).

PR #106, opened 2017-04-07
fix __toString() for AST Node

AndreasHeiberg:
previously it would only shallowly convert to array causing json_encode to fail

### DIFF
--- a/src/Language/AST/Node.php
+++ b/src/Language/AST/Node.php
@@ -81,11 +81,45 @@ abstract class Node
      */
     public function __toString()
     {
+        $tmp = $this->toArray();
+        $tmp['loc'] = [
+            'start' => $this->loc->start,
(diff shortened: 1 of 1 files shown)


src/Language/AST/Node.php
```diff
@@ -81,11 +81,45 @@ private function cloneValue($value)
      */
     public function __toString()
     {
-        $tmp = (array) $this;
+        $tmp = $this->toArray();
         $tmp['loc'] = [
```
AndreasHeiberg: Just noticed this isn't needed btw

vladar:
I removed it. But also changed `Node::toArray()` signature to `Node::toArray($recursive = false)` to preserve backwards compatibility, so double check if your custom code uses this method.

PR #113, opened 2017-04-24
Use directives when calculating query complexity

pascaldevink:
While using query complexity to make sure certain fields would never be used in conjunction with other fields (in my case, a `search` field, which should have no side-effects), I found out that directives are ignored when calculating query complexity.
This makes it harder to combine queries, specially with an Apollo frontend.

By actively checking the directives for a field and ignoring the complexity for that field (and subsequent fields), the calculation is much more correct and as expected.

I don't have too much experience with this library directly (I'm using it in conjunction with https://github.com/overblog/GraphQLBundle), so any help, tips, tricks and/or pointers are welcome.


vladar:
@mcg-web Can you please review this PR as query complexity is something you contributed in past?

lucasfeliciano:
👍  Nice one!

vladar:
Thanks for this PR!

PR #115, opened 2017-05-02
fix for duplicate usage of "id"

TerraSkye:
only uses current class to check for methods instead of all its parents.


vladar:
Thanks for the suggested fix. But I think the better solution would be to just prefix resolvers method names with `resolve (so that full method name became `resolve{$fieldName}`). This is example after all, so I'd like to keep it clear for end-users what's going on.

If you agree to change the fix - I'd greatly appreciate your help!

TerraSkye:
I also saw that its not the only point the problem occurs. So it needs a bit more wurk

vladar:
I fixed it in https://github.com/webonyx/graphql-php/commit/f668300cd87c471a026fd0026c9b9e65e7765554

PR #116, opened 2017-05-07
Resolve callables by default

leocavalcante:
Enables any `callable`s to be resolved by default. Then you can build a schema with the Builder (lefting resolver functions to the default field resolver) and create invokable classes as resolvers.

```graphql
type Query {
  message: String
}
```

```php
class MessageResolver
{
    public function __construct(string $greet)
    {
        $this->greet = $greet;
    }

    public function __invoke($root, $args, $context)
    {
        return $this->greet;
    }
}

$rootValue = [
    'message' => new MessageResolver('Hello')
];

GraphQL::execute(BuildSchema::build('schema.graphql'), $rootValue);
```

### DIFF
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -918,7 +918,7 @@ class Executor
             }
         }
 
-        return $property instanceof \Closure ? $property($source, $args, $context) : $property;
+        return is_callable($property) ? $property($source, $args, $context) : $property;
     }
 
     /**
--- a/tests/Executor/ResolveTest.php
+++ b/tests/Executor/ResolveTest.php
@@ -61,6 +61,30 @@ class ResolveTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @it default function calls callables
+     */
+    public function testDefaultFunctionCallsCallables()
+    {
+        $schema = $this->buildSchema(['type' => Type::string()]);
+        $_secret = 'secretValue' . uniqid();
+
+        $source = [
+            'test' => new class($_secret) {
+                public function __construct($_secret) {
+                    $this->_secret = $_secret;
+                }
+                public function __invoke() {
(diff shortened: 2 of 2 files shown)


vladar:
I do understand your intent, but I'd like to do some benchmarks first. Default resolver may be called thousands of times per query and replacing operation (instanceof) with function call (is_callable) might be expensive. 

Also even with current implementation you can set your own default resolver if you need such behavior. I am on vacation now, but will take a closer look at this once I am back.

vladar:
FIY, I did some benchmarks, and the results are tolerable. Can you change the target of this PR from `master` to `v0.10` - then I'll merge it.

leocavalcante:
Updated base branch to v0.10.

johanobergman:
I'm having some trouble with the `is_callable` check. There's a field for a certain type in my schema called `date`, and during the execution of an introspection query, the `is_callable` check for that field is reached. Since `date` is a callable php function this results in a call to `date()` instead of returning the field name.

In order to support your use-case without errors, you might want to also check against `property` not being a string, but I'm not sure how that's going to affect performance.

vladar:
That's funny (actually not) %) Anyway, I think I'll revert this PR. 

One of the new features in v0.10.0 is ability to pass custom default resolver to `Executor::execute()` (different from static one). So if `laravel-php` needs this, they can just pass their own default resolver.

@johanobergman Thanks for sharing this!

leocavalcante:
Hi @johanobergman, I can see that, but to clarify: is the field name `date` or its value is `date`?

johanobergman:
@leocavalcante It's the field name, the value/type is a string.

@vladar Sounds good! v0.10.0 looks promising, the improved lexer algorithm took one of my queries from 125 ms to 70 ms excluding validation, great work!

Issue #123, opened 2017-05-27
Confusing usage of InvariantViolation error


mcg-web:
There some confusing usage of `GraphQL\Error\InvariantViolation` error in the lib, sometime this error is use to send some useful information to graphQL clients (wrong types value parsing [like here](https://github.com/webonyx/graphql-php/blob/master/src/Type/Definition/IntType.php#L60)) and sometime we use it to send system error (sensitive information) that should not be display to client ([here an example](https://github.com/webonyx/graphql-php/blob/master/src/Type/Definition/ResolveInfo.php#L155)). Can we separate both information throw two distinct exception? This we help to create a more safe prod environment without hiding useful errors to client...

vladar:
Agree. In general invariant violation should be used for system errors only. This exception should not be used for errors exposed to clients. 

Can you prepare PR for suggested changes? Let me know if you need anything from me to make it happen.

mcg-web:
Ok i'll PR this. Thank you for quick answer ;).

PR #124, opened 2017-05-28
Use dedicated exception for scalar type parsing error

mcg-web:
see #123

### DIFF
--- a/src/Error/UserError.php
+++ b/src/Error/UserError.php
@@ -1,0 +1,14 @@
+<?php
+namespace GraphQL\Error;
+
+/**
+ * Class UserError
+ *
+ * Note:
+ * Error that can be display safely to client...
+ *
+ * @package GraphQL\Error
+ */
+class UserError extends InvariantViolation
+{
+}
--- a/src/Type/Definition/ObjectType.php
+++ b/src/Type/Definition/ObjectType.php
@@ -1,6 +1,7 @@
 <?php
 namespace GraphQL\Type\Definition;
 use GraphQL\Error\InvariantViolation;
+use GraphQL\Error\UserError;
 use GraphQL\Utils;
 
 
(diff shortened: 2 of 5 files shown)


src/Error/UserError.php
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Class UserError
+ *
+ * Note:
+ * Error that can be display safely to client...
+ *
+ * @package GraphQL\Error
+ */
+class UserError extends InvariantViolation
```
mcg-web: extends `InvariantViolation` to avoid BC, next major version we can change this.

TerraSkye: I think its an acceptable BC, if we put in release notes, as it is a security enhanchement?

mcg-web: BC is never acceptable or not it is just a reality, if your new release reflect the BC no trouble but it is very important to dependencies...

TerraSkye: Yeah i forgot that part. My bad, your suggestion seems the best then!

src/Type/Definition/ObjectType.php
```diff
@@ -145,7 +145,7 @@ public function getInterfaces()
             foreach ($interfaces as $iface) {
                 $iface = Type::resolve($iface);
                 if (!$iface instanceof InterfaceType) {
-                    throw new InvariantViolation("Expecting interface type, got " . Utils::printSafe($iface));
+                    throw new UserError(sprintf('Expecting interface type, got %s', Utils::printSafe($iface)));
```
vladar: This looks like a perfect InvariantViolationError for me, as it's server developer error (in type definition). Is there any reason why this must be UserError?

mcg-web: ok i'm rolling back this one!

PR #127, opened 2017-06-06
Support PHP7 error exceptions everywhere

danez:
This handles error exceptions in all places and also replace \Error with \Throwable.

This is a breaking change as the API for the interface PromiseAdapter changes, but otherwise there is now way in catching php7 errors that might happen in the Promise.

### DIFF
--- a/src/Deferred.php
+++ b/src/Deferred.php
@@ -54,6 +54,8 @@ class Deferred
             $this->promise->resolve($cb());
         } catch (\Exception $e) {
             $this->promise->reject($e);
+        } catch (\Throwable $e) {
+            $this->promise->reject($e);
         }
     }
 }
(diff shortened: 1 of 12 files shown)


vladar:
I'd really like to accept this PR. Minor breaks are ok for next major version, but this PR breaks every PHP version older than 7.0+. And we must still support previous PHP versions. Can you try running tests on PHP5.6 and fix failures?

danez:
It was meant to be bc. I had #126 before which broke php <7, this should have been the fixed version.
Let me check.

danez:
Okay I fixed it.

vladar:
Thanks! I'll merge it when I start working on next major version

vladar:
@danez Can you change target branch of this PR from `master` to `v10.0` ?

danez:
Done

Issue #108, opened 2017-04-07
AST as array


vladar:
This is required to leverage [libgraphqlparser](https://github.com/graphql/libgraphqlparser) in future.

joseluisq:
:+1:

vladar:
Unfortunately, after looking deeper into this I realized that this would be a really major break + it will make dealing with ASTs way less pleasant from IDE. 

I implemented a different solution to enable compatibility with libgraphqlparser: now it is possible to create AST from array returned by libgraphqlparser. 

To make such conversion performant I had to replace all arrays of nodes (such as `selections`, `directives`, `arguments`, etc) with `NodeList` class which lazily converts arrays from libgraphqlparser to Node instances.

This way typical integration with libgraphqlparser (e.g. using [graphql-parser-php](https://github.com/dosten/graphql-parser-php)) would looks like this:

```php
$parser = new \GraphQL\Parser(); // graphql-parser-php
$astAsArray = $parser->parse('query { name }');

// Convert AST array to lazily-evaluated DocumentNode:
$ast = GraphQL\Language\AST\DocumentNode::fromArray($astAsArray);
$result = GraphQL\GraphQL::execute($schema, $ast);
```

vladar:
Released in v0.10.0. See [docs](http://webonyx.github.io/graphql-php/reference/#graphqlutilsast) on how to convert AST from/to array. 

This is a lazy operation (array is converted to AST as it is traversed), so should be performant enough.

PR #128, opened 2017-06-12
Rework Lexer.php to avoid use of mb_substring.

aelnagger:
This resolves issues lexing large queries by implementing an internal array buffer and using an internal position to manage reading the query "stream". As of this commit, all unit tests pass.

As a bit of background, we found when we switched to this library for GraphQL processing that we experienced PHP timeouts on large requests, typically this was triggered by mutations which had input lists of nontrivial size. I ran a profiler over the requests, and identified mb_substring as the culprit.

Benchmark comparisons:

|Benchmark | Old Code | New Code |
| -------------- | ------------- | ------------ |
| Small Query | 993.013 (ms) | 275.589 (ms) |
| Large Query | 87,769.078 (ms) | 2,772.053 (ms) |

### DIFF
--- a/src/Language/Source.php
+++ b/src/Language/Source.php
@@ -11,6 +11,11 @@ class Source
     public $body;
 
     /**
+     * @var array
+     */
+    public $buf;
+
+    /**
      * @var int
      */
     public $length;
@@ -28,7 +33,8 @@ class Source
         );
 
         $this->body = $body;
-        $this->length = mb_strlen($body, 'UTF-8');
+        $this->buf =  array_map('GraphQL\Utils::ord', preg_split('//u', $body, -1, PREG_SPLIT_NO_EMPTY));
+        $this->length = count($this->buf);
         $this->name = $name ?: 'GraphQL';
     }
 
--- a/benchmarks/HugeSchemaBench.php
+++ b/benchmarks/HugeSchemaBench.php
@@ -46,9 +46,13 @@ class HugeSchemaBench
 
         $this->schema = $this->schemaBuilder->buildSchema();
 
-        $queryBuilder = new QueryGenerator($this->schema, 0.05);
+        $smallQueryBuilder = new QueryGenerator($this->schema, 0.05);
         $this->descriptor = $this->schema->getDescriptor();
-        $this->smallQuery = $queryBuilder->buildQuery();
+        $this->smallQuery = $smallQueryBuilder->buildQuery();
+
+        $largeQueryBuilder = new QueryGenerator($this->schema,.5);
+        $this->largeQuery = $largeQueryBuilder->buildQuery();
+
     }
 
     public function benchSchema()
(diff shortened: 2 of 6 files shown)


vladar:
Awesome! I was aware that current solution is suboptimal and had plans to rewrite it as soon as someone complains about parsing performance %) Really happy that you nailed it, thanks! 

But I have to do some internal testing with this new version before merging.

aelnagger:
Sounds good! This was a blocker for us adopting your framework as a graphQL provider, but it was a relatively straightforward optimization to make. I think long term the ideal solution would be to find a real multi-byte compatible stream, but this at least makes things considerably better!

src/Language/Source.php
```diff
@@ -28,7 +33,8 @@ public function __construct($body, $name = null)
         );
 
         $this->body = $body;
-        $this->length = mb_strlen($body, 'UTF-8');
+        $this->buf =  array_map('GraphQL\Utils::ord', preg_split('//u', $body, -1, PREG_SPLIT_NO_EMPTY));
```
aelnagger: Unfortunately this is causing pathological memory usage for larger requests (current testing is attempting to allocate 256M for a 5M `$body`.

vladar: Yeah, I was suspecting that memory problems are possible with this approach (had previous experience with `preg` functions behaving unpredictably on large strings), but didn't have a chance yet to test it myself. Good that you found this issue before merging.

I guess we should write own utf8 stream to actually resolve the issue.

vladar: @aelnagger I started experimenting with separate utf8 stream, but we should avoid duplicating our efforts, so please let me know if you are working on other solution now? If not, I'll continue.

Also, couple notes on best practices (regardless of our final solution):

1. GraphQL queries shouldn't be big. We should use variables as much as possible instead of inline values (because variables are parsed on C-level by `json_encode`).

2. Ideally parsing should be done by C module, so the most important issue for performance is #108. When it is implemented you should be able to use PHP module for parsing (e.g. https://github.com/dosten/graphql-parser-php) and then just pass this array to validation/execution steps.

aelnagger: @vladar I have a naive implementation using the original `$body` string as a packed value. I have the tests passing on my Windows environment, but I'm getting segfaults in my CentOS VM (this might be related to some other customizations going on). That said, I don't have a real stream implementation, but given how inefficient the base PHP value types are, parsing as a C module would certainly be much better.

vladar: I've built a proof-of-concept version without mb or preg usage, which shows good results (see https://github.com/webonyx/graphql-php/issues/137#issuecomment-310839638). Will push it to new branch for v0.10.0 when ready.

vladar:
Closing this. Other solution is implemeted in `v0.10` branch via https://github.com/webonyx/graphql-php/commit/a0657b784779d770b2aeef80bd98ed36c19dbc70. 

I encourage everybody interested to test their queries against it to catch possible bugs earlier %)

src/Language/Source.php
aelnagger: Hey @vladar, sorry for the silence yesterday, busy weekend! I worked up https://github.com/aelnagger/graphql-php/tree/utf_8_decoder a few days ago, passing all the tests, but haven't cleaned it up or run the benchmarks just yet. There's definitely an issue with garbage collection on larger queries though. Making use of graphQL variables is definitely the right solution.

PR #138, opened 2017-06-26
Layz loading reduced test case [DO NOT MERGE]

andheiberg:
Here is a reduced test case attempting to show the issues with type spreads when using lazy type resolution.

Test query:

```curl 'http://localhost:8080/' -H 'content-type: application/json' -H 'Connection: keep-alive' --data-binary '{"query":"{ node(id: 2) { id, ... on User { email } } }","variables":null}'```

Checking out 383ca58 and running the test query results in:

```
{"data":{"node":{"id":"ID","email":"jane@example.com"}}}
```

Checking out 	c20a1e7 and running the test query results in:

```
{"data":{"node":{"id":"ID"}}}
```

Ignoring the fact that something is obviously very strange with the `"id":"ID"` as it should be `"id":"2"` you will see that the second version using the LazyTypeResolution does not correctly handle the type spread in the query.

### DIFF
--- a/examples/01-blog/.gitignore
+++ b/examples/01-blog/.gitignore
@@ -1,0 +1,1 @@
+schema.descriptor
--- a/examples/01-blog/Blog/Type/QueryType.php
+++ b/examples/01-blog/Blog/Type/QueryType.php
@@ -55,7 +55,14 @@ class QueryType extends ObjectType
                         throw new \Exception("Exception message thrown in field resolver");
                     }
                 ],
-                'hello' => Type::string()
+                'hello' => Type::string(),
+                'node' => [
+                    'type' => Types::node(),
+                    'description' => 'Get any node. Will require type spread.',
+                    'args' => [
+                        'id' => Types::nonNull(Types::id())
+                    ]
+                ]
             ],
             'resolveField' => function($val, $args, $context, ResolveInfo $info) {
                 return $this->{$info->fieldName}($val, $args, $context, $info);
@@ -93,5 +100,10 @@ class QueryType extends ObjectType
     public function deprecatedField()
     {
         return 'You can request deprecated field, but it is not displayed in auto-generated documentation by default.';
+    }
+
+    public function node($rootValue, $args)
+    {
+        return DataSource::findUser($args['id']);
     }
 }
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by vladar** [comment]
Thanks, will look into this

**Comment by orls** [comment]
Note that you need to run the query _twice_ on the head commit to see the error (first request doesn't use LazyResolution, so it can create the `schema.descriptor` file).

**Comment by orls** [comment]
Ultimately, this is failing because of the [strict equality check in `Executor::doesFragmentConditionMatch`](https://github.com/AndreasHeiberg/graphql-php/blob/c20a1e7c5a31e8e7028e68ebcee86dfc19cd2934/src/Executor/Executor.php#L559).

The Lazy types resolver function given creates a fresh instance of `User` on-demand; but the _rest_ of the example fixures refer to a singleton created by `Types::user()`.

If you replace the references to `Types::user()`  with the string `'User' in the example's  `NodeType` and `QueryType` classes, then this works.

While it's _possible_ to fix this example in that way, it does feel like a wider/general issue for the library. If lazy resolution is used, then for every place that the library does strict equality checks, the Type instances in question must _all_  come from the lazy type resolver originally. There's a lot of risk of users shooting themselves in the foot.

Wouldn't it be safe to just check that instances are of the same class? Or only allow specifying Types by their name, _not_ providing whole instances? (i appreciate that's a bigger change for existing users)

**Comment by vladar** [comment]
@orls Thanks for your research. You are correct about the cause of this issue. 

But the general rule for both: this lib and reference `graphql-js` implementation is that type must be represented by single instance. That's why we recommend using [type registry](http://webonyx.github.io/graphql-php/type-system/#type-registry). 

Allowing multiple instances of the same class may cause other sort of issues if type class is extended and stateful.

In short, type loader of this example is incorrect. It has to return the same instance from registry. So it should be expressed like this instead:
```php
    $schemaConfig['typeResolution'] = new LazyResolution($schemaDescriptor, function ($typeName) {
        $method = lcfirst($typeName);
        if (method_exists(Types::class, $method)) {
            return Types::$method();
        }
        return null;
    }
```

But obviously the library does too little to instruct user about this. In our case it must throw when it sees different instances for the same type name with clear message of what is going on. 

That is the solution I plan to implement to fix this, but please let me know if you guys have other thoughts.

vladar:
@AndreasHeiberg Thanks a lot for reproducible test case!

vladar:
Added check with meaningful error message to make sure loader always returns the same instance.

PR #142, opened 2017-07-10
Add example for building schema from IDL

symm:
I couldn't find it documented so wrote an example of how to use `BuildSchema` with IDL shorthand and then wire it up to some resolvers.

### DIFF
--- a/examples/02-shorthand/README.md
+++ b/examples/02-shorthand/README.md
@@ -1,0 +1,19 @@
+# Parsing GraphQL IDL shorthand
+
+Same as the Hello world example but shows how to build GraphQL schema from shorthand 
+and wire up some resolvers
+
+### Run locally
+```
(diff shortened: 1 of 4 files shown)


vladar:
Thanks a lot!

jasonbahl:
@symm This is pretty sweet. . .but doesn't this cause all the resolve functions in the rootValue to be executed every time GraphQL is executed whether the field is queried for or not? Or is the GraphQL internals smart enough to only execute the resolvers for that rootValue the same it does for the rest of the resolvers?

symm:
@jasonbahl The anonymous functions will only be executed if the GraphQL lib needs to resolve them. You can verify by putting a sleep call on the `Addition` resolver and calling the echo method 👍 

That said, it's just a demo and I'd fetch the resolvers from a DI container in a real app.

PR #161, opened 2017-09-03
Error code

calvera:
Now it allows developer to return error codes from Exceptions

### DIFF
--- a/tests/ErrorTest.php
+++ b/tests/ErrorTest.php
@@ -81,6 +81,15 @@ class ErrorTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @it serializes to include message
+     */
+    public function testSerializesToIncludeMessageAndCode()
+    {
+        $e = new Error('msg', null, null, null, null, null, 1000);
+        $this->assertEquals(['message' => 'msg', 'code' => 1000], $e->toSerializableArray());
+    }
+
+    /**
      * @it serializes to include message and locations
      */
     public function testSerializesToIncludeMessageAndLocations()


src/Error/Error.php
```diff
@@ -124,9 +128,9 @@ public static function formatError(Error $error)
      * @param array|null $path
      * @param \Exception|\Error $previous
      */
-    public function __construct($message, $nodes = null, Source $source = null, $positions = null, $path = null, $previous = null)
+    public function __construct($message, $nodes = null, Source $source = null, $positions = null, $path = null, $previous = null, $code = null)
     {
-        parent::__construct($message, 0, $previous);
+        parent::__construct($message, $code ?: 0, $previous);
```
vladar: I don't think it's worth adding `$code` argument to `GraphQL\Error\Error`. This exception is for internal library usage and we don't use exception codes in the library.

Instead, I suggest to include the code in formatted error only if there is a previous exception and this exception has a code. 

So the only place which should be adjusted is `GraphQL\Error\FormattedError`

calvera: so it is misused in overblog/graphqlbundle? @overblog uses it in errorhandler

vladar: I am not familiar with overblog bundle as I don't use Symfony, but if you point me the line where it is misused, I'll take a look

calvera: as i understand your previous message, it is not supposed that anybody uses GraphQL\Error\Error. so https://github.com/overblog/GraphQLBundle/blob/0b3b90ec54f06c7cc686962b3528fc0aae0b2e5e/Error/ErrorHandler.php#L115 and https://github.com/overblog/GraphQLBundle/blob/0b3b90ec54f06c7cc686962b3528fc0aae0b2e5e/Error/ErrorHandler.php#L127 is therefore misuse...

those lines are the reason i propose my PR

thanks!

vladar: Well looks like they just replace a message of some exceptions to "Internal server error". They don't use exception code for `GraphQL\Error\Error` as well. The way they use this class is aligned with the way how we use it in this library.

I still don't quite understand why do you need code for base exception vs just using the code of the previous one?

calvera: yes, they do not use code, but i would like to throw exception with code in my resolver and have the code in the error
after the changes i propose, i get it like this:
```yml
{
    "data": {
        "checkOut": null
    },
    "errors": [
        {
            "message": "Cannot check out, not checked in",
            "code": 1052,
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "checkOut"
            ]
        }
    ]```

vladar: I do understand this. And you can access this code from `$error->getPrevious()->getCode();` right in error formatter. You don't need to add $code argument to constructor of `GraphQL\Error\Error` for this.

calvera: good point, i did not notice error formatter, thanks!

PR #185, opened 2017-10-16
Fix typo in error handling docs

ivome:
The function name in docs does not match the one from the library. I think the function name in the docs is actually more consistent, but changing that would mean a breaking change. So this fixes the docs only. Let me know if we should rather adjust the code.


vladar:
Thanks for noticing! But for some reason, this PR includes 3 commits (with some old ones) so I am not sure if it will process the merge correctly. So I just fixed it in master myself. Thanks again!

PR #186, opened 2017-10-17
Extend functionality

crirus:
Enum accepts both value and keys as integer
Utils have some new functions
FieldArgument accepts own type as config

### DIFF
--- a/src/Type/Definition/EnumType.php
+++ b/src/Type/Definition/EnumType.php
@@ -80,6 +80,8 @@ class EnumType extends Type implements InputType, OutputType, LeafType
                         }
                     } else if (is_int($name) && is_string($value)) {
                         $value = ['name' => $value, 'value' => $value];
+                    } else if (is_int($name) && is_int($value)) {
+                        $value = ['name' => $value, 'value' => $value];                        
                     } else {
                         throw new InvariantViolation("{$this->name} values must be an array with value names as keys.");
                     }
--- a/src/Type/Definition/EnumType.php
+++ b/src/Type/Definition/EnumType.php
@@ -79,9 +79,7 @@ class EnumType extends Type implements InputType, OutputType, LeafType
                             $value += ['name' => $name, 'value' => $name];
                         }
                     } else if (is_int($name) && is_string($value)) {
-                        $value = ['name' => $value, 'value' => $value];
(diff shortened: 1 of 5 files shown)


src/Type/Definition/EnumType.php
```diff
@@ -80,6 +80,8 @@ public function getValues()
                         }
                     } else if (is_int($name) && is_string($value)) {
                         $value = ['name' => $value, 'value' => $value];
+                    } else if (is_int($name) && is_int($value)) {
+                        $value = ['name' => $value, 'value' => $value];
```
vladar: This is not spec-compliant. Enum names must be alpha-numeric strings.

src/Utils/Utils.php
```diff
@@ -471,4 +474,49 @@ public static function withErrorHandling(callable $fn, array &$errors)
...
+    public static function mapValues($traversable, callable $valueFn) {
+        self::invariant(is_array($traversable) || $traversable instanceof \Traversable, __METHOD__ . ' expects array or Traversable');
+
+        $map = [];
+        foreach ($traversable as $key => $value) {
+            $newValue = $valueFn($value, $key);
+            $map[$key] = $newValue;
+        }
+        return $map;
+    }
```
vladar: How is this function different from regular `map()` in the same class, they look identical?

src/Type/Definition/EnumType.php
crirus: Yes, too bad as I have some enum with numbers that should show up in GraphIQL as so, not something like _1, _2

src/Utils/Utils.php
crirus: I removed mapValues, wasn't aware at the time of all the porting you did there with different names than JS version

vladar:
Just curious, do you plan to add extendsSchema functionality in this PR or plan to open another one? I would recommend putting all changes into single PR so that I could review the whole thing before merging. Thanks!

crirus:
It depends if  you want this part in the main project, I already have something in a separate repository crirus/graphql-php-tools, but it's not mature enough for a merge with main library.

I have there both extend and merge functionality. However, in that code there is a lot that sounds like BuildSchema bits.

Maybe someone can help tidying it a bit.

I am deep into building microservices graphql now, will get back to merging when I have those micros up and running.

PR #187, opened 2017-10-17
`$data` Unsupported operand types  error fix

gabidj:
the operator `+=` only works like `array_merge ($left, $right) ;` if both variables are arrays

### DIFF
--- a/examples/01-blog/graphql.php
+++ b/examples/01-blog/graphql.php
@@ -40,8 +40,6 @@ try {
         $data = $_REQUEST;
     }
     
-
-    
     $data += ['query' => null, 'variables' => null];
 
     if (null === $data['query']) {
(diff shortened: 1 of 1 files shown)


examples/01-blog/graphql.php
```diff
@@ -36,9 +36,18 @@
     if (isset($_SERVER['CONTENT_TYPE']) && strpos($_SERVER['CONTENT_TYPE'], 'application/json') !== false) {
         $raw = file_get_contents('php://input') ?: '';
         $data = json_decode($raw, true);
+        // $data += will cause 'Unsupported operand types' Fatal error on `null`
+        // check if decoded data is an array (or stdobject) - not null
+        // if any kind of data is present we don't want to lose it
+        if ($data === null) {
+            $data = [];
+        }
```
vladar: Thanks! But can you replace this with just `json_decode($raw, true) ?: []` and remove these comments, since we use comments in examples to explain how graphql works vs php internals?

gabidj: Sure. That was a quick solution. 
The comments were actually for who reads this pull request so they know what was my issue.

You are right about the `?: []` syntax.  

I will perform the changes.

PR #193, opened 2017-11-02
Add support for setting extensions from resolvers.

aelnagger:
This adds an API to ResolveInfo which allows resolvers
to directly add items to be returned to the client in
the `extensions` element of the response body.

see http://facebook.github.io/graphql/October2016/#sec-Response-Format

### DIFF
--- a/src/Executor/ExecutionContext.php
+++ b/src/Executor/ExecutionContext.php
@@ -56,6 +56,11 @@ class ExecutionContext
      */
     public $errors;
 
+    /**
+     * @var array
+     */
+    public $extensions;
+
     public function __construct(
         $schema,
         $fragments,
@@ -65,7 +70,8 @@ class ExecutionContext
         $variables,
         $errors,
         $fieldResolver,
-        $promiseAdapter
+        $promiseAdapter,
+        $extensions
     )
     {
         $this->schema = $schema;
@@ -77,11 +83,27 @@ class ExecutionContext
         $this->errors = $errors ?: [];
         $this->fieldResolver = $fieldResolver;
         $this->promises = $promiseAdapter;
+        $this->extensions = $extensions ?: [];
     }
 
     public function addError(Error $error)
     {
         $this->errors[] = $error;
         return $this;
+    }
+
+    public function setExtension($key, $value)
+    {
+        $this->extensions[$key] = $value;
+        return $this;
+    }
+
+    public function getExtension($key)
+    {
+        if(array_key_exists($key, $this->extensions))
+        {
+          return $this->extensions[$key];
+        }
+        return null;
     }
 }
--- a/src/Executor/ExecutionContext.php
+++ b/src/Executor/ExecutionContext.php
@@ -56,11 +56,6 @@ class ExecutionContext
      */
     public $errors;
 
-    /**
-     * @var array
-     */
-    public $extensions;
-
     public function __construct(
         $schema,
         $fragments,
@@ -70,8 +65,7 @@ class ExecutionContext
         $variables,
         $errors,
         $fieldResolver,
-        $promiseAdapter,
-        $extensions
+        $promiseAdapter
     )
     {
         $this->schema = $schema;
@@ -83,27 +77,11 @@ class ExecutionContext
         $this->errors = $errors ?: [];
         $this->fieldResolver = $fieldResolver;
         $this->promises = $promiseAdapter;
-        $this->extensions = $extensions ?: [];
     }
 
     public function addError(Error $error)
     {
         $this->errors[] = $error;
         return $this;
-    }
-
-    public function setExtension($key, $value)
-    {
-        $this->extensions[$key] = $value;
-        return $this;
-    }
-
-    public function getExtension($key)
-    {
-        if(array_key_exists($key, $this->extensions))
-        {
-          return $this->extensions[$key];
-        }
-        return null;
     }
 }
--- a/src/Executor/ExecutionResult.php
+++ b/src/Executor/ExecutionResult.php
@@ -53,6 +53,16 @@ class ExecutionResult implements \JsonSerializable
     private $errorsHandler;
 
     /**
+     * @var callable
+     */
+    private $extensionsHandler;
+
+    /**
+     * @var mixed
+     */
+    private $contextValue;
+
+    /**
      * @param array $data
      * @param array $errors
      * @param array $extensions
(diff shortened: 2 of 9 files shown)


vladar:
Thanks for the PR! While it seems logical, I am not sure if it should be included in this lib. The reason is that we actually embrace pure resolvers. And this PR may create an impression that the lib advocates for impure resolvers.

Someday I plan to implement memoization for the most common cases, which will be impossible in the general case with impure resolvers (or may produce unexpected results).

Of course, users can write impure resolvers even now (with context or statics), but this situation is very different from the situation when the lib embraces such behavior.

But if other library users will find it useful I can change my mind.

aelnagger:
Hey @vladar - thanks for the feedback. I originally went digging through the codebase to figure out how to use `extensions` to enable instrumentation and cache control for a gateway I intend to use in front of the service I'm running using this engine, and was surprised to find that the implementation in the ExecutionResult wasn't accessible anywhere. That said, I'm not sure you'd be actively advertising impure resolvers. Honestly I suspect that the thorough documentation around Deferred for solving the N+1 problem is probably a larger issue for that!

I certainly understand that you _can_ use statics or the context to drive this sort of behavior, but I wanted to try to fit it into the framework in a way that didn't seem like flagrant disregard for the specification.

I do think that you're going to see a number of people who are keenly interested in using extensions due to recent projects like [apollo-tracing](https://github.com/apollographql/apollo-tracing). I'll go ahead and ask folks that I know to comment on this patch as well!

vladar:
Note that even if we choose to merge this, the `extensions` section defined in specs is expected to be a Map, not a List.

aelnagger:
@vladar That's an excellent point!

I've amended the patch to ensure we get a map instead of a list.

jasonbahl:
This looks sweet! 

I have Apollo Tracing implemented for WPGraphQL via instrumenting the entire schema, but a more native way to interact with extensions from within Resolvers looks fantastic!

spinx:
First of all @vladar thank you and other contributors for your work on this lib, it made our journey towards adopting GraphQL just immensely easier. 

I believe apollo-tracing is solving one of the biggest downsides of GraphQL, especially for smaller teams - that's being able to easily debug and resolve potential issues and having confidence when migrating to this weird new way that can result in hundreds of DB calls, where seemingly nothing can be cached super effectively. 

I'd expect libraries like graphql-php and others to implement tracing/debugging and other extensions as a core feature, or at least a way to add that as a 'plugin' very easily. 

How this is implemented is another discussion altogether, perhaps more relevant to this thread. From a stand point of a lib, there needs to be a choice, obviously. And with that sorted I don't see a problem - I as probably many others will opt for unlocking the value that this would provide or at least weigh up prons and cons of different approaches - which can all be described in docs.

vladar:
After some thinking I agree that we need a way to help setting extensions from resolvers, but this is such a sugar feature that I don't like the idea of modifying execution context or ResolveInfo (the name implies that it is a readonly structure).

I suggest a different approach by adding GraphQL\ExtensionsContainer interface (suggest a better name) which anyone can implement in the user-land context class (we can also provide a trait with standard implementation as it is trivial).

Then after execution is complete and if context implements this interface we'll copy it to ExecutionResult->extensions. This way the feature will mainly stay in the user-land.

Also it will require less typing in resolvers since user context is a 3rd argument, and ResolveInfo is 4th.

Please let me know if it works for you.

crirus:
Why not having a singleton for Executor and access it via a static like Executor::instance()->setExtension() within a resolver?

vladar:
This feature will make them impure anyway but adding singleton here is even a bigger sin. Right now resolvers are not even aware of the executor, there is no reason to couple them like this. 

Plus technically it is possible to have multiple executors (for testing or some non-trivial usages).

crirus:
How about GraphQL class then? We already use it to set promise adapter. 
Could be better used to add extensions than forcing my context class to use a graphql interface?

vladar:
You are not forced, it's an opt-in and convenience feature. Just don't implement this interface and keep using your context as is. You can always set extensions manually the way you want (or don't set them at all).

GraphQL class is just a convenience facade class. You can use this library without it and certainly shouldn't rely on it for any specific logic.

aelnagger:
Hey folks, I've updated my branch with a cleverly named `ExtendableContext` interface which allows the executor to pull extensions from the user-supplied context.

I'm open to renaming if we can come up with something that more clearly communicates the intent.

vladar:
Hey, thanks for the update, I'll merge it as soon as I get a chance to work on the next release (but maybe will rename ExtendableContext, so open to suggestions as well)

vladar:
So what would you suggest on the subject of this PR (it is not entirely clear from your last reply)?

mcg-web:
Here an example of how this can be done in project:

```php
use GraphQL\GraphQL;

$contextValue = new \ArrayObject(['extensions' => ['foo' => [], 'bar' => []]]);

$result = GraphQL::executeQuery($schema, $requestString, $rootValue = null, $contextValue , $variableValues = null, $operationName);

// here you can do what you want with information you added in your resolvers, just using $contextValue
```

vladar:
@mcg-web But if you use [the server](http://webonyx.github.io/graphql-php/executing-queries/#using-server) it requires a bit more efforts to achieve the same result, so it's not always that obvious. 

But yeah, I still hesitate too. The other thing that bothers me is that reference implementation doesn't have this feature too, and they may add it in future in a conflicting manner, which will indeed make the maintance harder for us.

aelnagger:
Hey folks,
I'd love to help push this request along, but I'm not certain how. I see a lot of concerns about undue maintenance burden, but am not familiar enough with how graphql-php maintenance is performed to make helpful contributions here. Can someone elucidate the main concerns for me? Is this a matter of concern in attempting to maintain API parity with the [reference implementation](https://github.com/graphql/graphql-js) which doesn't exist yet, or is there something more subtle that I don't understand at play here?

mcg-web:
The JS implementation is the reference, i tomorrow their add a feature doing the same thing but a different way this feature will become a legacy that and not so easy to removing it without adding a useless BC

mcg-web:
Sorry my phone keyboard disapair at the wrong moment, i ll finish my explanation on my computer

mcg-web:
I updated my comment an I will soon add a poke to show your we could move to that without loosing your work @aelnagger but keeping the lib public api the same :+1:

aelnagger:
Sounds good @mcg-web I'm looking forward to it!

mcg-web:
Hi, I push a proposal that introduce extensions handler, to give user some more power on extension without touching the main api it's a concept totally aside like the error handler... tell me what you think of it please?

vladar:
I tend to close this PR because it leads us to the wrong direction. We shouldn't encourage using of context in such a manner on a library level - it will be way harder to "undo" later then not doing at all.

Instead, I think that we should follow the other route altogether, which I outlined in #193

Eventually, it will allow us to do what this PR suggests and much more (like optional resolver memoization).

In the meantime, the simple temporary workaround could be adding one more option to the standard server prior to result serialization where users will be able to modify extensions as needed.

mcg-web:
This look like a cleaner way of doing that and it offer more possibility in the future. :+1: for me anyway.

adri:
One concept for extensions that other GraphQL implementations adopted is the use of middlewares. E.g. see Absinthe (Elixir GraphQL Library) https://hexdocs.pm/absinthe/Absinthe.Middleware.html#content

Implementing e.g. tracing becomes then quite simple and encapsulated:
https://github.com/sikanhe/apollo-tracing-elixir/blob/master/lib/apollo_tracing/middleware.ex

Would a middleware layer be an idea? It can be used for other functionalities like access control as well.

vladar:
@adri Sorry I've missed your comment in this thread. I like the middleware approach but we will likely keep following the reference implementation. So they must implement it first for us to follow.

Issue #120, opened 2017-05-18
Example using Zend Expressive


benjamingb:
Do you know of an example that works with Zend Expressive? He tried, but it works

vladar:
None that I am aware of

vladar:
As of version 0.10.0 there is a PSR-7 compatible [standard server](http://webonyx.github.io/graphql-php/executing-queries/#using-server). It should be possible to use it as is with Zend Expressive.

PowerKiKi:
FYI, this is the integration I came up with. It could be cleaner, but at least it is simple. The middleware itself look like:

```php
<?php

declare(strict_types=1);

namespace Application\Action;

use Application\Api\Schema;
use GraphQL\Doctrine\DefaultFieldResolver;
use GraphQL\GraphQL;
use GraphQL\Server\StandardServer;
use Interop\Http\ServerMiddleware\DelegateInterface;
use Interop\Http\ServerMiddleware\MiddlewareInterface;
use Psr\Http\Message\ServerRequestInterface;
use Zend\Diactoros\Response\JsonResponse;

class GraphQLAction implements MiddlewareInterface
{
    public function process(ServerRequestInterface $request, DelegateInterface $delegate)
    {
        GraphQL::setDefaultFieldResolver(new DefaultFieldResolver());

        $schema = new Schema();
        $server = new StandardServer([
            'schema' => $schema,
            'queryBatching' => true,
            'debug' => true,
        ]);

        $parsedBody = json_decode($request->getBody()->getContents(), true);
        $request = $request->withParsedBody($parsedBody);
        $response = $server->executePsrRequest($request);

        return new JsonResponse($response);
    }
}
```

Configure it in `config/autoload/dependencies.global.php`:

```php
return [
    'dependencies' => [
        'invokables' => [
            GraphQLAction::class => GraphQLAction::class,
        ],
    ],
];
```

And declare a route in `config/routes.php`:

```php
$app->post('/graphql', Application\Action\GraphQLAction::class, 'graphql');
```

vladar:
@PowerKiKi Why you shouldn't need following lines:

```php
$parsedBody = json_decode($request->getBody()->getContents(), true);
$request = $request->withParsedBody($parsedBody);
```

They seem redundant as server should already do it (if `Content-Type` header in the request is `application/json`). 

Following should work as is (if something doesn't work, then it is probably a bug):
```php
public function process(ServerRequestInterface $request, DelegateInterface $delegate)
{
[...]

PR #202, opened 2017-11-26
Support non pre-parsed PSR-7 request body

PowerKiKi:
Because PSR-7 specification only specify that `getParsedBody()` **may**
return the parsed body for `application/json`, we cannot assume that it
is always the case. So if the value returned parsed body is an empty array,
it means we should try to parse it ourselves (`null` would mean no body at
all according to spec).

With this modification we try to used given parsed body, but fallback on
trying to parse the body if necessary. This leave the door open to custom
implementation of parsing if needed, while making it easier to use out of
the box.

This is in direct response to https://github.com/webonyx/graphql-php/issues/120#issuecomment-346949979

### DIFF
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -499,6 +499,10 @@ class Helper
                         Utils::printSafeJson($bodyParams)
                     );
                 }
+
+                if (empty($bodyParams)) {
+                    $bodyParams = json_decode($request->getBody(), true);
+                }
             } else {
                 $bodyParams = $request->getParsedBody();
 
--- a/tests/Server/StandardServerTest.php
+++ b/tests/Server/StandardServerTest.php
@@ -47,15 +47,17 @@ class StandardServerTest extends TestCase
             'query' => '{f1}'
         ]);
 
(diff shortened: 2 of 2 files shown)


src/Server/Helper.php
```diff
@@ -499,6 +499,10 @@ public function parsePsrRequest(ServerRequestInterface $request)
                         Utils::printSafeJson($bodyParams)
                     );
                 }
+
+                if (empty($bodyParams)) {
+                    $bodyParams = json_decode($request->getBody(), true);
+                }
```
vladar: Empty is a bit too loose. I guess we should check for null only. Thoughts?

PowerKiKi: This is intended, as mentioned in commit message a null value would mean no body at all:

> A null value indicates the absence of body content.

http://www.php-fig.org/psr/psr-7/#321-psrhttpmessageserverrequestinterface

So we test for an empty array, to try to parse the body ourselves. But if it was null it should throw an exception. IIRC the null case is correctly covered by pre-existing tests.

vladar: Yeah, you are right

PR #204, opened 2017-12-05
inherit resolve and args from the field type

otzy:
When I want lazy loading and define fields as a callback like it's done in documentation:
```
private function MyTypeA()
    {
        return new ObjectType([
            'name' => 'MyTypeA',
            'fields' => function() {
                return [
                    'b' => ['type' => $this->get('MyTypeB')]
                ];
            }
        ]);
    }
```

The field b in MyTypeA looses information about args and resolver. So if you want to resolve, you have to repeate resolve definition.

Suggested commit solves this problem. The field, defined as a callback inherits resolve and args from the type if they are not explicitly overwritten by this callback function.

### DIFF
--- a/src/Type/Definition/FieldDefinition.php
+++ b/src/Type/Definition/FieldDefinition.php
@@ -2,7 +2,6 @@
 namespace GraphQL\Type\Definition;
 use GraphQL\Error\InvariantViolation;
 use GraphQL\Language\AST\FieldDefinitionNode;
-use GraphQL\Language\AST\TypeDefinitionNode;
 use GraphQL\Utils\Utils;
 
 /**
@@ -186,9 +185,14 @@ class FieldDefinition
     {
         $this->name = $config['name'];
         $this->type = $config['type'];
-        $this->resolveFn = isset($config['resolve']) ? $config['resolve'] : null;
+
+	    // with lazy loading we might need to take resolveFn and args from the type
+	    $this->resolveFn = isset($config['resolve']) ? $config['resolve'] : $this->getPropertyFromType('resolve');
+
+	    $args = isset($config['args']) ? $config['args'] : $this->getPropertyFromType('args');
+	    $this->args = is_array($args) ? FieldArgument::createMap($args) : [];
+
         $this->mapFn = isset($config['map']) ? $config['map'] : null;
-        $this->args = isset($config['args']) ? FieldArgument::createMap($config['args']) : [];
 
         $this->description = isset($config['description']) ? $config['description'] : null;
         $this->deprecationReason = isset($config['deprecationReason']) ? $config['deprecationReason'] : null;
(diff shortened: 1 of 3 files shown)


vladar:
Hey, sorry I can't understand what your actual problem is. It would be great if you provide a test case to reproduce it because the solution in this PR doesn't seem right to me. 

I may be wrong, but to clarify this I need a test case that fails on the current version and passes on yours.

otzy:
Hi @vladar,
thanks for your response.

I added a test, that fails on the current version.

The schema is defined in /test/LazySchema.php.

SelfResolvedObject type has a resolve function. And this type is used as a field in Query type, but without resolve specified. My expectation was that if it's not specified for a field, then the 'resolve' function of type should be used to calculate values.

It works actually for normal, not lazy loading. So I think it's ok if lazy loading will have the same behaviour.

vladar:
In short, there is no `resolve` option on the type level, because GraphQL resolves data at the individual field level. It is done on purpose, the way you want to use it is not recommended.

But you can use [`resolveField`](http://webonyx.github.io/graphql-php/data-fetching/#default-field-resolver-per-type) option instead.

otzy:
you are right, it's not really how GraphQL is designed. I was just confused, because the same worked for eager loading.

On the other hand it can be convenient when an object does not depend on the parent in hierarchy. For example you need to render some overal statistics, which you simply calculate with sql query.
Or the object is dependent only on the parent object and does not depend on where it's used in schema. So you don't need to write a resolver every time you use the type (example - https://github.com/graphql/graphql-js/blob/master/src/__tests__/starWarsSchema.js - secretBackstory in Human and Droid)

Or you want to add a field with profiling info and render it in very end.
With type-based resolver you can easily reuse a field in any schema (you can have multiple in the project) and don't need to write resolve function every time you include such field

vladar:
I am really surprised that it worked for eager loading because there is no `resolve` option on the type level regardless of lazy/eager loading mode.

As for this feature in general - it will be very confusing for newbies. Resolving process is the one that causes most questions when starting with GraphQL. We actually want to keep it as simple as possible. Adding more features here is something we want to avoid.

Also, keep in mind that one of the main principles of GraphQL is to resolve only what is needed and the lib is trying to help people start thinking in these terms. Resolving the whole type if a client asks for 2 fields only is against this principle.

otzy:
> I am really surprised that it worked for eager loading because there is no resolve option on the type level regardless of lazy/eager loading mode.
Now I'm testing and it does not work, probably I did something wrong when was learning the library.

>Resolving the whole type if a client asks for 2 fields only is against this principle.
True. Although in real life we have fully instantiated objects in most cases anyway and resolving is just selecting the requested fields from these objects.
Also following this principle makes schema definition a bit verbouse, leads to extra function calls and makes reusing of types more complex.

I'm just starting with GraphQL though, so may be my mind will be changed after a while :)

Issue #203, opened 2017-12-01
Stringable ID types


theofidry:
In [`IDType`](https://github.com/webonyx/graphql-php/blob/master/src/Type/Definition/IDType.php#L46), you are checking that the value is a scalar and return a string value:

```php
    /**
     * @param mixed $value
     * @return string
     */
    public function serialize($value)
    {
        if ($value === true) {
            return 'true';
        }
        if ($value === false) {
            return 'false';
        }
        if ($value === null) {
            return 'null';
        }
        if (!is_scalar($value)) {
            throw new InvariantViolation("ID type cannot represent non scalar value: " . Utils::printSafe($value));
        }
        return (string) $value;
    }
```

This is actually wrong as there is one more kind of value that could be valid: classes implementing `__toString`. For example you could have an ID implementing a `Uuid` class or even something more specific to your domain like `UserId`.

vladar:
For now, it is recommended to use your own custom ID scalar type if you need custom behavior. ID is the only built-in scalar type which can be overridden by passing it via `types` option to the schema.

theofidry:
@vladar I'm using it with OverblogGraphQLBundle so maybe it's an issue there but I wouldn't override the ID type as it's internal and used in `Type::getInternalTypes()` which is static (so even if we can extend type to override it it's not enough). So you would need to register a new type with a different alias and use that new alias, which is too much of a change as it requires to update all the front-ends as well.

We have a dirty workaround for now, but the fix in the core is rather simple...:

```php
    /**
     * @param mixed $value
     * @return string
     */
    public function serialize($value)
    {
        if ($value === true) {
            return 'true';
        }
        if ($value === false) {
            return 'false';
        }
        if ($value === null) {
            return 'null';
        }
        try {
             return (string) $value;
        } catch (\Throwable $throwable) {
             throw new InvariantViolation("ID type cannot represent non scalar value: " . Utils::printSafe($value));
        }
    }
```

vladar:
OK, I'll reopen.

This lib's minimum PHP version is 5.6+ which has no Throwable support. Also, it will silently convert arrays to "Array" string which is not desirable (although I guess we could ignore this case).

So the only good way I see how this can be fixed is by additionally checking for:
```php
!(is_scalal($value) || (is_object($value) && method_exists($value, '__toString'))
```

Do you see any drawbacks to it?

theofidry:
[...]

PR #210, opened 2017-12-12
Allow objects with __toString in IDType

enumag:
I'm using Uuid objects from ramsey/uuid as IDs and this library complains because it's an object.

If you think I should do it differently (like using my own GraphQL type for uuids instead of ID) please enlighten me.

### DIFF
--- a/tests/Type/ObjectIdStub.php
+++ b/tests/Type/ObjectIdStub.php
@@ -1,0 +1,23 @@
+<?php
+namespace GraphQL\Tests\Type;
+
+class ObjectIdStub
+{
+	/**
+	 * @var int
+	 */
+	private $id;
+
+	/**
+	 * @param int $id
+	 */
+	public function __construct($id)
+	{
+		$this->id = $id;
+	}
+
+	public function __toString()
+	{
+		return (string) $this->id;
+	}
+}
--- a/src/Type/Definition/IDType.php
+++ b/src/Type/Definition/IDType.php
@@ -43,7 +43,7 @@ When expected as an input type, any string (such as `"4"`) or integer
         if ($value === null) {
             return 'null';
         }
-        if (!is_scalar($value)) {
+        if (!is_scalar($value) && (!is_object($value) || !method_exists($value, '__toString'))) {
             throw new InvariantViolation("ID type cannot represent non scalar value: " . Utils::printSafe($value));
         }
         return (string) $value;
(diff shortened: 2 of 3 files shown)


vladar:
Thanks for the PR. This is basically what we discussed in https://github.com/webonyx/graphql-php/issues/203#issuecomment-348707573

I would appreciate if you add couple test cases in https://github.com/webonyx/graphql-php/blob/master/tests/Type/ScalarSerializationTest.php 

Then it will be ready for merging.

enumag:
I can do that. Just not sure where to put that class with __toString which is needed for this test.

Also do you think this is a good idea or a bad one? I'm new to GraphQL so I'm unsure if it's expected for every GraphQL object to have an ID or if it's ok or even preferred to use a custom type for things like UUID (custom type might help with validation for example).

vladar:
Drop it in the same folder for now, just name with a `Stub` suffix. I'll move it later when working on next release.  Note that you may have to `require` it manually in test class - I just don't remember if we have autoloading for test classes.

As for using objects for ids - it should be OK. One problem that this solution doesn't solve though is parsing. Some ID types could be transformed back to object from string (like Relay's global id where you often concat type name and type instance id). 

But this is out of the scope of this PR. For now we should at least support `__toString` casting, so your PR is ok.

enumag:
Ok, I'll add the test. Still do you think I should write an UuidType for GraphQL? I can share it of course if I write it.

vladar:
If you use it for entity identification then `ID` is what you need. But if you rely on the fact that it is UUID anywhere in your code, then you should have a separate tab.

As a rule of thumb - if you can safely replace your ids with any other unique string without breaking clients - then it qualifies as `ID` type. Otherwise use custom scalars.

Also in a future version of GraphQL we can see this feature - https://github.com/graphql/graphql-js/pull/914 which can make custom scalars more convenient.

enumag:
Test added.

Your answer is a bit unhelpful... I use UUIDs as entity identification obviously AND rely on it being a valid UUID - if the user sends a string that is not a valid UUID it will cause an error.

I can safely replace the ids with other strings. Or to be more exact with other unique and valid UUIDs. The clients wont break if I change the IDs but they too will rely on them being UUIDs.

vladar:
Thanks! Released in v0.11.5

vladar:
I wouldn't use separate UUID type in your case.

enumag:
Alright, thank you!

tests/Type/ObjectIdStub.php
```diff
@@ -0,0 +1,23 @@
+<?php
+namespace GraphQL\Tests\Type;
+
+class ObjectIdStub
+{
+	/**
```
theofidry: is it expected to have tabs here?

theofidry:
Cheers 👍

enumag:
@theofidry Damn. No of course not. I'm using tabs in my project so IDE automatically used them here... I'll fix it in another PR.

PR #218, opened 2017-12-21
Drop support non pre-parsed PSR-7 request body

PowerKiKi:
I've realized that PR #202 was probably a mistake and should be reversed. The fact that `$request->getParsedBody()` might be empty was due to an incorrect use of Zend Expressive. And I now think that it is not, and should not be, the responsibility of `graphql-php` to parse a PSR-7 request.

I've since [updated my code](https://github.com/Ecodev/dilps/commit/83b7b8c55d1c2d1b604eeb842e48cc8b581aa5f4) to properly configure Zend Expressive to parse the request with [BodyParamsMiddleware](https://docs.zendframework.com/zend-expressive/features/helpers/body-parse/) before reaching `graphql-php`.

If we don't merge this PR, then suddenly `graphql-php` will have to re-implement all kind of request parsing method (i'm thinking `multipart/form-data`) just in case the request was not parsed beforehand. And that would probably not be something we would like to have in that project. Instead it should live in a different project specialized in request parsing.

### DIFF
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -493,19 +493,16 @@ class Helper
                     );
                 }
 
-                // Try parsing ourselves if PSR-7 implementation doesn't parse JSON automatically
-                if (is_array($bodyParams) && empty($bodyParams)) {
-                    $bodyParams = json_decode($request->getBody(), true);
-
-                    if (json_last_error()) {
-                        throw new RequestError("Could not parse JSON: " . json_last_error_msg());
-                    }
-                }
-
                 if (!is_array($bodyParams)) {
                     throw new RequestError(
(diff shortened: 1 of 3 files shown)


vladar:
It introduces a possible breaking change, so we'll have to release it in the next major version. Will merge it as soon as I start working on it (unless you need it for your other PRs).

PowerKiKi:
I will most likely need it for the next PR, but let's wait and re-consider it when I have something more to share...

vladar:
So when we merge uploads this PR will make no sense, right?

PowerKiKi:
The upload PR does include this PR's commit too, so yes if upload PR is merged, then this one becomes useless. But it could also be merged separately... depending on how long it takes for upload PR to be merged...

Issue #178, opened 2017-10-05
Uploads possible?


crirus:
Hello

I am wondering if file upload is possible via graphQL

Thank you

vladar:
Not directly. Usually, they move uploads to separate endpoint with progress reporting, etc.

But you should be able to integrate it with GraphQL mutations as well, using existing PHP tools like $_FILES in your mutation resolver. 

If someone has experience with file uploads + GraphQL, please share.

PowerKiKi:
@vladar I will have to implement file upload for my projects. Would you be interested in getting a PR to be merged here ? or should it be part of another project ?

Based on what I found, I believe we could implement something generic enough to to be usable by implementing the [graphql-multipart-request-spec](https://github.com/jaydenseric/graphql-multipart-request-spec). From what I gathered so far, it would be a new `Upload` scalar type that expose a few properties to the resolver such as `filename` and `mimetype`. IMHO this would be "low-level" enough to be included in `graphql-php`, but if you disagree I'll keep that separate. What do you think ?

caseycs:
@vladar I tried to make it work on the front-end side - but that was quite challenging, even while there are some apollo js components available on GitHub. I decided to use two-step approach for now - "legacy" ajax image upload to temporary storage returns uuid, and then I use this uuid in graphql.

PowerKiKi:
@caseycs did you try with [apollo-upload-server](https://github.com/jaydenseric/apollo-upload-server) ? If so would you be able to sump up the difficulties you had ?

caseycs:
@PowerKiKi I have PHP on the back-end, so was experimenting only with front-end part.

PowerKiKi:
Oh my bad, I meant `apollo-upload-client` of course. What difficulties did you have ?

vladar:
I am curious what problems do you guys have with uploads (apart from lack of "multipart/form-data" support in the StandardServer)?

@PowerKiKi since PHP supports file uploads natively, we could do it too, but would you be ready to maintain this type / code in the future? %) If yes, I'll accept your PR

crirus:
Well, I just added that enctype to StandardServer.
I send the files as a normal post upload and mention in graphql a field with a Input  type that validates the file and store it.

Something like this.
```php
$config = [
            'name' => 'UploadInput',
            'description' => 'Provide field name of the file input or URl to fetch the file from. If name is provided, a file must be uploaded',
            'fields' => [
                'field_name' => Types::string(),
            ],
[...]

PR #220, opened 2017-12-22
[WIP] Support for file uploads

PowerKiKi:
This is a work in progress to implement [GraphQL multipart request specification](https://github.com/jaydenseric/graphql-multipart-request-spec) to support file uploads. 

The `StandardServer` now supports file uploads for request with `multipart/form-data` content type, but only for PSR-7 requests (IMHO it would be out of scope to parse raw request for uploaded files). A new type `File` was introduced to represent uploaded file in the schema.

Usage should look like:

```php

$request = getPsr7RequestFromAnotherLibrary();

$schema = new Schema([
    'query' => new ObjectType([
        'name' => 'Query',
    ]),
    'mutation' => new ObjectType([
        'name' => 'Mutation',
        'fields' => [
            'testUpload' => [
                'type' => Type::string(),
                'args' => [
                    'text' => Type::string(),
                    'file' => Type::file(),
                ],
                'resolve' => function ($root, array $args) {

                    /** @var UploadedFileInterface $file */
                    $file = $args['file'];
                    $this->assertInstanceOf(UploadedFileInterface::class, $file);

                    // Do something more interesting with the file
                    // $file->moveTo('some/folder/in/my/project');

                    return 'Uploaded file was ' . $file->getClientFilename() . ' (' . $file->getClientMediaType() . ') with description: ' . $args['text'];
                },
            ],

        ],
    ]),
]);

$server = new StandardServer([
    'schema' => $schema,
]);

$response = $server->executePsrRequest($request);

```

On the implementation side of things, I would need advice on how to properly test `\GraphQL\Type\Definition\FileType`. For now we have `\GraphQL\Tests\Executor\UploadTest` which is more or less a functionnal tests for the entire stack. It's nice and working, but it only test a single case. It might be best to have unit tests covering only `FileType`, with more different cases, but I could not find anything similar. Would you have any recommendations about that ? where and how it should be done ?
 
Like I mentionned it includes PR #218, because at this point it would probably not be wise to parse uploaded file ourselves. But I suppose I could rebase without #218 if that's a problem for you.
 
It should be a solution for #178 (and https://github.com/Folkloreatelier/laravel-graphql/issues/125)

I'll write some docs, once we agree on the code.

### DIFF
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -482,34 +482,22 @@ class Helper
                 throw new RequestError('Missing "Content-Type" header');
             }
 
-            if (stripos('application/graphql', $contentType[0]) !== false) {
+            $contentType = $contentType[0];
+            if (stripos($contentType, 'application/graphql') !== false) {
                 $bodyParams = ['query' => $request->getBody()->getContents()];
-            } else if (stripos('application/json', $contentType[0]) !== false) {
+            } elseif (stripos($contentType, 'application/json') !== false) {
                 $bodyParams = $request->getParsedBody();
 
-                if (null === $bodyParams) {
-                    throw new InvariantViolation(
-                        "PSR-7 request is expected to provide parsed body for \"application/json\" requests but got null"
-                    );
-                }
-
-                if (!is_array($bodyParams)) {
-                    throw new RequestError(
-                        "GraphQL Server expects JSON object or array, but got " .
-                        Utils::printSafeJson($bodyParams)
-                    );
-                }
-
-                if (empty($bodyParams)) {
-                    throw new InvariantViolation(
-                        "PSR-7 request is expected to provide parsed body for \"application/json\" requests but got empty array"
-                    );
-                }
+                $this->validateParsedBody($bodyParams);
+            } elseif (stripos($contentType, 'multipart/form-data') !== false) {
+                $bodyParams = $request->getParsedBody();
+                $this->validateParsedBody($bodyParams);
+                $bodyParams = $this->parseUploadedFiles($request, $bodyParams);
             } else {
                 $bodyParams = $request->getParsedBody();
 
                 if (!is_array($bodyParams)) {
-                    throw new RequestError("Unexpected content type: " . Utils::printSafeJson($contentType[0]));
+                    throw new RequestError("Unexpected content type: " . Utils::printSafeJson($contentType));
                 }
             }
         }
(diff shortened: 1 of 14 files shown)


crirus:
```php
testUpload' => [
                'type' => Type::string(),
```
shouldn't be a sort of Upload type?

vladar:
Looks awesome, but I have couple notes:

1. I would like to separate utils like this from core types defined by GraphQL spec. So we should use something like `Utils\UtilTypes::psrFile()` for this instead of `Type::file()`. Also the type itself should live somewhere under `Utils` namespace (or maybe all of it should be placed under the `Server` namespace - what do you think?).

2. The server has to support "multipart/form-data" for default PHP globals method as well. Right now when you send `multipart/form-data` the server just won't do anything. But it should actually process the query passed in a request the same way it does for regular `application/x-www-form-urlencoded`. Then one can use $_FILES manually in the mutation resolver.<br>
We can probably implement a similar scalar type (e.g. `Utils\UtilTypes::phpFile()`) which just returns corresponding entry from $_FILES in the resolver.

3. I am not convinced that https://github.com/jaydenseric/graphql-multipart-request-spec is a way to go for us because it introduces separate request format which is different from what we currently use for `application/x-www-form-urlencoded`. I'd prefer to stay consistent here. We do not support batched queries for `application/x-www-form-urlencoded` anyways.<br>
We should probably stay closer to express-graphql, e.g. how they [suggest doing uploads](https://github.com/graphql/express-graphql/blob/6d5344a1e7c7e9bc09ce199bfe6665a254344010/src/__tests__/http-test.js#L711-L778) (with a difference that we use scalar type vs object type as in their example)

PowerKiKi:
> separate utils like this from core

I suspected you would say that. I really don't mind at all. I would tend to go `Utils\UtilTypes` rather than mix with server, because they could be used outside of standard server too.

> support "multipart/form-data" for default PHP globals

I am not sure about that. I am actually not sure about the "raw" server at all. I feel supporting raw request (as opposed to PSR request) double our workload for really no gain. It's very easy to add a PSR compliant lib to parse globals just before calling the standard server. Supporting "raw" request is definitely convenient for new users, but I feel like it will bite us in the end, because we have to re-implement lots of PSR related things and maintain two very similar yet different paths in our code (one for raw, and one for PSR).

Instead of implementing `multipart/form-data` for globals, I would take the drastic opposite decision of dropping the raw server entirely.

> not convinced about `graphql-multipart-request-spec`

This might be a deal breaker for me. I use Apollo Client on the browser side and it seemed to me that `apollo-upload-client` was the de-facto standard in that context. It's about 1 year old and rather popular according to GitHub stars and forks. So I can only assume that the spec that came out of that project should reflect all the experience accumulated from real-world experiences. And I did made a proof-of-concept with this PR and `apollo-upload-client` in my project very easily. It's also the only attempt to create a specification about uploads AFAIK.

I am not quite sure, but it seems that what `express-graphql` suggests would not be compatible (at all) with `apollo-upload-client`. And that would be a problem for me. I am willing to be convinced to use another format if there are ready-to-be-used solutions out there for Apollo, but I am not willing to work on a format that I could not use with Apollo on the client side almost out of the box.

What do you use on client side ? would you have a suggestion to combine `express-graphql` and Apollo Client ?

vladar:
We have a separate endpoint for file uploads (not doing it via GraphQL). And also use our own simple GraphQL client (not Apollo or Relay)

I actually didn't realize that you want to implement some 3rd-party spec, just thought we'll mirror `multipart/form-data` after `application/x-www-form-urlencoded` with some sugar (as express-server does via [multer](https://github.com/expressjs/multer)).

As for `graphql-multipart-request-spec`... the problem is that we can't really say that it is a standard. Apollo server by itself [doesn't support file uploads](https://github.com/apollographql/apollo-server/issues/229), Relay recommends doing it [this way](https://github.com/facebook/relay/issues/1844#issuecomment-316893590) (a-la express).

I think it's better to extract this implementation as a separate project (same way as apollo-upload-server is separate from apollo-server). But still, the question remains if we can make life simpler in a Standard Server for those using such tools.

vladar:
Actually apollo-server (at least for express) uses the [same approach](https://github.com/apollographql/apollo-server/blob/7cf0580122408f139fbc36a919bc07fe393ed430/packages/apollo-server-express/src/apolloServerHttp.test.ts#L220-L290) as express-server, so this actually looks like a standard for me.

PowerKiKi:
Fair enough, I'll extract this PR as an independent project for the time being, and we can always re-consider the situation if/when a standard more clearly emerges.

PowerKiKi:
Just released [graphql-upload](https://github.com/Ecodev/graphql-upload) 1.0.0 for anyone interested...

PR #222, opened 2017-12-30
Exclude files in release

giansalex:
- Package size: 4.3Mb -> 650kb
- Package result
captura

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,16 +1,12 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text eol=lf
 /benchmarks export-ignore
-/docs export-ignore
-/examples export-ignore
 /tests export-ignore
 /tools export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
 .travis.yml export-ignore
-CHANGELOG.md export-ignore
 CONTRIBUTING.md export-ignore
 mkdocs.yml export-ignore
 phpbench.json export-ignore
 phpunit.xml.dist  export-ignore
-UPGRADE.md  export-ignore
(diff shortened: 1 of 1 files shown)


PowerKiKi:
I am not saying you should or shouldn't merge this, but you might be interested in [having a look a current practice](https://github.com/PHPOffice/PhpSpreadsheet/pull/139#issuecomment-294277982) before merging. You might also consider that Composer itself [stopped recommending the practice](https://github.com/composer/composer/commit/bf328697e1ae7d290fb17df2e902a652ec4d9378) to exclude things.

IMHO I would definitely not exclude any docs... and tests might be considered a form of documentation depending on the project...

vladar:
@PowerKiKi I do agree that we should keep docs and examples (as they are rather small), but everything else can be dropped (I've seen codebases where vendor folder grows > 1gb which is insane). 

So I think following entries could be safely deleted:
```
/benchmarks export-ignore
/tests export-ignore
/tools export-ignore
.gitattributes export-ignore
.gitignore export-ignore
.travis.yml export-ignore
CONTRIBUTING.md export-ignore
mkdocs.yml export-ignore
phpbench.json export-ignore
phpunit.xml.dist  export-ignore
```
@giansalex  Can you upgrade the PR and I'll merge it?

giansalex:
I think that if someone wants to see the documentation, they should go to https://webonyx.github.io/graphql-php/

vladar:
Docs are quite useful at least for offline mode. Some people also like doing docs search straight in their IDEs, so there are valid cases for docs in a release. Plus docs and examples take ~200kb which is tolerable.

Issue #219, opened 2017-12-21
Use of `setExpectedException()` in tests ?


PowerKiKi:
Would you accept a PR to relace all usage of `try/catch` with the more concise `$this->setExpectedException()` ?

For instance:
```php
    public function testFailsWithMissingContentType()
    {
        try {
            $this->parseRawRequest(null, 'test');
            $this->fail('Expected exception not thrown');
        } catch (RequestError $e) {
            $this->assertEquals('Missing "Content-Type" header', $e->getMessage());
        }

        try {
            $this->parsePsrRequest(null, 'test');
            $this->fail('Expected exception not thrown');
        } catch (RequestError $e) {
            $this->assertEquals('Missing "Content-Type" header', $e->getMessage());
        }
    }
```
would then become:
```php
    public function testFailsWithMissingContentTypeRaw()
    {
        $this->setExpectedException(RequestError::class, 'Missing "Content-Type" header');
        $this->parseRawRequest(null, 'test');
    }

    public function testFailsWithMissingContentTypePsr()
    {
        $this->setExpectedException(RequestError::class, 'Missing "Content-Type" header');
        $this->parsePsrRequest(null, 'test');
    }
```

There is quite a bunch of them, so I'd rather ask before doing it...

vladar:
Yeah, will gladly accept such PR!

PR #224, opened 2018-01-01
Leverage PHPUnit `setExpectedException()` instead of custom code

PowerKiKi:
As discussed in #219


vladar:
Thanks a lot!

P.S. Looks like Travis experiences some troubles now, I'll merge it as soon as Travis recovers and we get it built correctly.

PR #232, opened 2018-01-21
Fix Standard Server PSR request parsing

Poky85:
Sending PSR request to GraphQL Standard server doesn't work.

```php
/** @var ExecutionResult|ExecutionResult[] $result */
$result = $server->executePsrRequest($psrRequest);
```

This ends with following error.

> GraphQL Request must include at least one of those two parameters: "query" or "queryId".

Caused by bug in request's content-type matching.


vladar:
Oops, good catch, thanks!

PR #234, opened 2018-02-01
Some fast fixes

mcg-web:


mcg-web:
ok didn't know about those tools, i will remove this commit :+1:

PR #236, opened 2018-02-02
Add missing params for promiseToExecute calls

dimitri-koenig:
The GraphQL\GraphQL class contains some calls to promiseToExecute() without all parameters, like $fieldResolver and $validationRules. This PR adds those missing params

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -152,7 +152,8 @@ class GraphQL
                     $context,
                     $variableValues,
                     $operationName,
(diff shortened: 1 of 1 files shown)


mcg-web:
hi, why do you want to change a deprecated method signature? It would not be easier to use directly new methods?

dimitri-koenig:
That's a really good question. Sorry, didn't see that. Will change the call in another package which depends on this one. Thx anyway

PR #237, opened 2018-02-08
[WIP][0.13] Separate multiple inherited interfaces with `&`

danez:
> This replaces:
> 
> ```graphql
> type Foo implements Bar, Baz { field: Type }
> ```
> 
> With:
> 
> ```graphql
> type Foo implements Bar & Baz { field: Type }
> ```
> 
> With no changes to the common case of implementing a single interface.
> 
> This is a breaking change for existing uses of multiple inheritence. To allow for an adaptive migration, this adds a parse option to continue to support the existing experimental SDL: `parse(source, {allowLegacySDLImplementsInterfaces: true})`

Ref: graphql/graphql-js#1169

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -696,6 +696,14 @@ Parses string containing GraphQL query or [type definition](type-system/type-lan
  * (By default, the parser creates AST nodes that know the location
  * in the source that they correspond to. This configuration flag
  * disables that behavior for performance or testing.)
+ * 
+ * allowLegacySDLImplementsInterfaces: boolean,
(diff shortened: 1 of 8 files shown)


vladar:
As far as I understand this PR doesn't make sense as a standalone PR anymore? Or am I missing something?

danez:
This is not included in the other big PR as the other PR was only adding stuff up to 0.12.3, and this is part of 0.13. But i guess i will create a new PR at some point with all the 0.13 changes.

vladar:
I already ported this as a part of the work on v0.13

PR #251, opened 2018-02-16
Fix node list iterator to check for offsets before accessing it becau…

cpartica:
Fix node list iterator to check for offsets before accessing it because offsets can be unset by a public method of this class per \ArrayAccess interface

### DIFF
--- a/src/Language/AST/NodeList.php
+++ b/src/Language/AST/NodeList.php
@@ -107,7 +107,9 @@ class NodeList implements \ArrayAccess, \IteratorAggregate, \Countable
     {
         $count = count($this->nodes);
         for ($i = 0; $i < $count; $i++) {
-            yield $this->offsetGet($i);
+            if ($this->offsetExists($i)) {
+                yield $this->offsetGet($i);
+            }
         }
     }
 


vladar:
Thanks for the PR! One problem with selected implementation is that I'd really want to avoid another function call  (offsetExists) there: this method is used quite often, so it is critical to keep the number of function calls there as low as possible.

I guess we could change the implementation a bit:
1. Just use `foreach` loop vs for (it will account for missing keys much more efficiently)
2. Do not use `unset` in `offsetUnset` and instead, splice an array (to always have sequential keys). Newer versions of PHP have optimizations in place for such arrays
3. Make NodeList immutable (prohibit unset and set) and just require a new NodeList in such situations (but this is a BC break)

As far as I understand you do use unset for some of your code. Can you share your thoughts on those options?

spawnia:
Hey @vladar, to understand this better: What is the reasoning behind using NodeList in the first place?

vladar:
@spawnia We needed it for lazy initialization of AST object tree from an array. Without NodeList you'll have to traverse such array twice: the first time when initializing AST from array, the second time when you actually visit it during validation or schema creation.

You may need it in couple cases:

1. When you use [C-based parser](https://github.com/graphql/libgraphqlparser)
2. When you serialize and cache AST as an array (especially big schema language AST)

NodeList enables us to create a Node on the first visit (thus only visiting it once).

nagledb:
Running this:
```php
$data = ['a', 'b', 'c', 'd', 'e'];
unset($data[2]);
$count = count($data);
for($i = 0; $i < $count; $i++) {
    echo $data[$i] . "\n";
}
```

Results in this:
```
a
b
PHP Notice:  Undefined offset: 2 in test.php on line 30

d
```

Note that not only does it trigger an error notice when it tries to access the second element, but also it doesn't access the last item because there are only 4 items but the last item has index 4. The code in `NodeList::getIterator` has the same two issues.

This PR fixes the first issue, but it doesn't address the second. Using `foreach` as in option 1 would solve both issues.

As for option 2:

> Do not use unset in offsetUnset and instead, splice an array (to always have sequential keys).

That would actually be a breaking change. Let's pretend we have `$nodes` that contains a `NodeList` with 5 nodes whose names are "a", "b", "c", "d", and "e". If we ran this code:
```php
$count = count($nodes);
for($i = 0; $i < $count; $i++) {
    if ($i == 1) {
        unset($nodes[$i]);
    } else {
        echo $nodes[$i]->name->value . "\n";
    }
}
```
Then we'd see this:
```
a
d
e

PHP Notice:  Undefined offset: 4 in test.php on line 30
```

I would find that behavior very unexpected and arguably it even violates the spirit of `ArrayAccess` since it would be behaving very differently than a normal array would.

Issue #201, opened 2017-11-25
Descriptions lost in BuildSchema when used cached AST array.


szonov:
I’m using last example from http://webonyx.github.io/graphql-php/type-system/type-language/

```
if (!file_exists($cacheFilename)) {
    $document = Parser::parse(file_get_contents('./schema.graphql'));
    file_put_contents($cacheFilename, "<?php\nreturn " . var_export(AST::toArray($document), true));
} else {
    $document = AST::fromArray(require $cacheFilename); // fromArray() is a lazy operation as well
}
```

When there is no cache file yet - graphiql shows descriptions defined in schema.graphql,
But second request (when cache file already exists) shows all fields without description.

I’ve made quick research and found that in BuildSchema::getDescription on second request node already contain ‘description’ and next parsing not needed.

I propose made simple check in begin of this function

```
if (property_exists($node, 'description'))
    return $node->description;
```

Torsten85:
I did stumble over the same problem. Your proposed quick fix works great for me. Do you want to make a pull request for that?

vladar:
Feel free to submit a PR, I'll merge it.

Poky85:
@szonov Thanks for suggested solution. There have been no PR for a long time, so I created mine in #253.

vladar:
This issue should be fixed as a part of #248 (with new way of defining descriptions in SDL)

PR #253, opened 2018-02-19
Fix lost description in BuildSchema

Poky85:
Right now description get lost when calling:

```php
$source = AST::fromArray($array);
$schema = BuildSchema::build($source);
```

Mentioned in issues #201 and #88.


vladar:
Looks like this PR breaks some tests, can you please double check?

PR #256, opened 2018-03-03
Parse operation name from operationName instead of operation

camuthig:
The operation name in `OperationParams` is attempting to pull from `operation` instead of `operationName`. Because of this, applications leveraging the `StandardServer` cannot handle a request with multiple operations (it returns an error saying the operation name is required).

In this PR, I only updated the way the class parses the data on the `create` function, to avoid changing the API. It might be good, however, if you are okay with it, to mark `operation` as deprecated and add a new `operationName` property on the class instead, such that `operation` can be removed on the next major version bump.

### DIFF
--- a/src/Server/OperationParams.php
+++ b/src/Server/OperationParams.php
@@ -67,7 +67,7 @@ class OperationParams
             'queryid' => null,
             'documentid' => null, // alias to queryid
             'id' => null, // alias to queryid
-            'operation' => null,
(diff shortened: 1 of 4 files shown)


vladar:
Can you provide a test case which fails on the previous version and works on this one? It will help to catch regressions later and to make sure we are on the same page with this issue.

vladar:
Thanks!

Issue #255, opened 2018-02-28
Circular reference handled via proxies


simPod:
I'm injecting types via DI constructor. When schema is built, I'm getting this error:

`Schema must contain unique named types but contains multiple types named "CircularType" (see http://webonyx.github.io/graphql-php/type-system/#type-registry).`

That is because I'm using proxies https://github.com/Ocramius/ProxyManager to get actual services so I avoid circular reference.

```php
class TypeOne
{
    public function __construct(TypeTwo $typeTwo)
    {
        // ...
    }
}

class TypeTwo
{
    public function __construct(TypeOne $typeOne)
    {
        // ...
    }
}
```

However the test for duplicate types is `$typeMap[$type->name] === $type` and that's always `false` when using proxies even though the wrapped type is the same instance.

How to handle this properly? My only solution so far is to fork this and remove that check

```php
        if (!empty($typeMap[$type->name])) {
            Utils::invariant(
                $typeMap[$type->name] === $type,
                "Schema must contain unique named types but contains multiple types named \"$type\" ".
                "(see http://webonyx.github.io/graphql-php/type-system/#type-registry)."
            );
            return $typeMap;
        }
```

simPod:
The similar check is on Executor.php:942

vladar:
Do you use proxies for circular types only? Have you seen this [article in docs](http://webonyx.github.io/graphql-php/type-system/object-types/#recurring-and-circular-types) and why it won't work for you?

simPod:
Yup, only for circular types. I use dependency injection for types that I autowire and not using static (global) types which I consider to be a cleaner way. 

But I might have found a way -> opening PR as I believe it's not working properly in this library.

vladar:
I am curious why your proxy manager (or DI container) creates multiple instances for the same underlying type? This seems wrong to me. It leads to subtle errors and may cause significant performance overhead (as basically, you have `{number of fields of this type}` objects per type vs 1) unless I am missing something.

simPod:
The ocramius' https://github.com/Ocramius/ProxyManager creates multiple proxy instances with the same instance of underlying type within. Therefore, it doesn't create circular references when using DI. However I think this approach is no go with this library as it would require adding a new dependency.

Instead I started using callables those when invoked return the actual type instance. This way I can avoid circular references and still use DI. (https://github.com/webonyx/graphql-php/pull/257#issuecomment-373685194)

vladar:
Closing after conversation in #257

PR #257, opened 2018-03-04
Allow callable types

simPod:
This allows to define the type as callable that is invoked later. I see that it is partially implemented eg. https://github.com/webonyx/graphql-php/blob/master/src/Type/Definition/NonNull.php#L19


vladar:
This functionality existed in one of the first versions of the library but was deprecated in favor of callables in `fields` (and yes, some traces of it still exist in the code). It caused more troubles (with maintenance and performance) than gave benefits. I don't see any reason why we should restore it.

You are the first who needs it in 2.5 years of library existence for a quite specific reason (trying to avoid registry/factory of types). But I'll keep this PR open to collect more evidence that this feature is needed by other library users as well.

simPod:
I consider callables and their __invoke as something great when you want something to be lazy / performance sensitive.

Not sure why I'm first tho. I guess it's because people gave up on circular schemas (just a thought 🤔 ) 

I have a registry of types. Created a compiler pass that takes all type services from DI container and puts them into registry from where I can request them. 

That doesn't solve the problem tho. When there's a type A in a registry that requires type B as a field's type and also there's type B that requires type A as its field's type, there's no other way how to create that. Type registry have no role it this case.

The only way to solve it (I might not have found 'em all) is to have eg. type A that has field with type B within a callback so it doesn't reference the type B directly. And vice versa. 
I, for example, set the type B of A type field as:

```php
Type::listOf(
        function () : Type {
            return $this->typeRegistry->getTypeB();
        }
)
```
So type B is resolved only when needed and the infinite circular reference cannot happen.

vladar:
I've [posted you a link](https://github.com/webonyx/graphql-php/issues/255#issuecomment-370143586) on how to deal with circular types.

```php
class User extends ObjectType
{
    public function __construct(TypeRegistry $types)
    {
        parent::__construct([
            'fields' => function() use ($types) {
                return [
                    'lastPost' => [
                        'type' => $types->post(),
                    ]
                ];
            }
        ]);
    }
}


class BlogPost extends ObjectType
{
    public function __construct(TypeRegistry $types)
    {
        parent::__construct([
            'fields' => function() use ($types) {
                return [
                    'author' => [
                        'type' => $types->user(),
                    ]
                ];
            }
        ]);
    }
}
```

Why doesn't this work for you?

simPod:
Because TypeRegistry would look like this:

```php

class TypeRegistry
{
    private $user;

    private $blogPost;

    public function user()
    {
        return $this->user ?: ($this->user = new User($this));
    }

    public function post()
    {
        return $this->blogPost ?: ($this->blogPost = new BlogPost($this));
    }
}
```

So then it goes like this

1. User is instantiated from TypeRegistry (`new User` called)
2. Withing `User` `__construct` `$types->post()` is called
3. In `$types->post()` `new BlogPost` is called
4. Withing `BlogPost` `__construct` `$types->user()` is called
5. It all repeats over and over

vladar:
```php
2. Withing User __construct $types->post() is called
```
No, it is not called within constructor, because `fields` is a closure. It is called later during execution. The whole point to make `fields` (and `interfaces`) a closure to break a cycle.

simPod:
Oh, missed that one. Thanks, that works. Going to refactor my schema a bit but it seems to be a way and this PR is not necessary then I guess.

PR #260, opened 2018-03-09
Drop support for EOL PHP versions

simPod:
Support only current versions of php so we can leverage modern language features

http://php.net/supported-versions.php

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,12 @@
 language: php
 
-# Required for HHVM, see https://github.com/travis-ci/travis-ci/issues/7712
 dist: trusty
 
 php:
-    - 5.5
     - 5.6
     - 7.0
     - 7.1
-    - hhvm
+    - 7.2
     - nightly
 
 matrix:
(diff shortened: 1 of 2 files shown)


vladar:
We will support PHP5.6 at least until the end of the year (when PHP security support will end). But we can drop PHP5.4 and 5.5. Please adjust your PR and I'll merge it.

simPod:
And don't we at least wanna lift the drop to 5.6 as well so at least 7.0 is required and we can start strict-typing things?

vladar:
@simPod not really. At least not yet. I think we should do a prior notice before doing such major things. So we will release one more version as 5.6 and the next major version will only support PHP 7.0+

simPod:
@vladar I must admit I really liked how Doctrine project did that http://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html and especially _Why dropping PHP support in a minor version is not a BC break_ section there. But it's up to you, will add 5.6 tomorrow.

vladar:
But even in Doctrine announcement, they write: "Since many people are encountering issues with these updates..."

I wouldn't call this a good approach to deprecations/upgrades.

simPod:
Yup but you took it out of context. They wrote so with intention to help inexperienced developers to cleanup their composer config. They didn't mean that it's a bad approach because it is not. 

The only problem is that people often use `dev-master` as a dependency constraint, run `composer update` or do similar dumb things but that is not your responsibility as a dependency maintainer.

In my opinion it draws back progress and inovations when library tries to satisfy users with unconstrained dependencies but it's only my sole and humble opinion, no need to care about it ;) We can proceed with the merge, it's step forward anyway.

vladar:
Thanks!

PR #262, opened 2018-03-21
Resolve ambiguous `null` interpretation with regards to `defaultValue`s

samsouder:
There is some ambiguity in both the GraphQL spec and various implementations with regards to when `defaultValue`s are applied to arguments.

A [spec change to resolve this](https://github.com/facebook/graphql/pull/418) is being considered and there is a proposed [PR for the JS implementation](https://github.com/graphql/graphql-js/pull/1274) as well.

I have run into this during use of graphql-php where I wanted `defaultValue` to be applied for an optional argument (which can't be `null` either), but the client used sends `null` as the value for the argument. I expected the `defaultValue` to be substituted, but `null` was the coerced value instead.

This PR tries to apply a similar fix for the ambiguity in the above PR for the JS implementation, always using the `defaultValue` for a node if the value is explicitly sent as `null` (either directly or via variables).

### DIFF
--- a/tests/Executor/VariablesTest.php
+++ b/tests/Executor/VariablesTest.php
@@ -853,6 +853,36 @@ class VariablesTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @it when null value variable provided
+     */
+    public function testWhenNullValueVariableProvided()
+    {
+        $ast = Parser::parse('query optionalVariable($optional: String) {
+            fieldWithDefaultArgumentValue(input: $optional)
+        }');
+
+        $this->assertEquals(
+            ['data' => ['fieldWithDefaultArgumentValue' => '"Hello World"']],
+            Executor::execute($this->schema(), $ast, null, null, ['optional' => null])->toArray()
+        );
+    }
+
+    /**
+     * @it when null value provided directly
+     */
+    public function testWhenNullValueProvidedDirectly()
+    {
+        $ast = Parser::parse('query {
+            fieldWithDefaultArgumentValue(input: null)
+        }');
+
+        $this->assertEquals(
+            ['data' => ['fieldWithDefaultArgumentValue' => '"Hello World"']],
+            Executor::execute($this->schema(), $ast)->toArray()
+        );
+    }
+
+    /**
      * @it not when argument cannot be coerced
      */
     public function testNotWhenArgumentCannotBeCoerced()
--- a/src/Executor/Values.php
+++ b/src/Executor/Values.php
@@ -12,6 +12,7 @@ use GraphQL\Language\AST\FieldNode;
 use GraphQL\Language\AST\FragmentSpreadNode;
 use GraphQL\Language\AST\InlineFragmentNode;
 use GraphQL\Language\AST\NodeList;
+use GraphQL\Language\AST\NullValueNode;
 use GraphQL\Language\AST\VariableNode;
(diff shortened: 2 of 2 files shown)


vladar:
I think we should wait for a fix in the reference implementation.

nagledb:
The reference implementation merged the referenced PR in April.

simPod:
Needs rebase tho

SevInf:
Also, i think spec and reference implementation changed since this PR.  Explicit "null" don't sets value to default anymore.

tests/Executor/VariablesTest.php
```diff
@@ -852,6 +852,36 @@ public function testWhenOmittedVariableProvided()
...
+        $this->assertEquals(
+            ['data' => ['fieldWithDefaultArgumentValue' => '"Hello World"']],
+            Executor::execute($this->schema(), $ast, null, null, ['optional' => null])->toArray()
+        );
+    }
+
+    /**
+     * @it when null value provided directly
+     */
+    public function testWhenNullValueProvidedDirectly()
```
simPod:
```suggestion
    public function testWhenNullValueProvidedDirectly() : void
```

tests/Executor/VariablesTest.php
```diff
@@ -852,6 +852,36 @@ public function testWhenOmittedVariableProvided()
         );
     }
 
+    /**
+     * @it when null value variable provided
+     */
+    public function testWhenNullValueVariableProvided()
```
simPod:
```suggestion
    public function testWhenNullValueVariableProvided() : void
```

tests/Executor/VariablesTest.php
```diff
@@ -852,6 +852,36 @@ public function testWhenOmittedVariableProvided()
...
+    {
+        $ast = Parser::parse('query optionalVariable($optional: String) {
+            fieldWithDefaultArgumentValue(input: $optional)
+        }');
+
+        $this->assertEquals(
```
simPod:
```suggestion
        self::assertEquals(
```

tests/Executor/VariablesTest.php
```diff
@@ -852,6 +852,36 @@ public function testWhenOmittedVariableProvided()
...
+    /**
+     * @it when null value provided directly
+     */
+    public function testWhenNullValueProvidedDirectly()
+    {
+        $ast = Parser::parse('query {
+            fieldWithDefaultArgumentValue(input: null)
+        }');
+
+        $this->assertEquals(
```
simPod:
```suggestion
        self::assertEquals(
```

spawnia:
This looks like an important bugfix, especially considering that the reference implementation is now also fixed.

The spec definitely says that explicit null is not the same as omitting the argument: https://graphql.github.io/graphql-spec/draft/#sec-Null-Value

samsouder:
I no longer have interest in fixing this specific issue to ensure it matches the current reference implementation.

spawnia:
@vladar i would be willing to champion this and create a new PR, ok?

vladar:
@spawnia Yeah, as long as it matches reference implementation and spec, sure.

spawnia:
Fixed in https://github.com/webonyx/graphql-php/commit/bf4e7d4a9f078fb926f92a32725a6ccf0c0e09a2

@vladar Thanks for tackling this important issue. You are doing a great job with this library!

vladar:
@spawnia thanks %)

PR #267, opened 2018-04-05
Fix missing sprintf in Schema.php

ruudk:
```
Wrong parameters for GraphQL\\Error\\InvariantViolation([string $message [, long $code [, Throwable $previous = NULL]]])
```


vladar:
Thanks!

PR #277, opened 2018-04-24
Make Types throw instead of returning Utils::undefined()

danez:
I finally had time and found a way for all the cases, see inline comments


src/Type/Definition/BooleanType.php
```diff
@@ -32,22 +34,30 @@ public function serialize($value)
...
      */
     public function parseLiteral($valueNode, array $variables = null)
     {
         if ($valueNode instanceof BooleanValueNode) {
             return (bool) $valueNode->value;
         }
-        return Utils::undefined();
+
+        // Intentionally without message, as all information already in wrapped Exception
+        throw new \Exception();
```
danez: It throws an empty exception as otherwise the error messages do not match anymore with the expected error messages copied from graphql-js. Not super nice, but the other solution would have been updating all tests and diverging from graphql-js in that sense.

vladar:
We will probably have to deviate from the reference implementation eventually when defining scalar types. I need some time to consider if chosen path is a correct one as both solutions (undefined vs throw) have their pros and cons.

Issue #280, opened 2018-05-07
Why data is the array that response has an error?


ilusha2012:
Hello, dear developers!

In specification contain ([GQL-specification](http://facebook.github.io/graphql/draft/#sec-Response)) that if there are errors and data is empty then response does not have data.

The **webonyx/graphql-php** response data with errors where data is array:

![2018-05-07 12-34-34](https://user-images.githubusercontent.com/2123117/39686426-1c37dc50-51f3-11e8-93fb-5072da3f0edd.png)

And the apollo-client for Android handle this exception (The error is mean that data expects as an object):

`05-07 10:36:33.159 22628-22628/com.miakid.dev E/RequestCodeViewModel$su: com.apollographql.apollo.exception.ApolloParseException: Failed to parse http response
        at com.apollographql.apollo.internal.interceptor.ApolloParseInterceptor.parse(ApolloParseInterceptor.java:101)
        at com.apollographql.apollo.internal.interceptor.ApolloParseInterceptor$1.onResponse(ApolloParseInterceptor.java:53)
        at com.apollographql.apollo.internal.interceptor.ApolloServerInterceptor$1$1.onResponse(ApolloServerInterceptor.java:94)
        at okhttp3.RealCall$AsyncCall.execute(RealCall.java:153)
        at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
        at java.lang.Thread.run(Thread.java:818)
     Caused by: com.apollographql.apollo.json.JsonDataException: Expected BEGIN_OBJECT but was BEGIN_ARRAY at path $.data
        at com.apollographql.apollo.internal.json.BufferedSourceJsonReader.beginObject(BufferedSourceJsonReader.java:176)
        at com.apollographql.apollo.internal.json.ResponseJsonStreamReader.nextObject(ResponseJsonStreamReader.java:85)
        at com.apollographql.apollo.response.OperationResponseParser.parse(OperationResponseParser.java:91)
        at com.apollographql.apollo.internal.interceptor.ApolloParseInterceptor.parse(ApolloParseInterceptor.java:87)
        at com.apollographql.apollo.internal.interceptor.ApolloParseInterceptor$1.onResponse(ApolloParseInterceptor.java:53) 
[...]

PR #281, opened 2018-05-07
Removing data elements from response if the error throwing

ilusha2012:
More info: https://github.com/webonyx/graphql-php/issues/280


vladar:
Thanks!

PR #283, opened 2018-05-10
setRawVariableValues fix

ndrkltsk:
In my experience, $rawVariableValues ​​is null or a string representing a json. 
Tested with different implementations of graphiql clients


vladar:
Can you create an issue first with some details of your problem? The fix proposed doesn't seem right to me, but maybe we can do something else here.

Issue #284, opened 2018-06-10
Code Style


simPod:
Hello,

is there any codestyle to comply with when contributing?

Thanks!

vladar:
Ideally, we should follow PSR-4. But since there is no automatic PHP code style checks or PHP code style fixer in Travis, this is not enforced yet.

The best solution would be to set up code style checks or fixer for PSR-4, but I never found time for it. PRs are welcome!

simPod:
There are no PSR-4 coding standards AFAIK 🤔 

Some defaults are MySource, PEAR, PHPCS, PSR1, PSR2, Squiz and Zend but are pretty useless in their raw form TBH (eg. missing useful sniffs).

Personally I fancy this one https://github.com/doctrine/coding-standard It's the most advanced coding standard out there currently.

If interested, I can look into it.

vladar:
Yeah, sorry, I meant PSR-2 initially

vladar:
Doctrine code style looks reasonable for me. Go ahead with it if you have a moment!

simPod:
Kk, you can assign me to this issue. Expect PR next week.

fesor:
@vladar 

> there is no automatic PHP code style checks or PHP code style fixer in Travis

Why not use php-cs-fixer in dry run to be sure that all commits has proper code style?

simPod:
@fesor Travis CS step will be part of the PR

vladar:
Can you also add a note to [CONTRIBUTING.md](https://github.com/webonyx/graphql-php/blob/master/CONTRIBUTING.md) about code style requirements?

mfn:
@vladar are you open for using a non-open source tool?

https://styleci.io/ comes to my mind which has a very nice feature: you can configure it to automatically push a "corrected" source version on PRs (disclaimer: not 100% this works on PRs created from remote repos, but it can be set to work on master branch too).

This means: once a style has been configured/agreed upon, every PR or master will automatically receive commits from the service with the violations being resolved. No more manual checking or discussion what is correct.

Also: it's free for open source:
> We love Open Source and as such, public PHP projects will always be free with StyleCI.

I apologize in advance if this is inappropriate to suggest. I'm a commercial customer of the service for some time now. It's simply a blessing.

(PS: it's not about StyleCI specifically, I just know this service. The point: auto-commit source correction saves everyone time).

vladar:
[...]

PR #288, opened 2018-06-19
RFC: PHP CS

simPod:
Following https://github.com/webonyx/graphql-php/issues/284

- Added Code Sniffer with Doctrine CS
- Min PHP v is 7.1 as it's constraint for modern CS
- Fixed CS in `src/Errors`

For smoother transition typehint checks are now disabled and only modified lines are checked for CS compliance.

Doctrine and Slevomat CS require PHP ^7.1. Not sure, what's the roadmap of this project, can be merged after the PHP version constraint is set to not-outdated PHP versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
-.idea/
 composer.phar
 composer.lock
+phpcs.xml
 vendor/
-bin/
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ dist: trusty
 language: php
 
 php:
-    - 5.6
-    - 7.0
     - 7.1
     - 7.2
     - nightly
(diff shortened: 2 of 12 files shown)


vladar:
Next major version (0.13) will use PHP7.1+, so we can start converting the code to PHP 7.1

vladar:
Will this PR check pull requests only or will it fail builds after a regular push as well (if coding standard is not followed)?

simPod:
For PRs, it will only check lines changed so people don't have to fix the whole file for build to pass but only their changes must be compliant.

Otherwise, it starts full check when not PR.

vladar:
Can you also add composer scripts for linting and fixing?

vladar:
Yes. At least linting the whole project. We can probably add fixing the whole project later %)

vladar:
@simPod I've merged the PR, but it looks like it checks the whole project. Now build is failing on code style (with a ton of errors).

I doubt that we'll have enough time soon to fix all code style errors, so can you double check that it actually only checks PRs and maybe send a fix for it?

vladar:
Actually, ignore it. I commented out a check of the whole project. Only new PRs are checked for now.

PR #294, opened 2018-06-25
Use multi-line block for trailing quote

danez:
This fixes an issue that the following description `some "thing"` would be reprinted as `"""some "thing""""` which the parser cannot parse afterwards. Instead it will print now a multiline description
```
"""
some "thing"
"""
```

ref: https://github.com/graphql/graphql-js/commit/fdc10bb918e5d84ce6b34e122069851ca3e6f6a4#diff-ebaed8492e8d884ee4f2255e39909568


vladar:
Thank you!

PR #300, opened 2018-07-05
Adds support for the multipart/form-data content type

iainmckay:
Very simple PR, it adds support for multipart/form-data. It's not handled any differently to some of the other supported content types but the library just needs to recognize it.


vladar:
Thanks!

vladar:
FYI, published this in the v0.12.3

PR #305, opened 2018-07-09
Do not specify test groups when running tests in CI

simPod:


vladar:
Why not? We actually exclude ReactPromise from the default test group. But we still want to test it in Travis.

simPod:
Ah I see it is excluded in phpunit dist. Why is it so?

vladar:
Because it is an optional feature which relies upon ReactPHP promises which is not installed by default (it is suggested in the composer config, not required)

simPod:
What if I make ReactPHP promises dev dependency and therefore simple `vendor/bin/phpunit` can be used without any group exclusion anywhere?

PR #312, opened 2018-07-20
Added check for date objects on strings

chrispage1:
Hi,

I've added a check on the StringType to convert date instances (both Carbon and DateTime) to a JavaScript supported string.


vladar:
Each app has very different requirements to date formatting (that's why DateTime has no `__toString()` method). We prefer staying close to PHP logic here and only convert things with `__toString()` method.

PR #318, opened 2018-07-30
FIX Ensure that __toString return value is always casted as a string

robbieaverill:
`json_encode` can return false on failure, which causes PHP errors since `__toString` must return a string.

### DIFF
--- a/src/Language/AST/Node.php
+++ b/src/Language/AST/Node.php
@@ -84,7 +84,7 @@ abstract class Node
     public function __toString()
     {
         $tmp = $this->toArray(true);
-        return json_encode($tmp);
+        return (string) json_encode($tmp);
     }
 
     /**


vladar:
I am curious about when Node classes produce errors in this case? Do you have an example?

vladar:
Agree. Well if you face it again, please let us know when and why had it happened

robbieaverill:
Will do, thank you

zzdjk6:
Encounter the same problems, thanks for this fix.

I think this situation happens when the AST node has a lot of RECURSIVE information.

The data structure that causes `json_encode` failure is attached (in `print_r` format).

[json_encode_failed.txt](https://github.com/webonyx/graphql-php/files/2249503/json_encode_failed.txt)

PR #319, opened 2018-08-01
Fix for fatal php error: Call to a member function getLocation() on null, before normal error handling

jrots:
If you do a graphql call with a field that is required & doesn't exists in your request -  
you'll get a php fatal error that is triggered in the error handling of graphql. 
Because source of that node is not parsed/ empty..  
so you can never catch errors from the normal graphql error handler.


mfn:
Can you please also add a test/example?

PR #323, opened 2018-08-13
Fix CS in src/Language

simPod:


src/Utils/AST.php
```diff
@@ -230,9 +259,10 @@ static function astFromValue($value, InputType $type)
                 return new IntValueNode(['value' => $serialized]);
             }
             if (is_float($serialized)) {
-                if ((int) $serialized == $serialized) {
+                if ($serialized <= PHP_INT_MAX && floor($serialized) === $serialized) {
```
vladar: I do understand the motivation for this change, but an old form was chosen for performance. This method is called so many times during query execution that micro-optimizations which are usually unnecessary do make sense here. Can we optionally ignore code style checks for such intentional things?

simPod: This actually is not CS related directly. But as I required declared strict types after `<?php` this part stopped working as it's not type-safe. I'll look into it again and see if I come up with more performant way.

src/Language/AST/ArgumentNode.php
```diff
@@ -1,17 +1,23 @@
...
-     * @var NameNode
-     */
+    /** @var NameNode */
     public $name;
+
+
+    public function getKind() : string
+    {
+        return NodeKind::ARGUMENT;
+    }
```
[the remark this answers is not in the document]
simPod: We can leave them available publicly but create getters compliant with interfaces we define. `getKind()` is the first getter I defined.

I planned to continue with defining other properties via interfaces so we can better leverage type-safety.

image

image


So I suggest to rework it as
```php
    public function getKind() : string
    {
        return $this->kind;
    }
```

Performance shouldn't suffer as long as the property stays public https://steemit.com/php/@crell/php-use-associative-arrays-basically-never

src/Language/Printer.php
```diff
@@ -70,256 +80,330 @@ public static function doPrint($ast)
...
+                        // Anonymous queries with no directives or variable definitions can use
+                        // the query short form.
+                        return ! $name && ! $directives && ! $varDefs && $op === 'query'
+                            ? $selectionSet
+                            : $this->join([$op, $this->join([$name, $varDefs]), $directives, $selectionSet], ' ');
+                    },
+                    NodeKind::VARIABLE_DEFINITION       => function (VariableDefinitionNode $node) {
+                        return $node->variable . ': ' . $node->type . $this->wrap(' = ', $node->defaultValue);
+                    },
+                    NodeKind::SELECTION_SET             => function (SelectionSetNode $node) {
```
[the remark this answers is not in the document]
simPod: Usually there's a way how to format the code better rather than ignoring the CS rule :) Check the last commit, I tried to separate the blocks. To me, they make better sense like this.

src/Language/AST/ArgumentNode.php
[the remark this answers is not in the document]
simPod: BTW the reason I'm doing all this is to (in the end) add types to this lib and make it better prepared for future php releases so it can leverage all the new features that are going to come (with performance gains as well :)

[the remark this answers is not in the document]
vladar: Actually, your link says that getters do add a quite significant overhead to performance (8.21 vs 11.08 seconds in tests).

And that's actually the main reason why AST is designed like this (without getters). PHP is not the best language for projects like this, so we really have to sacrifice some of the best practices for performance gains.

So I would actually suggest reverting the getKind() getter. And it's better to separate the work on the code style from changes to interfaces and logic, because those require some benchmarking first.

src/Language/Printer.php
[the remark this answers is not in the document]
vladar: The new version is OK for me. Thanks!

P.S. I used to write the code with paddings like in this code style, but it was such a pain when you renamed something, that I abandoned it. Suprised that Doctrine uses it, but I guess I can live with it with a help of IDE and auto fix.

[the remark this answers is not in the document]
simPod: PS: same for me. But it was the time when it was hard to hide whitespace in diffs (no such feature in github) etc. 

But it seems to make code blocks to look well-aranged to me so I adopted it again :) Tools make it easy to use.

src/Language/AST/ArgumentNode.php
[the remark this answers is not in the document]
simPod: In the link it is said that 11.08s avg is for private properties but I suggested to keep having public properties and adding getters for gaining API advantage via interfaces.

Also, I understand that making such changes is better for separate PRs. I'll try to clean it up from this one and open PR when we can reopen this discussing, alright?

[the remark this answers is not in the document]
vladar: Sure

src/Language/AST/FieldDefinitionNode.php
```diff
@@ -1,35 +1,31 @@
...
-    /**
-     * @var TypeNode
-     */
+    /** @var TypeNode */
     public $type;
 
-    /**
-     * @var DirectiveNode[]|NodeList
-     */
+    /** @var DirectiveNode[]|NodeList */
```
spawnia: Shouldn't those be changed around (as well as in other places where NodeList occurs), because the property actually IS a NodeList

src/Utils/AST.php
spawnia: Two notes on this:

1. Such micro-optimization deserve a line comment.
2. Yes, there are comments to disable style-checks on a per-line basis. Depends on the tool that is used.

src/Language/AST/FieldDefinitionNode.php
simPod: Could you be more specific please? Not sure what you mean by that. Thanks

spawnia: I believe that all lists of related nodes in the AST are actually of the type `NodeList`, so the type annotation should match that. It is fine to have the underlying type of the NodeList as a secondary hint, so it should be `NodeList|DirectiveNode[]`

simPod: Ah, I didn't touch this TBH. Only reformatted the annotation to be CS compliant.

The order of union type hints doesn't matter and sometimes can be relative. We have a CS check for having `null` as the last type but as long as there's no custom CS check for having NodeList as the first type, I don't really see a point in it.

I'd suggest you open a issue / PR with union types proper ordering if you want to pursue this (inspiration https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/TypeHints/NullTypeHintOnLastPositionSniff.php).

spawnia: I am not actually sure that having both is actually correct. I just looked up the meaning of the pipe character `|` in types. In this case, it actually always is one exact type, `NodeList`.

simPod: AFAIK pipe means `either a or b` but haven't checked whether both are correct or just one.

spawnia: Agreed. Go ahead then

PR #329, opened 2018-08-23
Code quality

mcg-web:
Adding some tools to help maintainers.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,9 +18,6 @@
   },
   "config": {
     "bin-dir": "bin",
-    "platform": {
-      "php": "5.6.0"
-    },
     "preferred-install": "dist",
     "sort-packages": true
 },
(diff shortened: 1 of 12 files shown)


vladar:
Seet, thanks! But can you please change the target branch from 0.12.x to master? I don't think we'll do much work in 0.12.x branch anymore (just minor fixes).

composer.json
```diff
@@ -18,9 +18,6 @@
   },
   "config": {
     "bin-dir": "bin",
-    "platform": {
```
mcg-web: Remove this to test lib in environment without faking platform.

mcg-web:
@vladar 0.12 is break for php 7.3 nightly and platform config was hiding this

vladar:
Yeah, but those fixes will stay in 0.12.x, or do you plan to merge them later to master?

mcg-web:
0.12.x should always be merged in master, this is the best way to maintain both branch while 0.12 is still maintained, yes I will merged it in master.

composer.json
```diff
@@ -36,6 +34,15 @@
       "GraphQL\\Examples\\Blog\\": "examples/01-blog/Blog/"
     }
   },
+  "scripts": {
+    "static-analysis": [
+      "rm phpstan.phar || true",
```
simPod: why not simply use phpstan as a dev dependency? this looks unnecessarily complicated

mcg-web: cause phpstan require php 7.1 ;)

simPod: This lib does [too](https://github.com/webonyx/graphql-php/blob/master/composer.json#L12), hm 🤔?

mcg-web: I m not on master ;) but 0.12.x

simPod: Not sure if I understand correctly :D This is how it should look like 

https://github.com/doctrine/dbal/blob/master/composer.json#L23

And in travis, simply use `script: vendor/bin/phpstan analyse`

mcg-web:
@vladar can you please review this and tell me if I need to split this in two PR. I need to fix nightly build for the bundle ;)

mcg-web:
Thank you I'm working on a PR to merge these changes in master?

PR #331, opened 2018-08-26
Fix CS in src/Type

simPod:


src/Type/Definition/BooleanType.php
```diff
@@ -28,7 +27,7 @@ class BooleanType extends ScalarType
      */
     public function serialize($value)
     {
-        return !!$value;
+        return ! ! $value;
```
vladar: I think we should use `(bool) $value` with a new code style. `! ! $value` looks really weird

src/Type/Definition/CustomScalarType.php
```diff
@@ -54,16 +61,18 @@ public function assertValid()
...
             'functions are also provided.'
         );
-        if (isset($this->config['parseValue']) || isset($this->config['parseLiteral'])) {
-            Utils::invariant(
-                isset($this->config['parseValue']) && isset($this->config['parseLiteral']) &&
-                is_callable($this->config['parseValue']) && is_callable($this->config['parseLiteral']),
-                "{$this->name} must provide both \"parseValue\" and \"parseLiteral\" functions."
-            );
+        if (! isset($this->config['parseValue']) && ! isset($this->config['parseLiteral'])) {
+            return;
```
vladar: I am curious if this style of logic is recommended by the coding standard (when you return or continue and leave the main logic at the bottom of the function / loop)?

simPod: Yes, it's called early exit. It emphatises the actual logic so it is not unecessarily nested and all distracting conditions is executed before it.

The code is then easier to read as the possitive path is on the first level in the function. Typical example https://github.com/doctrine/dbal/pull/2937/commits/6013259628fae7ac1e5cddebce8f83528b9031a6

src/Type/Definition/BooleanType.php
simPod: Agreed, double `!!` looks somewhat magical. TBH I had no idea what it does before.

src/Type/Definition/CustomScalarType.php
simPod: Don't see any advantage of not using early return, only that one that the code somehow mirrors that reference implementation, that would make sense.

For debugging, well, ¯\_(ツ)_/¯  
Since I use xdebug, I never dump values. Of course it's about personal preference but code should not be influenced by that preference.

Also, adding some more reference  
https://www.slideshare.net/guilhermeblanco/php-for-adults-clean-code-and-object-calisthenics 39+  
https://www.slideshare.net/guilhermeblanco/object-calisthenics-applied-to-php 9  
https://github.com/jupeter/clean-code-php#avoid-nesting-too-deeply-and-return-early-part-1  
https://github.com/jupeter/clean-code-php#avoid-nesting-too-deeply-and-return-early-part-2

It's your call, we can exclude that sniff ofc.

Where can I find the reference code BTW? Thanks!

vladar: So it is a sniff from the coding standard, right?

[Reference GraphQL implementation](https://github.com/graphql/graphql-js)

simPod: Yes

simPod: Kk, so I guess that's all.

Thanks for the link btw

PR #332, opened 2018-08-26
Fix CS in tests/Error

simPod:


tests/Error/ErrorTest.php
```diff
@@ -9,131 +12,104 @@
 
 class ErrorTest extends TestCase
 {
-    /**
-     * @it uses the stack of an original error
-     */
-    public function testUsesTheStackOfAnOriginalError()
```
vladar: Please do not delete those comments in tests. They are very helpful when migrating tests from the reference implementation.

simPod: Ayright, what is `@it` annotation's meaning?

vladar: It contains the name of the mocha test case from the reference implementation. I.e. `it('uses the stack of an original error')`.

This way it is easy to find the same test case in the reference implementation to check if something had changed there.

simPod: Can we use `* @see it('uses the stack of an original error')` instead?

vladar: I don't care much as long as the original test description is preserved

simPod: Cool, I'll use standard @see and we can drop confusing `@it` annotation ;)

PR #333, opened 2018-08-28
Static analysis

mcg-web:


composer.json
```diff
@@ -16,9 +16,12 @@
   "require-dev": {
     "doctrine/coding-standard": "^4.0",
     "phpunit/phpunit": "^7.2",
-    "psr/http-message": "^1.0"
+    "phpstan/phpstan-shim": "^0.10",
+    "psr/http-message": "^1.0",
+    "react/promise": "2.*"
   },
   "config": {
+    "bin-dir": "bin",
```
vladar: We removed this

.travis.yml
```diff
@@ -6,10 +6,6 @@ php:
     - 7.2
     - nightly
 
-matrix:
-    allow_failures:
-        - php: nightly
-
```
vladar: Why removing this?

mcg-web: Because it is a duplicated in job entry

vladar: oh, I see now

vladar:
@simPod mind to review? As far as I remember, you've authorized Travis-related stuff.

.travis.yml
```diff
@@ -41,7 +34,7 @@ jobs:
         - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{.disabled,}
         - if [[ ! $(php -m | grep -si xdebug) ]]; then echo "xdebug required for coverage"; exit 1; fi
       script:
-        - ./vendor/bin/phpunit --coverage-clover clover.xml
+        - bin/phpunit --coverage-clover clover.xml
```
simPod: as mentioned in https://github.com/webonyx/graphql-php/pull/333/files#r213260516, this should be reverted

.travis.yml
```diff
@@ -18,12 +14,9 @@ before_install:
...
     - travis_retry composer self-update
 
-install:
-    - composer require react/promise:2.*
-    - composer require psr/http-message:1.*
-    - travis_retry composer update --prefer-dist
+install: composer install --dev --prefer-dist
 
-script: ./vendor/bin/phpunit --group default,ReactPromise
+script: bin/phpunit --group default,ReactPromise
```
simPod: as mentioned in https://github.com/webonyx/graphql-php/pull/333/files#r213260516, this should be reverted

composer.json
```diff
@@ -39,6 +42,7 @@
     "psr/http-message": "To use standard GraphQL server"
   },
   "scripts": {
-    "lint" : "phpcs"
+    "lint" : "phpcs",
+    "static-analysis": "@php bin/phpstan.phar analyse --ansi -l 1 -c phpstan.neon src"
```
simPod: why use phar instead of standard binary?

`composer require --dev phpstan/phpstan` and then `vendor/bin/phpstan`

composer.json
```diff
@@ -16,9 +16,12 @@
   "require-dev": {
     "doctrine/coding-standard": "^4.0",
     "phpunit/phpunit": "^7.2",
-    "psr/http-message": "^1.0"
+    "phpstan/phpstan-shim": "^0.10",
+    "psr/http-message": "^1.0",
+    "react/promise": "2.*"
```
simPod: I'd also add `composer require --dev phpstan/phpstan-phpunit`

.travis.yml
```diff
@@ -62,4 +55,9 @@ jobs:
 #      php: 7.1
 #      install: travis_retry composer install --prefer-dist
 #      script:
-#        - ./vendor/bin/phpcs
+#        - ./vendor/./vendor/bin/phpcs
```
simPod: 1) I'd add blank line below, 
2) typo introduced

composer.json
simPod: stil valid

composer.json
```diff
@@ -39,6 +41,7 @@
     "psr/http-message": "To use standard GraphQL server"
   },
   "scripts": {
-    "lint" : "phpcs"
+    "lint" : "phpcs",
+    "static-analysis": "@php ./vendor/bin/phpstan.phar analyse --ansi -l 1 -c phpstan.neon src"
```
simPod: why use phar instead of standard binary?

`composer require --dev phpstan/phpstan` and then `vendor/bin/phpstan`

composer.json
mcg-web: We using this only for src in this case so no need an extension for phpunit

composer.json
mcg-web: cause phar has less dependencies and this lib don't likes to have useless requirements

mcg-web:
why tests should be tested? this is only for the package not the source code

simPod:
Because tests can have bugs too. Tests describe what the source code should do. So tests (because they work as a specification) should be checked as well to easily discover poorly/wrongly written tests.

composer.json
simPod: They're dev requirements so it doesn't really matter. Also, whether they're hidden in phar file or in vendor folder doesn't make much difference.

However, adding it via composer is more flexible (eg. easier to add phpstan phpunit or strict rules)

composer.json
simPod: Followup https://github.com/webonyx/graphql-php/pull/333#pullrequestreview-150080275 First let's resolve that

composer.json
```diff
@@ -16,7 +16,9 @@
   "require-dev": {
     "doctrine/coding-standard": "^4.0",
     "phpunit/phpunit": "^7.2",
-    "psr/http-message": "^1.0"
+    "phpstan/phpstan-shim": "^0.10",
+    "psr/http-message": "^1.0",
+    "react/promise": "2.*"
```
simPod: I like how you resolved those missing dependencies. Them being missing worried me for a long time. No more 👍

composer.json
mcg-web: I'll change it :+1:

src/Validator/Rules/KnownDirectives.php
```diff
@@ -59,7 +61,9 @@ public static function unknownDirectiveMessage($directiveName)
     }
 
     /**
-     * @param (Node|NodeList)[] $ancestors
+     * @param Node[]|NodeList[] $ancestors
```
simPod: Previous version was correct

image

mcg-web:
I'll work on feedback today :+1:

phpstan.neon
```diff
@@ -0,0 +1,3 @@
+includes:
```
simPod: this file should have .dist extension (phpstan.neon.dist)

same as for phpunit or phpcs configuration

`phpstan.neon` should be added to `.gitignore`

mcg-web:
@simPod what is the command to automatique fix phpcs please?

simPod:
@mcg-web it is `phpcbf` but not recommended for now as all files are not migrated to new CS yet. So if you run it now, it will fix codestyle in all files and produce huge diff.

If you want specific files, you can run `phpcbf <relative filepath>`

simPod:
Migration to CS is continuously in progress ;) https://github.com/webonyx/graphql-php/issues/284

mcg-web:
I can't fix CS on this PR right now cause it will remove all `@it` annotations. I'll rebase this PR on master after merging #334 .

simPod:
@mcg-web merged ;)

mcg-web:
I rebase it but when I fix cs it is changing a lot of thing, is this normal?

simPod:
https://github.com/webonyx/graphql-php/pull/333#issuecomment-417638287 that's because it has not been migrated all yet

Therefore, it's better to address those issues manually for now.

I'm getting to the point when the whole code base is compliant with the CS.

mcg-web:
@simPod @vladar last feedback please and we ready to merge! thank you.

simPod:
This is still incorrect https://github.com/webonyx/graphql-php/pull/333/files/87ecc3eb23e318ef8d871a2060c052070b85a0f2#diff-b5af65a9ceb44d8fdf02823c66ccec3b

.travis.yml
```diff
@@ -46,8 +39,15 @@ jobs:
         - wget https://scrutinizer-ci.com/ocular.phar
         - php ocular.phar code-coverage:upload --format=php-clover clover.xml
 
-    - stage: Coding standard
+    - stage: Code Quality
       php: 7.1
+      env: STATIC_STANDARD
```
simPod: `CODING_STANDARD`

mcg-web:
`We went looking everywhere, but couldn’t find those commits.`?

simPod:
Ah wrong link

https://github.com/webonyx/graphql-php/pull/333/files#diff-b5af65a9ceb44d8fdf02823c66ccec3bR64
https://github.com/webonyx/graphql-php/pull/333#discussion_r213306378

src/Validator/Rules/KnownDirectives.php
mcg-web: But this version is correct too, I'll change if this can make the code great again :laughing:

mcg-web: done!

.travis.yml
```diff
@@ -46,8 +39,15 @@ jobs:
...
-    - stage: Coding standard
+    - stage: Code Quality
       php: 7.1
+      env: CODING_STANDARD
       install: travis_retry composer install --prefer-dist
       script:
         - ./vendor/bin/phpcs
+
+    - stage: Code Quality
+      php: 7.2
```
simPod: We should have 7.1 here (the minimum required version). If we leave 7.2, it will force features not available in 7.1.

mcg-web: Can you please do a single review I'm done of changing code every second, thank you ^^

src/Validator/Rules/KnownDirectives.php
mcg-web: It's seam that this syntax introduce error phpcs https://travis-ci.org/webonyx/graphql-php/jobs/423635233 can you please tell me what is going wrong here?

.travis.yml
simPod: Tryin' but I always find something new when I go through it again 😅

src/Validator/Rules/KnownDirectives.php
mcg-web: can you please provide a example I don't really get how to note an annotation sorry

mcg-web:
It look good to me now, It seems that Scrutinizer fails on an old part of the code, can you please check this out @simPod :+1:

mcg-web:
Party hard

PR #341, opened 2018-09-01
Remove the word "simply"

adactio:
Using the word "simply" or "just" is probably intended to convey the ease of use, but it can have the unintended effect of making the reader feel like the instructions are being blasé.

https://adactio.com/journal/13521


vladar:
Thanks!

PR #349, opened 2018-09-03
Improve PHPStan Configuration

simPod:
Moving some parameters inside config file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .phpcs-cache
-composer.phar
 composer.lock
 phpcs.xml
 phpstan.neon
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,3 +1,10 @@
+parameters:
+    level: 1
+
+    paths:
+        - %currentWorkingDirectory%/src
+        - %currentWorkingDirectory%/tests
+
 includes:
 	- vendor/phpstan/phpstan-phpunit/extension.neon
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,6 +1,6 @@
 .phpcs-cache
-composer.phar
```
mcg-web: Why removing this rule?

phpstan.neon.dist
```diff
@@ -1,3 +1,12 @@
+parameters:
+    level: 1
+
+    paths:
+        - %currentWorkingDirectory%/src
+        - %currentWorkingDirectory%/tests
+
+    tmpDir: %currentWorkingDirectory%/var/phpstan
```
mcg-web: why change phpstan tmp dir? the default sys tmp seems to be good choose, no?

.gitignore
simPod: It is not part of this project and is rather user specific. 

It should be eventually [ignored globally](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)

phpstan.neon.dist
simPod: I see, I'll revert it

.gitignore
mcg-web: maybe we should add this to the documentation? cause composer documentation push to use composer.phar direct in the project itself. but I agree this is no a be deal.

PR #354, opened 2018-09-19
Use self:: in tests where appropriate

simPod:
Those methods are declared static so should be called that way. They do not contain any instance-specific behaviour, they're rather wrappers for throwing exceptions.

### DIFF
--- a/tests/Error/ErrorTest.php
+++ b/tests/Error/ErrorTest.php
@@ -20,7 +20,7 @@ class ErrorTest extends TestCase
         $prev = new \Exception('Original');
         $err  = new Error('msg', null, null, null, null, $prev);
 
-        $this->assertSame($err->getPrevious(), $prev);
+        self::assertSame($err->getPrevious(), $prev);
     }
 
     /**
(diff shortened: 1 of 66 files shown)


mfn:
What's the benefit?

Even the official docs in their introduction to suggest the existing code: https://phpunit.readthedocs.io/en/7.3/writing-tests-for-phpunit.html

Also all test code I've ever seen uses `$this->` for the assertions, so I'm wondering.

Thanks

simPod:
[The official docs say there's no wrong or good way](https://phpunit.readthedocs.io/en/7.3/assertions.html#static-vs-non-static-usage-of-assertion-methods).

But there's actually a good way -> the standard and common one everyone use when they write their code. When you declare the method static, you call it using `::` and not using `->`, right? This messy habit of calling static methods using `->` while working with PHPUnit has origin unknown to me 🤔.

The main reason for this change, as I stated above, is that those static methods doesn't contain any instance-specific logic, they're static wrappers. Compare `self::assertTrue()` and `$this->expectException()` for instance. You simply cannot call `self::expectException()` because it works with test instance and have to be called using `->`.

Next steps: When we want to enhance static analysis, [PHPStan's Strict Rules](https://github.com/phpstan/phpstan-strict-rules) analyse that `Statically declared methods are called statically`. You wouldn't want to exclude this check for your code. There's no reason to exclude it for tests either.

mfn:
> Also seems to me the whole test suite is now running faster according to travis' times.

🤔

> Next steps: When we want to enhance static analysis, [PHPStan's Strict Rules](https://github.com/phpstan/phpstan-strict-rules) analyse that `Statically declared methods are called statically`. You wouldn't want to exclude this check for your code. There's no reason to exclude it for tests either.

Never occurred to me to statically analyze the tests ;)

But hey, awesome, thanks for the details => much appreciated!

simPod:
;)

> Never occurred to me to statically analyze the tests ;)

It kinda forces you to write them more clean and also sometimes discover bugs in them like that they test something slighly different than you intended to.

PR #360, opened 2018-10-01
Add missing extensionASTNodes

Torsten85:
In my effort to port [extendSchema](https://github.com/graphql/graphql-js/blob/master/src/utilities/extendSchema.js) from graphql-js to php, I noticed that some extensionASTNodes are currently not defined. This PR will add missing extensionASTNodes and also allows for extending the schema.

My big PR including "extendSchema" (with all the tests) will follow when this PR is merged.

### DIFF
--- a/src/Language/AST/NodeKind.php
+++ b/src/Language/AST/NodeKind.php
@@ -69,6 +69,9 @@ class NodeKind
 
     const DIRECTIVE_DEFINITION = 'DirectiveDefinition';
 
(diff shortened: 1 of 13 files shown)


Torsten85:
I don't really know how to improve the test coverage of the static create method of SchemaConfig. I don't see any test for this method... could use some help here.

vladar:
Great! Looking forward to the extendSchema feature!

PR #363, opened 2018-10-05
Bump cs

simPod:
Bumping CS to v5, some nice sniffs there, see https://github.com/doctrine/coding-standard/releases/tag/5.0.0


vladar:
Is it ready for merge?

simPod:
Yeap!

src/Executor/Executor.php
```diff
@@ -1481,6 +1523,6 @@ public static function defaultFieldResolver($source, $args, $context, ResolveInf
             }
         }
 
-        return $property instanceof \Closure ? $property($source, $args, $context, $info) : $property;
+        return is_callable($property) ? $property($source, $args, $context, $info) : $property;
```
vladar: Actually, I've missed this part during a review. I'll restore it to the original version since `instanceof` is 2-10 times faster than `is_callable()` (and that's a reason why it was there in the first place)

simPod: Sounds reasonable 👍

simPod: BTW those things would deserve a comment to explain why it is there if there's a reason like this.

It helps future refactoring as you know whether and how you are allowed to change it.

PR #364, opened 2018-10-05
Replace node kind checks by InstanceOf checks

simPod:
This allows better code analysis as every tool/IDE knows what object type goes into a condition branch

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -138,8 +138,8 @@ class ReferenceExecutor implements ExecutorImplementation
         $operation                    = null;
         $hasMultipleAssumedOperations = false;
         foreach ($documentNode->definitions as $definition) {
-            switch ($definition->kind) {
-                case NodeKind::OPERATION_DEFINITION:
+            switch (true) {
+                case $definition instanceof OperationDefinitionNode:
                     if (! $operationName && $operation) {
                         $hasMultipleAssumedOperations = true;
                     }
@@ -148,7 +148,7 @@ class ReferenceExecutor implements ExecutorImplementation
                         $operation = $definition;
                     }
                     break;
-                case NodeKind::FRAGMENT_DEFINITION:
+                case $definition instanceof FragmentDefinitionNode:
                     $fragments[$definition->name->value] = $definition;
                     break;
             }
@@ -334,8 +334,8 @@ class ReferenceExecutor implements ExecutorImplementation
     ) {
         $exeContext = $this->exeContext;
         foreach ($selectionSet->selections as $selection) {
-            switch ($selection->kind) {
-                case NodeKind::FIELD:
+            switch (true) {
+                case $selection instanceof FieldNode:
                     if (! $this->shouldIncludeNode($selection)) {
                         break;
                     }
@@ -345,7 +345,7 @@ class ReferenceExecutor implements ExecutorImplementation
                     }
                     $fields[$name][] = $selection;
                     break;
-                case NodeKind::INLINE_FRAGMENT:
+                case $selection instanceof InlineFragmentNode:
                     if (! $this->shouldIncludeNode($selection) ||
                         ! $this->doesFragmentConditionMatch($selection, $runtimeType)
                     ) {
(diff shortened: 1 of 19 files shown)


mfn:
The `switch (true)` _truly_ reads weird.

On a related note: shouldn't all those `switch` statements have a "default: throw…" case? I noticed some have, some don't (I know, existing code).

simPod:
Yup, it is indeed but PHP has no better way yet (though I believe it might be changed in the future).

I suppose I could add a default throw, that would make sense as all logical branches would be covered then 👍

simPod:
Seems like `default: throw...` statemens are not there for a reason (eg. they're handled later).

mfn:
Ops 🤷‍♀️ Who would knew…

vladar:
One reason why `instanceof` checks were not used there in the first place was performance. I was 100% sure that strict equality checks were faster than `instanceof`.

So changes like this do require some benchmarking (and we don't have a good enough benchmark for such things in our phpbench yet). I think we need some bench for a query with ~2000 field calls (and ideally make sure all changed paths are affected).

Those micro-optimizations seem minor, but when we sum them up and then multiply x 1000 (or more) - it becomes noticeable.

simPod:
Started with a micro benchmark

```php
<?php

class A
{
    public $kind = 'A';
}

$s = microtime(true);
for ($i = 0; $i < 10000000; $i++) {
    $a = new A();

    if ($a instanceof A) {
    }
}
var_dump(microtime(true) - $s);

$s = microtime(true);
for ($i = 0; $i < 10000000; $i++) {
    $a = new A();

    if ($a->kind === 'A') {
    }
}
var_dump(microtime(true) - $s);
```

On PHP 7.2
```
double(7.4557127952576)

double(7.6205501556396)
```

mfn:
wow, why do our numbers differ so much?

```
php -v
PHP 7.2.10-0ubuntu0.18.04.1 (cli) (built: Sep 13 2018 13:45:02) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
$ php bench.php
float(0.53607106208801)
float(0.6186990737915)
$ php bench.php
float(0.49281096458435)
float(0.59114003181458)
$ php bench.php
float(0.4991409778595)
float(0.6029200553894)
```
But yeah, `instanceof` seems to be fater.

Did you maybe forget a 0? In that case I get:
```
$ php bench.php
float(5.1001918315887)
float(6.1185019016266)
```

simPod:
Left xdebug enabled ;) With `php -n` having similar numbers

simPod:
@vladar just checked HugeSchemaBench and it seems that it does query operations on huge schema 🤔

vladar:
Those benches just test instanceof hits, not misses. There are actually several use cases we need to bench: hits, misses, hits and misses on the parent type (like a extends b, $a instanceof b). I think the latest one is the one we actually use the most.

simPod:
What about this

```php
<?php

class A
{
    public $kind = 'A';
}

class B extends A
{
}

$s = microtime(true);
for ($i = 0; $i < 10000000; $i++) {
    $b = new B();

    if ($b instanceof A) {
    }
}
var_dump(microtime(true) - $s);

$s = microtime(true);
for ($i = 0; $i < 10000000; $i++) {
    $b = new B();

    if ($b->kind === 'A') {
    }
}
var_dump(microtime(true) - $s);
```

```
float(0.54834389686584)
float(0.62606191635132)
```

--------------

```php
<?php

class A
{
    public $kind = 'A';
}

class B extends A
{
}

$s = microtime(true);
for ($i = 0; $i < 10000000; $i++) {
    $b = new B();

    if (!$b instanceof A) {
    }
}
var_dump(microtime(true) - $s);

$s = microtime(true);
for ($i = 0; $i < 10000000; $i++) {
    $b = new B();

    if ($b->kind !== 'A') {
    }
}
var_dump(microtime(true) - $s);
```

```
float(0.62735915184021)
float(0.65307688713074)
```

vladar:
I am on christmas/ny vacation at the moment, will get back to it later.

PR #366, opened 2018-10-09
Fix annotation

enumag:
[Parser](https://github.com/webonyx/graphql-php/blob/v0.12.6/src/Language/Parser.php#L739) creates BooleanValueNode with boolean, not string.

BooleanType and Printer also both expect it to be boolean.


vladar:
Thanks!

PR #367, opened 2018-10-09
Use PHPStan strict rules

simPod:
Two rules excluded: missing type hints and empty() usage

### DIFF
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -258,14 +258,14 @@ class Executor
                 $rawVariableValues ?: []
             );
 
-            if ($coercedVariableValues['errors']) {
-                $errors = array_merge($errors, $coercedVariableValues['errors']);
-            } else {
+            if (empty($coercedVariableValues['errors'])) {
                 $variableValues = $coercedVariableValues['coerced'];
+            } else {
+                $errors = array_merge($errors, $coercedVariableValues['errors']);
             }
         }
 
-        if ($errors) {
+        if (! empty($errors)) {
             return $errors;
         }
 
@@ -322,6 +322,7 @@ class Executor
         if ($data !== null) {
             $data = (array) $data;
         }
+
         return new ExecutionResult($data, $this->exeContext->errors);
     }
 
@@ -354,6 +355,7 @@ class Executor
                     null,
                     function ($error) {
                         $this->exeContext->addError($error);
+
                         return $this->exeContext->promises->createFulfilled(null);
                     }
                 );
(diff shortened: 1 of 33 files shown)


src/Validator/ValidationContext.php
```diff
@@ -154,15 +154,15 @@ public function getRecursivelyReferencedFragments(OperationDefinitionNode $opera
...
             $fragments      = [];
             $collectedNames = [];
             $nodesToVisit   = [$operation];
             while (! empty($nodesToVisit)) {
                 $node    = array_pop($nodesToVisit);
                 $spreads = $this->getFragmentSpreads($node);
-                for ($i = 0; $i < count($spreads); $i++) {
-                    $fragName = $spreads[$i]->name->value;
+                foreach ($spreads as $iValue) {
+                    $fragName = $iValue->name->value;
```
vladar: Maybe rename `$iValue` to `$spread`?

src/Executor/Executor.php
```diff
@@ -258,14 +258,14 @@ private static function buildExecutionContext(
...
-            } else {
+            if (empty($coercedVariableValues['errors'])) {
                 $variableValues = $coercedVariableValues['coerced'];
+            } else {
+                $errors = array_merge($errors, $coercedVariableValues['errors']);
             }
         }
 
-        if ($errors) {
+        if (! empty($errors)) {
```
mfn: Is there a real/technical benefit to this?

- I always prefer to read the "positive" form: "if something is…"
  than
- the negated form "if something is not …"

vladar: I don't bother much, but just curious - is it phpstan requirement and we'll have to stick with it or just your preference?

simPod: PHPStan preference is not to use `empty()` at all but use `count()` (cuz `empty()` is less strict but IMO for arrays it's just fine). That's why I added it as excluded rule (otherwise it would force to change all empty()ies to count()s).

But here the check is that you have to use boolean comparison in IF statement, that is eg. `if ($value !== null)` instead of `if ($value)`

simPod: @vladar what about this? Have I addressed everything?

vladar: @simPod I am travelling at the moment, so not always available

simPod: @vladar  sure, thanks for letting me know. Enjoy ur travels ;)

vladar:
Weird, but the build is broken. Can you double check your changes?

simPod:
This is the reason https://github.com/phpstan/phpstan/releases/tag/0.10.5 :)

PR #369, opened 2018-10-09
Updated default field resolver code in documentation.

nagledb:
Updated code to match definition in GraphQL\Executor\Executor::defaultFieldResolver. Change is notable because previous version did not indicate that the callable receives the $info variable.

### DIFF
--- a/docs/data-fetching.md
+++ b/docs/data-fetching.md
@@ -118,7 +118,7 @@ function defaultFieldResolver($source, $args, $context, \GraphQL\Type\Definition
         }
     }
 
-    return $property instanceof \Closure ? $property($source, $args, $context) : $property;
+    return $property instanceof Closure ? $property($source, $args, $context, $info) : $property;
 }
 ```
 


docs/data-fetching.md
```diff
@@ -118,7 +118,7 @@ function defaultFieldResolver($source, $args, $context, \GraphQL\Type\Definition
         }
     }
 
-    return $property instanceof \Closure ? $property($source, $args, $context) : $property;
+    return is_callable($property) ? $property($source, $args, $context, $info) : $property;
```
vladar: We actually reverted `is_callable()` here back to `instanceof Closure`. Please revert this part and I'll merge. Thanks for noticing!

nagledb:
Sure, I reverted `is_callable()` back to `instanceof Closure` and updated the PR.

zorji:
> Sure, I reverted `is_callable()` back to `instanceof Closure` and updated the PR.

@nagledb I think it should be `\Closure` instead of `Closure`.

nagledb:
The current code uses `Closure`. The leading slash isn't needed because the file uses `use Closure;`.

zorji:
> The current code uses `Closure`. The leading slash isn't needed because the file uses `use Closure;`.

Oh, I thought the default resolver was provided as an template for users to start their custom resolver. Sorry about that.

PR #377, opened 2018-10-23
Error handling improvements

yura3d:
**graphql-php** shows incorrect error when we pass a field argument value of incorrect type.

How to reproduce:

1. Create field in **QueryType** that takes a non-null argument of **StringType**. This step is described in [Hello world example](https://github.com/webonyx/graphql-php/blob/0.12.x/docs/getting-started.md#hello-world)
2. Query this field, but put a **ListOfType** value to the argument (simply []) instead of **StringType** value.

For this query **graphql-php** returns an exception of category "internal", so we get "Internal server error" string at the output (in "message" field). Based on this information, it's unobvious to get field caused the error (we must count query lines and columns) and we haven't the reason of the error:
image

But let's query the field without the argument. Because it's non-null, we get an exception of category "graphql" and very detailed "message" field, that contains field name, argument name and required type of value:
image

I think the situation with incorrect value type should cause similar detailed error, so in this pull request I did the next output:
image

### DIFF
--- a/src/Validator/Rules/ValuesOfCorrectType.php
+++ b/src/Validator/Rules/ValuesOfCorrectType.php
@@ -16,9 +16,9 @@ use GraphQL\Language\AST\StringValueNode;
 use GraphQL\Language\AST\ValueNode;
 use GraphQL\Language\Printer;
 use GraphQL\Language\Visitor;
-use GraphQL\Type\Definition\FieldArgument;
 use GraphQL\Type\Definition\EnumType;
 use GraphQL\Type\Definition\EnumValueDefinition;
+use GraphQL\Type\Definition\FieldArgument;
 use GraphQL\Type\Definition\InputObjectType;
 use GraphQL\Type\Definition\ListOfType;
 use GraphQL\Type\Definition\NonNull;
@@ -39,10 +39,10 @@ class ValuesOfCorrectType extends AbstractValidationRule
 
     static function badValueMessage($typeName, $valueName, $message = null, $context = null)
     {
-        if ($context && ($arg = $context->getArgument()) && $arg instanceof FieldArgument) {
-            $fieldName = self::$fieldName;
+        $fieldName = self::$fieldName;
+        if ($context AND $arg = $context->getArgument() AND $arg instanceof FieldArgument) {
             $argName = $arg->name;
-            return "Field \"{$fieldName}\" argument \"{$argName}\" requires type {$typeName}, found {$valueName}" . ($message ? "; ${message}" : '.');
+            return "Field \"{$fieldName}\" argument \"{$argName}\" requires type {$typeName}, found {$valueName}" . ($message ? "; {$message}" : '.');
         }
         return "Expected type {$typeName}, found {$valueName}"  .
             ($message ? "; ${message}" : '.');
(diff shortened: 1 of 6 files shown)


yura3d:
Refactored tests are coming soon

vladar:
Can I see the original error message and a stack trace of this internal error? ([see here](https://webonyx.github.io/graphql-php/error-handling/#debugging-tools) how to get it) My gut feeling is that we can just fix the underlying bug causing this to get proper error messages.

yura3d:
> Can I see the original error message and a stack trace of this internal error? ([see here](https://webonyx.github.io/graphql-php/error-handling/#debugging-tools) how to get it) My gut feeling is that we can just fix the underlying bug causing this to get proper error messages.

Of course, here it is:
image

And here I'll simply repeat the output after my changes (for comparing):
image

vladar:
But how about the stack trace?

yura3d:
I'm sorry, it took some time to remove all confidential data from the stack trace.
https://gist.github.com/yura3d/48ebc6e91ef1a65b595043ceb0d7ec82

src/Validator/Rules/ValuesOfCorrectType.php
```diff
@@ -33,8 +35,15 @@
  */
 class ValuesOfCorrectType extends AbstractValidationRule
 {
-    static function badValueMessage($typeName, $valueName, $message = null)
+    private static $fieldName;
```
vladar: We shouldn't rely on static $fieldName here. In theory, you may have multiple rules running in parallel which would break with static variable here. See below for suggestions.

src/Validator/Rules/ValuesOfCorrectType.php
```diff
@@ -33,8 +35,15 @@
...
+
+    static function badValueMessage($typeName, $valueName, $message = null, $context = null)
     {
+        $fieldName = self::$fieldName;
+        if ($context AND $arg = $context->getArgument() AND $arg instanceof FieldArgument) {
+            $argName = $arg->name;
+            return "Field \"{$fieldName}\" argument \"{$argName}\" requires type {$typeName}, found {$valueName}" . ($message ? "; {$message}" : '.');
+        }
```
vladar: This message should be in a separate function (like `badArgumentValue`). Also, those functions shouldn't contain any logic related to errors - only formatting logic. The logic of triggering an error should be defined in the visitor.

src/Validator/Rules/ValuesOfCorrectType.php
```diff
@@ -56,12 +65,17 @@ static function unknownFieldMessage($typeName, $fieldName, $message = null)
     public function getVisitor(ValidationContext $context)
     {
         return [
```
vladar: You could put `$fieldName` at the top of the visitor function and then `use` it inside closures instead of having it defined as static.

src/Validator/Rules/ValuesOfCorrectType.php
yura3d: Resolved in 9dc41f0

src/Validator/Rules/ValuesOfCorrectType.php
yura3d: Resolved in 9dc41f0

src/Validator/Rules/ValuesOfCorrectType.php
yura3d: Resolved in 9dc41f0

yura3d:
The changes have been made

vladar:
Thanks! Hope it will improve the UX for all library users.

PR #381, opened 2018-10-26
Fix resolve Fn key

jbtbnl:
Otherwise all existing resolve functions on fields get lost.

### DIFF
--- a/src/Utils/SchemaExtender.php
+++ b/src/Utils/SchemaExtender.php
@@ -350,7 +350,7 @@ class SchemaExtender
                 'type' => static::extendType($field->getType()),
                 'args' => static::extendArgs($field->args),
                 'astNode' => $field->astNode,
-                'resolveFn' => $field->resolveFn,
+                'resolve' => $field->resolveFn,
             ];
         }
 


vladar:
Can you add a test case for this to avoid regressions in the future?

jbtbnl:
@vladar @simPod this is a test case:

```php
<?php

$queryType = new \GraphQL\Type\Definition\ObjectType([
	'name' => 'Query',
	'fields' => [
		'hello' => [
			'type' => \GraphQL\Type\Definition\Type::string(),
			'resolve' => function() {
				return 'Hello World!';
			}
		],
	]
]);

$schema = new \GraphQL\Type\Schema([
	'query' => $queryType,
]);

$documentAST = \GraphQL\Language\Parser::parse('
extend type Query {
	misc: String
}
');

$extendedSchema = \GraphQL\Utils\SchemaExtender::extend($schema, $documentAST);
$helloResolveFn = $extendedSchema->getQueryType()->getField('hello')->resolveFn;

$assertTrue = call_user_func($helloResolveFn) === 'Hello World!';
```

Please advice where to add this. Or feel free to copy/alter it.

simPod:
It seems weird to me that accessor `->resolveFn` is not consistent with `'resolve'` key 🤔

vladar:
Thanks!

PR #392, opened 2018-11-11
Fix possessive its vs contraction it's

markhuot:


nagledb:
The change is making it fail a unit test. You need update the expected output for function testWarnsAboutSlowIsTypeOfForLazySchema in tests/Executor/ExecutorLazySchemaTest.php.

markhuot:
Tests should pass now. Thank you and sorry about that.

vladar:
Thanks!

PR #399, opened 2018-11-21
Update index.md

robsontenorio:
It is great to see that another packages are built on top of this very solid package.


vladar:
We actually have a dedicated page for such tools at https://github.com/webonyx/graphql-php/blob/master/docs/complementary-tools.md

PR #406, opened 2018-11-27
Added all possible scalar types to Node constructor

simPod:
When constructing BooleanValueNode

```php
new BooleanValueNode(['value' => false]))
```

This error occured

> Parameter #1 $vars of class GraphQL\Language\AST\BooleanValueNode constructor expects array<GraphQL\Language\AST\Location|GraphQL\Language\AST\NameNode|GraphQL\Language\AST\NodeList|GraphQL\Language\AST\SelectionSetNode|string|null>, array<string, false> given.

### DIFF
--- a/src/Language/AST/Node.php
+++ b/src/Language/AST/Node.php
@@ -40,7 +40,7 @@ abstract class Node
     public $loc;
 
     /**
-     * @param (string|NameNode|NodeList|SelectionSetNode|Location|null)[] $vars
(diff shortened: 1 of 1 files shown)


vladar:
What kind of error is this? Is it PHPStan?

simPod:
Yes. Occured in my library where I use PHPStan lvl 7.

----------------

_I'd like to move PHPStan here to higher levels as well but we must be careful cuz performance :D 1st step is here https://github.com/webonyx/graphql-php/pull/364_

PR #413, opened 2018-12-02
Apollo persisted query compatibility: Look for queryId in extensions.

chriszarate:
Apollo Engine's implementation of [automatic persisted queries](https://blog.apollographql.com/improve-graphql-performance-with-automatic-persisted-queries-c31d27b8e6ea) expects the query hash in the operation's `extensions`. In JSON request body terms, it looks like this:

```
{
  "operationName": "MyQuery",
  "variables": null,
  "extensions": {
    "persistedQuery": {
      "sha256Hash": "9F86D081884C7D659A2FEAA0C55AD015A3BF4F1B2B0B822CD15D6C15B0F00A08",
      "version": 1
    }
}
```

Naturally, [Apollo client](https://github.com/apollographql/apollo-link-persisted-queries#protocol) follows this protocol. Considering Apollo's popularity and the large number of users likely to use Apollo client to query `graphql-php`-powered servers, it seems useful to provide out-of-the-box compatibility for this protocol.

This PR looks for the query hash in the operation's extensions and copies it to the `queryId` (if neither a `query` nor `queryId` were separately provided). Implementors are then able to implement `persistentQueryLoader` for Apollo clients.

----

After some more thought, I think it makes sense to add `extensions` to the public API of `OperationParams`. This would allow easier implementation of other and future extensions. And if you disagreed that we should support Apollo's protocol out-of-the-box, this would still provide an easy implementation path for persisted queries since the operation params are passed to `persistentQueryLoader`.

### DIFF
--- a/src/Server/OperationParams.php
+++ b/src/Server/OperationParams.php
@@ -95,6 +95,11 @@ class OperationParams
         $instance->variables = $params['variables'];
         $instance->readOnly  = (bool) $readonly;
 
+        // Apollo server/client compatibility: look for the queryid in extensions
+        if (isset($params['extensions']['persistedQuery']['sha256Hash']) && empty($instance->query) && empty($instance->queryid)) {
+          $instance->queryId = $params['extensions']['persistedQuery']['sha256Hash'];
+        }
+
         return $instance;
     }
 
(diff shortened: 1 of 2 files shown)


vladar:
Thanks!

Daniel15:
Do you have an example of how to implement automatic persisted queries? How do you return the `PersistedQueryNotFound` error, and how do you actually persist it? I guess it can be persisted to APC or Memcache or something similar.

Issue #180, opened 2017-10-10
Schema extensions utility?


greyexpert:
Hello there,

Do you have any plans of porting `extendSchema` utility from `graphql-js`?
https://github.com/graphql/graphql-js/blob/master/src/utilities/extendSchema.js

It would be great to have ability to build modular schema and to extend type definitions using `extend type`.

vladar:
That would be nice, but it's unlikely we'll port it soon (too little time now). So PRs are welcome!

crirus:
Hello guys,

I already have the code ported along with a merging/stitching function taken from Apollo. It's not 100 percent as they keep pushing more features but I can drop it on a repository so you can play with it and maybe help polishing the code

greyexpert:
It work this way in `graphql-js` at the moment:

The `extend type` statement is not processed when you build schema in `graphql-js` using `buildASTSchema`. The statement is processed only in `extendSchema` which accepts already built root schema and applies extensions to it.

What do you think, if we follow a slightly different approach:

We can add the support of extensions into `BuildSchema` itself without adding an extra utility like `ExtendSchema`. Having this feature supported, we will be able to build modular schema by just concatinating multiple schemas before passing them into `BuildSchema::build`

For example:

RootSchema.graphqls
```graphql
schema {
  query: Query
}

type Query {
  version: String
}
```

UsersSchema.graphqls
```graphql
type User {
  name: String
}

extend type Query {
  users: [User]
}
```

Then, all we need is to concatenate them into result schema and pass to `BuildSchema::build`

```php
$schemaSources = [
    file_get_contents("RootSchema.graphqls"),
    file_get_contents("UsersSchema.graphqls")
];
        
$schema = BuildSchema::build(implode("\n\n", $schemaSources));
```
This feature will require less time, comparing to developing new utility. Also there are a lot of useful private methods in `BuildSchema` class, which can be re-used.

And, yes, I can create a pull request, if you guys don't mind this feature?

Related discussion in `graphql-js` repo:
https://github.com/graphql/graphql-js/issues/922

greyexpert:
@crirus 
It would be great if you share your implementation

crirus:
@greyexpert  here is the initial code, any questions let me know
https://github.com/crirus/graphql-php-tools

MidnightDesign:
[...]

PR #417, opened 2018-12-05
BUGFIX: ExtendSchema should be able to introduce new types

Torsten85:
Currently ExtendSchema throws if the new document node has types that are not present in the base schema. 

Described in this issue: https://github.com/webonyx/graphql-php/issues/180#issuecomment-444478026

### DIFF
--- a/src/Utils/SchemaExtender.php
+++ b/src/Utils/SchemaExtender.php
@@ -513,7 +513,14 @@ class SchemaExtender
                 $schemaExtensions[] = $def;
             } elseif ($def instanceof TypeDefinitionNode) {
                 $typeName = isset($def->name) ? $def->name->value : null;
-                if ($schema->getType($typeName)) {
+
+                try {
+                    $type = $schema->getType($typeName);
+                } catch (Error $error) {
+                    $type = null;
+                }
+
+                if ($type) {
                     throw new Error('Type "' . $typeName . '" already exists in the schema. It cannot also be defined in this type definition.', [$def]);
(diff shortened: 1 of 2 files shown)


chriseskow:
@vladar Any chance you can release a new version with this fix soon?

src/Utils/SchemaExtender.php
```diff
@@ -513,7 +513,14 @@ public static function extend(Schema $schema, DocumentNode $documentAST, ?array
...
                 $typeName = isset($def->name) ? $def->name->value : null;
-                if ($schema->getType($typeName)) {
+
+                try {
+                    $type = $schema->getType($typeName);
+                } catch (Error $error) {
+                    $type = null;
+                }
+
+                if ($type) {
```
simPod: if type what? I'd prefer to see more explicit and therefore readable condition. `$type !== null` ? 🤔

tests/Utils/SchemaExtenderTest.php
```diff
@@ -1962,4 +1963,48 @@ public function testOriginalResolversArePreserved()
...
+            type Foo {
+              value: Int
+            }
+            
+            type Query {
+              defaultValue: String
+            }
+        ';
+
+        static::assertEquals($this->dedent($expected), SchemaPrinter::doPrint($extendedSchema));
```
simPod: maybe simply `self::` would do? I guess `static::` works too but `self` is more common and I don't see a reason for `static`

PR #421, opened 2018-12-27
Fix new executor is removing associative array when using custom scalar

JanBukva:
Hi, I just came across that the new CoroutineExecutor is converting associative arrays to simple array, if such array would ever happen in your schema. The ReferenceExecutor is keeping the associative array.

**Failing test**
```
1) GraphQL\Tests\Executor\ExecutorSchemaTest::testExecutesUsingASchema
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
             'body' => 'This is a post'
             'author' => Array (...)
             'meta' => Array (
-                'title' => 'My Article 1 | My Blog'
+                0 => 'My Article 1 | My Blog'
             )
         )
     )
 )
```

**Example data / my use case:**
In my side project I wanted to pass cropped image URLs for my js component, but the list is dynamic (some images have only 1 configuration, some can have 20 different ones).

```json
{
  "id": "4574BB",
  "variantUrls": {
    "00": "/uploads/00/16029619-136d-4110-a31c-3e00c06cb482",
    "cb": "/uploads/cb/16029619-136d-4110-a31c-3e00c06cb482",
    "og": "/uploads/og/aa6f99d0-4710-4cf1-8e30-604d47be5cf4",
  }
}
```

When I created simple Scalar type, that was just passing variable out (as I trust the associative array from backend) the new Coroutine Executor is changing associative array to simple array without keys.

I probably should have created an ObjectType for that, but this way is slightly simpler to foreach, not having to check for nulls.

---

After I made all of this I noticed, that if I wrap the associative array in some kind of stdClass, this problem disappears. But at least I found out about this difference between executors.

### DIFF
--- a/src/Experimental/Executor/CoroutineExecutor.php
+++ b/src/Experimental/Executor/CoroutineExecutor.php
@@ -34,11 +34,8 @@ use GraphQL\Utils\Utils;
 use SplQueue;
 use stdClass;
 use Throwable;
-use function array_keys;
-use function count;
 use function is_array;
 use function is_string;
-use function range;
 use function sprintf;
 
 class CoroutineExecutor implements Runtime, ExecutorImplementation
@@ -154,14 +151,9 @@ class CoroutineExecutor implements Runtime, ExecutorImplementation
         }
 
         if (is_array($value)) {
-            $array   = [];
-            $isAssoc = array_keys($value) !== range(0, count($value) - 1);
+            $array = [];
             foreach ($value as $key => $item) {
-                if ($isAssoc) {
-                    $array[$key] = self::resultToArray($item);
-                } else {
-                    $array[] = self::resultToArray($item);
-                }
+                $array[$key] = self::resultToArray($item);
             }
             return $array;
         }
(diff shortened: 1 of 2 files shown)


vladar:
@jakubkulhan Could you please review it?

src/Experimental/Executor/CoroutineExecutor.php
```diff
@@ -151,9 +154,14 @@ private static function resultToArray($value, $emptyObjectAsStdClass = true)
         }
 
         if (is_array($value)) {
-            $array = [];
-            foreach ($value as $item) {
-                $array[] = self::resultToArray($item);
+            $array   = [];
+            $isAssoc = array_keys($value) !== range(0, count($value) - 1);
```
simPod: I wonder, what if I had an assoc array like for example `[0=>0,1=>1]`? IMO there's no way to detect whether it's "associative" or not. What would have happened if `$array[$key] = self::resultToArray($item);` was used just and only?

PR #423, opened 2018-12-27
Fix FormattedError

simPod:


vladar:
This is a breaking change which seems unnecessary at the moment. We can probably merge it before the next major release, so I'll keep it open until then.

simPod:
Yup, no rush. I'm continuously trying to add typechecks and reduce type variations like here `int OR bool` etc.

UPGRADE.md
```diff
@@ -233,7 +233,7 @@ exceptions were reported to clients (e.g. database connection errors, file acces
 Also every error reported to client now has new `category` key which is either `graphql` or `internal`.
 Exceptions implementing `ClientAware` interface may define their own custom categories.
 
-During development or debugging use `$executionResult->toArray(true)`. It will add `debugMessage` key to 
+During development or debugging use `$executionResult->toArray(DebugFlag::INCLUDE_DEBUG_MESSAGE)`. It will add `debugMessage` key to
```
vladar: This shouldn't change (as it relates to migration to version 0.10.x. Instead could you add a separate entry in Changelog.md and Upgrade.md about this change?

vladar:
Can you rebase this and add a proper note about breaking part in the Changelog (if this whole change still makes sense of course)? I was hoping to publish a new major at the week-end but not sure if it should go there. Thoughts?

vladar:
Resolved conflicts, merged manually. Thanks for the PR!

PR #424, opened 2018-12-27
Fix Error

simPod:


vladar:
This is also an unnecessary breaking change. I'll postpone it till the next major release.

UPGRADE.md
```diff
@@ -15,7 +15,7 @@ $e = new Error(
     'msg',
     null,
     null,
-    null,
+    [],
```
vladar: This shouldn't be changed as it relates to migration to version `0.13.x`. Instead, we should have a new entry in Upgrade.md for this specific change.

vladar:
Same for this PR, must be rebased first

vladar:
Merged this manually via https://github.com/webonyx/graphql-php/commit/0fa6330f9f9ccc93ad90a049bd5b8dce38d92181

PR #427, opened 2019-01-05
Preserve array indices when computing path for ListOf type

shmax:
I noticed that the behavior of `ReferenceExecutor.completeListValue` is slightly different than the equivalent method in the reference library. The reference library doesn't alter the array keys as it builds up `path`. It uses `forEach` from Lee Byron's own "iterall" library, which preserves the original array indices. 

Here's the relevant section of code from your `completeListValue`:

```php
        $i               = 0;
        $completedItems  = [];
        foreach ($result as $item) {
            $fieldPath     = $path;
            $fieldPath[]   = $i++;
```

You can see that any existing array indices are ignored, and they are renumbered starting from 0.

This is causing some issues with my dynamically generated user error types (I expose the array position of the error for `ListOf` types to the client). Please let me know if you would like to see some sample code or further analysis.

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -917,11 +917,11 @@ class ReferenceExecutor implements ExecutorImplementation
             'User Error: expected iterable, but did not find one for field ' . $info->parentType . '.' . $info->fieldName . '.'
         );
         $containsPromise = false;
-        $i               = 0;
+
         $completedItems  = [];
-        foreach ($result as $item) {
+        foreach ($result as $idx => $item) {
             $fieldPath     = $path;
-            $fieldPath[]   = $i++;
+            $fieldPath[]   = $idx;
             $completedItem = $this->completeValueCatchingError($itemType, $fieldNodes, $info, $fieldPath, $item);
             if (! $containsPromise && $this->getPromise($completedItem)) {
                 $containsPromise = true;


mfn:
Tests are absolutely required IMHO for this. Otherwise it will silently break in the future.

src/Executor/ReferenceExecutor.php
```diff
@@ -917,11 +917,11 @@ private function completeListValue(ListOfType $returnType, $fieldNodes, ResolveI
...
         );
         $containsPromise = false;
-        $i               = 0;
+
         $completedItems  = [];
-        foreach ($result as $item) {
+        foreach ($result as $idx => $item) {
             $fieldPath     = $path;
-            $fieldPath[]   = $i++;
+            $fieldPath[]   = $idx;
```
simPod: I'd rather not use abbreviations. `$index` would be 👌

shmax: SGTM

shmax:
> Tests are absolutely required IMHO for this.

Sure, no objection. The reference library doesn't have tests around this particular use case, and I was just following their lead. I'll come up with something.

shmax:
I added a test, which was a good idea because it's making me rethink things. This PR does bring our functionality inline with the reference library, but I'm thinking that the bug is actually in the reference library, and that this codebase is correct. 

Let's have another look at the reference library code:

```
function completeListValue(
  exeContext: ExecutionContext,
  returnType: GraphQLList<GraphQLOutputType>,
  fieldNodes: $ReadOnlyArray<FieldNode>,
  info: GraphQLResolveInfo,
  path: ResponsePath,
  result: mixed,
): MaybePromise<$ReadOnlyArray<mixed>> {
  invariant(
    isCollection(result),
    `Expected Iterable, but did not find one for field ${
      info.parentType.name
    }.${info.fieldName}.`,
  );

  // This is specified as a simple map, however we're optimizing the path
  // where the list contains no Promises by avoiding creating another Promise.
  const itemType = returnType.ofType;
  let containsPromise = false;
  const completedResults = [];
  forEach((result: any), (item, index) => {
    // No need to modify the info object containing the path,
    // since from here on it is not ever accessed by resolver functions.
    const fieldPath = addPath(path, index);
    const completedItem = completeValueCatchingError(
      exeContext,
      itemType,
      fieldNodes,
      info,
      fieldPath,
      item,
    );

    if (!containsPromise && isPromise(completedItem)) {
      containsPromise = true;
    }
    completedResults.push(completedItem);
  });
```

You can see that while the keys are preserved for the purposes of computing `path`, they are not preserved for the result that is returned from the resolver. This means that if you were to write a resolver that prepares an array of mostly null rows and happened to call `array_filter` (which preserves keys) before returning the result, then the `path` would have an index in it that refers to something that is not in the result set.

While that would be convenient for the code in my own project, it's probably not really correct.

What do you think? Is this an oversight on the reference library?

vladar:
Keep in mind that lists in JS and arrays in PHP are a bit different. You can't have gaps in JS arrays. If you write:
```js
let a = [];
a[2] = 'foo';
```
You will actually get a list of `[undefined, undefined, 'foo']`. So indexes in the reference implementation are actually implicitly preserved, so no error there.

shmax:
> You will actually get a list of [undefined, undefined, 'foo'] which is not the case with PHP arrays. So indexes in the reference implementation are actually implicitly preserved, so no error there.

Sure, but it seems that `forEach` doesn't consider the non-existent array enties when iterating. I did attempt to verify this behavior for myself before raising this issue. Here's a fiddle:

http://jsfiddle.net/epaz1m50/4/

To be fair, I have not gone as far to prepare a javascript environment to verify that it is possible for the indices to get out of sync. That's a little more effort than I want to put into it at this point, but I might see if someone at work can try it out for me.

I'll let you know. Thanks for the reply!

tests/Executor/DeferredFieldsTest.php
```diff
@@ -196,6 +196,13 @@ static function ($story) {
                         );
                     },
                 ],
+                'topStoriesSparse'       => [
+                    'type'    => Type::listOf($this->storyType),
+                    'resolve' => function ($val, $args, $context, ResolveInfo $info) {
```
simPod:
```suggestion
                    'resolve' => static function ($val, array $args, $context, ResolveInfo $info) : array {
```

vladar:
Yeah, you are correct, `forEach` seems to ignore keys. Then it could indeed be a potential issue in the reference implementation.

Anyways as for our library, I've posted you an example explaining why we reset keys.

mfn:
@shmax still if you ever follow up with the reference implementation and have or create an issue there, would be nevertheless be interesting to follow up! thanks

Issue #425, opened 2018-12-29
Lazy loading of types...


shmax:
I have a large number of types (130+) in my code base, and I'm worried about the overhead of building them all for every little request. I followed the advice in your documentation, and made sure that my objects return callbacks for their `field` properties, and I provided a `typeLoader` callback when instantiating my `Schema`. I understand the general concept, but I must be missing a piece, because at least some types are still being constructed even though they are not referenced in any way by the query being made.

After analyzing the sample [here](http://webonyx.github.io/graphql-php/type-system/schema/#lazy-loading-of-types) it's not clear to me where the savings are. Imagine if `MyTypeA` had a list of 100 fields on it instead of 1; wouldn't 100 calls to `$this->get(<fieldName>)` trigger as soon as its `fields` callback resolves? Or is there some magic I'm missing?

I guess I was expecting something more like this:

```php
private function MyTypeA()
    {
        return new ObjectType([
            'name' => 'MyTypeA',
            'fields' => function() {
                return [
                    'b' => ['type' => 'MyTypeB'] // use a string (or possibly a callable) here, rather than resolve the type on the spot
                ];
            }
        ]);
    }
```

...with the idea being that I use the name of the type in my field definition rather than a reference to the field itself, and then at runtime it would use the name to look up the type when needed.

I'm pretty sure this is just a matter of me being dumb and not getting something fundamental, so I guess my real question here is whether you think the lazy loading documentation could be improved a bit, or perhaps incorporated into the samples. If you can help me get my questions sorted out I'd be happy to help with either.

Thanks for the wonderful library!

vladar:
You are right here. Types are being initiated on `fields` closure call. So even if you request single field of a type, all types for those fields are being instantiated.

But instance creation is not expensive. It's fields, arguments, and interfaces which add significant overhead. If you don't create them on type instantiation you should get a significant performance win (which we benchmarked).

We actually [considered](https://github.com/webonyx/graphql-php/issues/62) loading types by name, but it didn't pass a cost/benefit analysis at the moment (wins were little, costs to implement were relatively significant).

Anyways, if you do some benchmarks, illustrating the opposite we can get back to this idea.

shmax:
> But instance creation is not expensive. It's fields, arguments, and interfaces which add significant overhead. 

[...]

PR #429, opened 2019-01-07
Lazy loading for arguments

mad-ice:
Hi!

This pull request ensures that field arguments can be lazy loaded in the same manner as fields, which should improve performance quite a bit. I've ran a majorly tiny benchmark on the performance impact and found out that this change decreased our schema build time from 28ms to 3ms.

I'd like to get some help on testing this feature as I'm not familiar with maintaining automated tests and got lost trying to find the right place to add them.

And, of course, a big thank you for the amazing library!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ composer.phar
 phpcs.xml
 phpstan.neon
 vendor/
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ composer.phar
 phpcs.xml
 phpstan.neon
 vendor/
-.idea/
--- a/src/Type/Definition/FieldArgument.php
+++ b/src/Type/Definition/FieldArgument.php
@@ -70,6 +70,7 @@ class FieldArgument
      * @param mixed[]|callable $config
      *
      * @return FieldArgument[]
+     * @throws \InvalidArgumentException
      */
     public static function createMap($config)
     {
--- a/src/Type/Definition/FieldArgument.php
+++ b/src/Type/Definition/FieldArgument.php
@@ -8,6 +8,7 @@ use GraphQL\Error\InvariantViolation;
 use GraphQL\Language\AST\InputValueDefinitionNode;
 use GraphQL\Utils\Utils;
 use function is_array;
+use function is_callable;
 use function is_string;
 use function sprintf;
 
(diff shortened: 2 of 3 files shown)


mad-ice:
Might be related to @shmax's issue: https://github.com/webonyx/graphql-php/issues/425

.gitignore
```diff
@@ -4,3 +4,4 @@ composer.phar
 phpcs.xml
 phpstan.neon
 vendor/
+.idea/
```
simPod: Hi, for this is not project related you should add it to your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

mad-ice: Didn't know that existed, thanks!

shmax:
Ah, this looks promising! I have to head to work, but when I get back I'll pull your branch and run my profiler. Very excited to see the results...

.gitignore
```diff
@@ -3,4 +3,4 @@ composer.lock
 composer.phar
 phpcs.xml
 phpstan.neon
-vendor/
+vendor/
```
simPod: we're missing a line feed

If you're using Jetbrains IDE, it can be automatically handled by `Editor -> General -> Other`

image

src/Type/Definition/FieldArgument.php
```diff
@@ -66,12 +67,21 @@ public function __construct(array $def)
...
      */
-    public static function createMap(array $config)
+    public static function createMap($config)
     {
+        if (is_callable($config)) {
+            $config = $config();
+        }
+
+        if (! is_array($config)) {
+            throw new \InvalidArgumentException('$config must be an array or a callable which returns such an array.');
```
simPod:
```suggestion
            throw new \InvalidArgumentException('"$config" must be an array or a callable which returns such an array.');
```

spawnia:
@mad-ice to move this issue along, you can add a test to https://github.com/webonyx/graphql-php/blob/master/tests/Type/DefinitionTest.php

Check out https://github.com/webonyx/graphql-php/blob/master/CONTRIBUTING.md for more information on the dev tooling. Try running `composer check-all` on the project.

src/Type/Definition/FieldArgument.php
```diff
@@ -66,12 +67,21 @@ public function __construct(array $def)
...
-     * @param mixed[] $config
+     * @param mixed[]|callable $config
      *
      * @return FieldArgument[]
+     * @throws \InvalidArgumentException
      */
-    public static function createMap(array $config)
+    public static function createMap($config)
     {
+        if (is_callable($config)) {
```
shmax: Sorry for the long delay, but as promised I tried out your changes. So far I don't quite understand how this changes things; `createMap` is called by the `FieldDefinition` constructor, and then you just check to see if `$config` is callable, and if it is you immediately call it? Where do the savings come in? 

I tested this by converting a few of the `args` on my top-level `Query` fields to functions, and stuck breakpoints inside them; they get called immediately.

What am I missing, here?

shmax: I just POC'ed my own full lazy loading concept; so far seems to work pretty well. I'll try to put together a PR by the end of the week.

vladar:
Thanks for the PR. But we will likely go the path suggested in #557 which addresses an issue with lazy loading of individual types

PR #431, opened 2019-01-10
Invert instance of in BCFinder

simPod:
I tried to construct a schema from remote endpoint and the problem was that remote types were then constructed as `GraphQL\Type\Definition\ObjectType` where types from local schema were constructed as object from local namespace that exteds `GraphQL\Type\Definition\ObjectType`.

Therefore this check didn't pass and they were marked as BC breaks but they're not.


vladar:
I think such change requires a test case with a link to this PR. Otherwise, it can get erased when we port changes from newer reference implementation versions.

PR #435, opened 2019-01-28
Fix return annotation of resolveType() in InterfaceType

JanBukva:
Hi,
When I run phpstan on my project I found out about wrong return type on InterfaceType::resolveType(). It should return _resolved_ Type | null, not callable, as the resolving is done inside that method.

I'm not sure if you want to be more specific about the returned type.


vladar:
You are right, thanks!

Issue #65, opened 2016-11-04
Introduce query plan as a replacement for ResolveInfo::getFieldSelection


vladar:
Currently we have `ResolveInfo::getFieldSelection()` to do simple look-aheads. But it is too limited for real-world scenarios and provides little help when resolving field of union or interface type. Better tool has to address more use-cases. Basically it should provide convenient interface to answer following questions:

1. What sub-fields were requested (regardless of fragment types)? 
2. What types were referenced by fragments?
3. What sub-fields were requested for given type?
4. What args/directives were set for given field?

Current `getFieldSelection` only answers question 1. 

The syntax for this could be something like this:
```php
function resolve($value, $args, $context, ResolveInfo $info) {
    $queryPlan = $info->lookAhead();

    // Answers #1:
    $fields = $queryPlan->subFields();
    $deepFieldRequested = $queryPlan->has(['nested', 'deeper', 'evenDeeper']);

    // Answers #2:
    $types = $queryPlan->referencedTypes();

    // Answers #3:
    $typeSpecificFields = $queryPlan->subFields('TypeReferencedByFragment');
}
```

This needs further design with regards to nested syntax + things may get quite complicated for type-dependent deep fields.

decebal:
@vladar I had to write a wrapper around `ResolveInfo` In my use case that answers  #4 in a similar way `foldSelectionSet` does.

decebal:
@vladar sure thing, I'll do the pull request asap (this week), while I really need to start working on this for my project as it currently has some only hacks in order to cover for missing type definitions in resolveInfo

PR #436, opened 2019-02-01
add queryplan

keulinho:
Provide the possibility to lookahead the Query Execution.
This one adds the API proposed in [#65](https://github.com/webonyx/graphql-php/issues/65).
From #65:

1. What sub-fields were requested (regardless of fragment types)? -> still possible
2. What types were referenced by fragments? -> could be added
3. What sub-fields were requested for given type? -> possible
4. What args/directives were set for given field? -> possible for args, directives could be added

### DIFF
--- a/src/Type/Definition/QueryPlan.php
+++ b/src/Type/Definition/QueryPlan.php
@@ -1,0 +1,239 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Type\Definition;
+
+use GraphQL\Error\Error;
+use GraphQL\Executor\Values;
+use GraphQL\Language\AST\FieldNode;
+use GraphQL\Language\AST\FragmentDefinitionNode;
+use GraphQL\Language\AST\FragmentSpreadNode;
+use GraphQL\Language\AST\InlineFragmentNode;
+use GraphQL\Language\AST\SelectionSetNode;
+use GraphQL\Type\Schema;
+use function array_filter;
+use function array_key_exists;
+use function array_keys;
+use function array_merge;
+use function array_merge_recursive;
+use function array_unique;
+use function array_values;
+use function count;
+use function in_array;
+use function is_array;
+use function is_numeric;
+
+class QueryPlan
+{
+    /** @var string[][] */
+    private $types = [];
+
+    /** @var Schema */
+    private $schema;
+
+    /** @var mixed[] */
+    private $queryPlan = [];
+
+    /** @var mixed[] */
+    private $variableValues;
+
+    /** @var FragmentDefinitionNode[] */
+    private $fragments;
+
+    /**
+     * @param FieldNode[]              $fieldNodes
+     * @param mixed[]                  $variableValues
+     * @param FragmentDefinitionNode[] $fragments
+     */
+    public function __construct(ObjectType $parentType, Schema $schema, iterable $fieldNodes, array $variableValues, array $fragments)
+    {
+        $this->schema         = $schema;
+        $this->variableValues = $variableValues;
+        $this->fragments      = $fragments;
+        $this->analyzeQueryPlan($parentType, $fieldNodes);
+    }
+
+    /**
+     * @return mixed[]
+     */
+    public function queryPlan() : array
+    {
+        return $this->queryPlan;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getReferencedTypes() : array
+    {
+        return array_keys($this->types);
+    }
+
+    public function hasType(string $type) : bool
+    {
+        return count(array_filter($this->getReferencedTypes(), static function (string $referencedType) use ($type) {
+                return $type === $referencedType;
+        })) > 0;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getReferencedFields() : array
+    {
+        return array_values(array_unique(array_merge(...array_values($this->types))));
+    }
+
+    public function hasField(string $field) : bool
+    {
+        return count(array_filter($this->getReferencedFields(), static function (string $referencedField) use ($field) {
+            return $field === $referencedField;
+        })) > 0;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function subFields(string $typename) : array
+    {
+        if (! array_key_exists($typename, $this->types)) {
+            return [];
+        }
+
+        return $this->types[$typename];
+    }
+
+    /**
+     * @param FieldNode[] $fieldNodes
+     */
+    private function analyzeQueryPlan(ObjectType $parentType, iterable $fieldNodes) : void
+    {
+        $queryPlan = [];
+        /** @var FieldNode $fieldNode */
+        foreach ($fieldNodes as $fieldNode) {
+            if (! $fieldNode->selectionSet) {
+                continue;
+            }
+
+            $type = $parentType->getField($fieldNode->name->value)->getType();
+            if ($type instanceof WrappingType) {
+                $type = $type->getWrappedType();
+            }
+
+            $subfields = $this->analyzeSelectionSet($fieldNode->selectionSet, $type);
+
+            $this->types[$type->name] = array_unique(array_merge(
+                array_key_exists($type->name, $this->types) ? $this->types[$type->name] : [],
+                array_keys($subfields)
+            ));
+
+            $queryPlan = array_merge_recursive(
+                $queryPlan,
+                $subfields
+            );
+        }
+
+        $this->queryPlan = $queryPlan;
+    }
+
+    /**
+     * @return mixed[]
+     *
+     * @throws Error
+     */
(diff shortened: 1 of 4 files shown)


tests/Type/QueryPlanTest.php
```diff
@@ -0,0 +1,588 @@
...
+        $result = GraphQL::executeQuery($schema, $doc)->toArray();
+
+        self::assertTrue($hasCalled);
+        self::assertEquals(['data' => ['article' => null]], $result);
+        self::assertEquals($expectedQueryPlan, $queryPlan->queryPlan());
+        self::assertEquals($expectedReferencedTypes, $queryPlan->referencedTypes());
+        self::assertEquals($expectedReferencedFields, $queryPlan->referencedFields());
+        self::assertEquals(['url', 'width', 'height'], $queryPlan->subFields('Image'));
+
+        static::assertTrue($queryPlan->hasField('url'));
```
mfn: What's up with the `self::` vs. `static::` calls here?

Maybe it's through out the codebase and I just didn't notice it before 🤔

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+
+    /** @var FragmentDefinitionNode[]|null */
+    private $fragments;
+
+    /**
+     * @param FieldNode[]|null         $fieldNodes
+     * @param mixed[]                  $variableValues
+     * @param FragmentDefinitionNode[] $fragments
+     *
+     * @throws Error
```
simPod: IMO the @throws anotation can be dropped. I have never found a way how to sustainably maintain those across the whole project.

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+     * @return string[]
+     */
+    public function referencedTypes() : array
+    {
+        return array_keys($this->types);
+    }
+
+    public function hasType(string $type) : bool
+    {
+        return count(array_filter($this->referencedTypes(), static function ($element) use ($type) {
```
simPod: I guess we can typehint `string $element`. And then change variable name to `$referencedType`

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+     */
+    public function queryPlan() : array
+    {
+        return $this->queryPlan;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function referencedTypes() : array
```
simPod: I'd prefix with `get`

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+    {
+        return count(array_filter($this->referencedTypes(), static function ($element) use ($type) {
+                return $type === $element;
+        })) > 0;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function referencedFields() : array
```
simPod: prefix `get` as well.

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+    {
+        return array_values(array_unique(array_merge(...array_values($this->types))));
+    }
+
+    public function hasField(string $field) : bool
+    {
+        return count(array_filter($this->referencedFields(), static function ($element) use ($field) {
```
simPod:
```suggestion
        return count(array_filter($this->referencedFields(), static function (string $referencedField) use ($field) {
```

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+
+        return $this->types[$typename];
+    }
+
+    /**
+     * @param FieldNode[]|null $fieldNodes
+     *
+     * @throws Error
+     */
+    private function analyzeQueryPlan(ObjectType $parentType, $fieldNodes) : void
```
simPod:
```suggestion
    private function analyzeQueryPlan(ObjectType $parentType, ?array $fieldNodes) : void
```

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+            return [];
+        }
+
+        return $this->types[$typename];
+    }
+
+    /**
+     * @param FieldNode[]|null $fieldNodes
+     *
+     * @throws Error
```
simPod: can be dropped

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+     */
+    private function analyzeQueryPlan(ObjectType $parentType, $fieldNodes) : void
+    {
+        $queryPlan = [];
+        /** @var FieldNode $fieldNode */
+        foreach ($fieldNodes as $fieldNode) {
```
simPod: Please fix typehint or handle nullability, otherwise it's not safe to iterate `$fieldNodes`, not sure what's correct here :)

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+    /** @var Schema */
+    private $schema;
+
+    /** @var mixed[] */
+    private $queryPlan = [];
+
+    /** @var mixed[] */
+    private $variableValues;
+
+    /** @var FragmentDefinitionNode[]|null */
```
simPod: I think this cannot be null.

src/Type/Definition/QueryPlan.php
```diff
@@ -0,0 +1,246 @@
...
+
+                $fields[$fieldName] = [
+                    'type' => $selectionType,
+                    'fields' => $subfields ?? [],
+                    'args' => Values::getArgumentValues($type, $selectionNode, $this->variableValues),
+                ];
+            } elseif ($selectionNode instanceof FragmentSpreadNode) {
+                $spreadName = $selectionNode->name->value;
+                if (isset($this->fragments[$spreadName])) {
+                    /** @var FragmentDefinitionNode $fragment */
```
simPod: this can be dropped, no?

src/Type/Definition/ResolveInfo.php
```diff
@@ -99,6 +99,13 @@ class ResolveInfo
      */
     public $variableValues = [];
 
+    /**
+     * @internal
```
simPod: the var is private so I guess we can drop this

src/Type/Definition/ResolveInfo.php
```diff
@@ -99,6 +99,13 @@ class ResolveInfo
...
+     *
+     * @var QueryPlan
+     */
+    private $queryPlan = null;
```
simPod: null is default, can be dropped

src/Type/Definition/ResolveInfo.php
```diff
@@ -179,6 +186,22 @@ public function getFieldSelection($depth = 0)
 
         return $fields;
     }
+
+    public function lookahead() : QueryPlan
```
simPod: If meant as verb, the proper camel case is `lookAhead`

tests/Type/QueryPlanTest.php
```diff
@@ -0,0 +1,588 @@
...
+
+use GraphQL\GraphQL;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\QueryPlan;
+use GraphQL\Type\Definition\ResolveInfo;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Schema;
+use PHPUnit\Framework\TestCase;
+
+class QueryPlanTest extends TestCase
```
simPod:
```suggestion
final class QueryPlanTest extends TestCase
```

src/Type/Definition/QueryPlan.php
keulinho: `$fieldNodes` isn't an array but an `ArrayObject`, type hinted `iterable` instead

vladar:
This is great! I need some time to digest it before I can post any feedback or merge. Thanks!

tests/Executor/ExecutorTest.php
```diff
@@ -284,22 +284,6 @@ public function testProvidesInfoAboutCurrentExecutionState() : void
...
-                'path',
-                'schema',
-                'fragments',
-                'rootValue',
-                'operation',
-                'variableValues',
-            ],
-            array_keys((array) $info)
-        );
-
```
vladar: Why deleting this? Can't you just add `queryPlan` key to this list?

keulinho: `queryPlan` is declared private because it should only be loaded lazy when you call `lookAhead`
When you cast the `resolveInfo` to array now it cpontains some binary string instead of `queryPlan` because it is private.
Furthermore i really saw no value added by this test, because every key mentionied there would be used for own assertions, so the test will fail anyway if you remove one of the public properties.

vladar: Sorry, missed your reply. Makes sense.

PR #437, opened 2019-02-04
Update docs intro verbiage

stefanotorresi:
Just throwing my 2 cents: I don't think it's fair to say that "it's intended to be a replacement", given that multiple API paradigms can coexist in the same system and each of them have their trade-offs.


vladar:
Fair enough. Thanks!

PR #438, opened 2019-02-06
Fix incorrect array type of contextValue in PHPDocs

skodak:
Hi guys,

I was playing with GraphQL to learn more about it and I got to point where I ended up calling default field resolver but my IDE told me I have incorrect parameter for context. So I looked at the PHPDocs and some classes say context is "mixed|NULL" and others "mixed[]|NULL".

Here is a patch that marks the execution context parameter as "mixed|NULL" everywhere, I hope that is what was intended.

Cheers,
Petr

### DIFF
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -161,9 +161,9 @@ class Executor
      * and returns it as the result, or if it's a function, returns the result
      * of calling that function while passing along args and context.
      *
-     * @param mixed        $source
-     * @param mixed[]      $args
-     * @param mixed|null   $context
+     * @param mixed      $source
+     * @param mixed[]    $args
+     * @param mixed|null $context
      *
      * @return mixed|null
      */
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -119,7 +119,7 @@ class Executor
      *
      * Useful for async PHP platforms.
      *
-     * @param mixed[]|null $rootValue
+     * @param mixed|null   $rootValue
      * @param mixed|null   $contextValue
      * @param mixed[]|null $variableValues
      * @param string|null  $operationName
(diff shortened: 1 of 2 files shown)


skodak:
hmm, the fails in travis above do not seem to be related to my changes

src/Executor/Executor.php
```diff
@@ -163,7 +163,7 @@ public static function promiseToExecute(
      *
      * @param mixed        $source
      * @param mixed[]      $args
-     * @param mixed[]|null $context
+     * @param mixed|null   $context
```
simPod: you have 2 more spaces here

src/Executor/Executor.php
```diff
@@ -120,7 +120,7 @@ public static function execute(
      * Useful for async PHP platforms.
      *
      * @param mixed[]|null $rootValue
-     * @param mixed[]|null $contextValue
+     * @param mixed|null   $contextValue
```
vladar: Could you also change `mixed[]` to `mixed` for the `$rootValue` as it has the same issue?

src/Executor/ReferenceExecutor.php
```diff
@@ -116,7 +116,7 @@ public function doExecute() : Promise
      * execute, which we will pass throughout the other execution methods.
      *
      * @param mixed[]             $rootValue
-     * @param mixed[]             $contextValue
+     * @param mixed               $contextValue
```
vladar: `$rootValue` here should also be typed as `mixed` then

src/Executor/Executor.php
skodak: done, I have readjusted the whole param block

src/Executor/Executor.php
skodak: done

src/Executor/ReferenceExecutor.php
skodak: done

vladar:
Thanks!

PR #440, opened 2019-02-09
Test against PHP 7.4

simPod:


vladar:
Can we mark nightly builds as can fail? As they fail %)

simPod:
I suppose it it 🤔 

https://github.com/webonyx/graphql-php/pull/440/files#diff-354f30a63fb0907d4ad57269548329e3R35

simPod:
I guess that was not the proper way to define it. Now should be fine.

vladar:
Thanks!

PR #449, opened 2019-02-28
Add WP-GraphQL to complementary tools.

hughdevore:
We utilized this repo for our work in WP GraphQL and wanted to let WP users know that it's available for use.


vladar:
Awesome, thanks!

PR #455, opened 2019-03-14
Remove deprecated tools and improve Lighthouse description

spawnia:
- The folklore package was archived by the owner.
- The lighthouse relay tools were integrated with the core


vladar:
Thanks!

PR #465, opened 2019-03-28
Enum default values

spawnia:
This behaviour is consistent with the reference implementation https://github.com/graphql/graphql-js/issues/1604

Just wanted to emphasize that point in the docs and add a test to ensure the behaviour is well defined.


vladar:
Can you please fix [coding standard errors](https://travis-ci.org/webonyx/graphql-php/jobs/512443947)?

vladar:
Thanks!

PR #470, opened 2019-05-02
Drop PHP 7.1 support

simPod:
Sounds like bold move but support for [PHP 7.1 ended almost half year ago](https://www.php.net/supported-versions.php). Let's move forward. No issue here for legacy systems can continue using this lib as composer won't install platform-unsupported versions for them.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        php: [7.1, 7.2, 7.3, 7.4, 8.0]
+        php: [7.2, 7.3, 7.4, 8.0]
         env: [
           'EXECUTOR= DEPENDENCIES=--prefer-lowest',
           'EXECUTOR=coroutine DEPENDENCIES=--prefer-lowest',
@@ -68,7 +68,7 @@ jobs:
     - name: Install PHP
       uses: shivammathur/setup-php@2.9.0
       with:
-        php-version: 7.1
+        php-version: 7.2
         coverage: none
         extensions: json, mbstring
 
@@ -99,7 +99,7 @@ jobs:
     - name: Install PHP
       uses: shivammathur/setup-php@2.9.0
       with:
-        php-version: 7.1
+        php-version: 7.2
         coverage: none
         extensions: json, mbstring
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Unreleased
 
+- PHP version required: 7.2+
+
 #### 14.4.1
 
 Fix:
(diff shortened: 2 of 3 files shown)


vladar:
Makes sense for the next major version (14.0.0). Will merge it (along with some other pending PRs) as soon as I start working on it.

thoaionline:
I'm actually of the conflicting opinion here. I don't think a utility library should force its consumer to use the latest possible version of X only for the sake of it. PHP 7.1 may be EOL but there are always legacy projects somewhere using it, and a force upgrade to the next version will, unfortunately, cut off those projects from getting the latest enhancement/bugfixes.

A valid reason to force such upgrade would be:

* A: to address a relevant security vulnerability
* B: to make use of features that are only available in the new software release (PHP 7.2+ in this case).

**A** is not quite applicable at the language (PHP) level because an application is either on 7.1, or 7.2, and the people still stuck on 7.1 probably have one or more practical reasons why they can't yet move one to 7.3.

So I guess only **B** remains, and such upgrade will make sense only if this library needs PHP 7.2 specific functionality.

mfn:
> PHP 7.1 may be EOL

I think technically it is not (yet): https://www.php.net/supported-versions.php

7.0 is EOL, 7.1 isn't until 2019-12-01

simPod:
@thoaionline it doesn't force you, you can always use versions of this lib supporting it. Composer will install them for you.

SerafimArts:
I would recommend switching immediately to PHP 7.4 because this version contains a lot of useful functionality.

At one time, projects also used 7.1 as soon as it appeared as a stable version (like Doctrine and Laravel) for a long time, instead of 7.0 due to the presence of nullable types and `void` return typehint. 7.4 is the same version that adds lambdas, field typing, weak refs and other useful functionality.

vladar:
I had some conversations recently precisely about this question and people I spoke with raised the same concerns as @thoaionline 

PHP has a vast user base. Even old-school CMSes like Wordpress are exposing GraphQL APIs now (and many of the Wordpress sites are still on PHP5 - Wordpress itself will switch to PHP7 this winter as far as I know).

So this is getting a bit tricky - we basically cut them off from the new features. I know from my own experience that upgrading a huge codebase is not easy and that there are valid reasons not to upgrade for some time.

Generally speaking convenience for us, having the latest version, leads to pain for many users. So I will hold off this for a while.

simPod:
[Officially EOLed](https://twitter.com/krakjoe/status/1201010799034195968)
image

spawnia:
@simPod how about we integrate the changes from https://github.com/webonyx/graphql-php/pull/642 in this PR?

I think it will be easier to judge when we should make the switch if we bundle the nice changes it affords us.

simPod:
@spawnia what if I setup a `drop-php71` branch for that a we do the work on top of it?

simPod:
We missed the bump for 14 :/

PR #474, opened 2019-05-08
Ignore examples directory

theofidry:
With some graphQL plugins, the graphQL schemas declared in the examples mess things up.

You could argue you want to keep doc & example in the installed package, but I would then argue you should also keep tests since tests are part of the documentation as well.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,6 +2,7 @@
 * text eol=lf
 /benchmarks export-ignore
 /tests export-ignore
+/examples export-ignore
 /tools export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore


vladar:
Could you explain a bit, how examples mess things up?

theofidry:
A few plugins scans all graphql schemas and as a result includes the one from the demo, which raises errors.

Whilst it would be worth checking and fixing it in the plugin directly, this "fix" is an order of magnitude simpler (although doesn't solve the issue on the long-term).

spawnia:
This is a reasonable change. The distributed package should contain only essential files, which examples are certainly not.

enumag:
Agreed. :+1:

PR #476, opened 2019-05-10
Fix BC

mcg-web:
hi v0.13.2 introduce a BC since changing the signature or a public or a protected method is not allowed in patch releases

### DIFF
--- a/src/Validator/Rules/QueryDepth.php
+++ b/src/Validator/Rules/QueryDepth.php
@@ -99,7 +99,7 @@ class QueryDepth extends QuerySecurityRule
     /**
      * Set max query depth. If equal to 0 no check is done. Must be greater or equal to 0.
      */
-    public function setMaxQueryDepth(int $maxQueryDepth)
+    public function setMaxQueryDepth($maxQueryDepth)
     {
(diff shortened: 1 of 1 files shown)


simPod:
If only it was in docblock before :( :D

vladar:
Technically 0.x.x versions are not considered stable and [anything can change at any time](https://semver.org/spec/v2.0.0.html#spec-item-4).

But yeah, we consider our releases stable, so I've merged this. FYI, we will switch to 14.0.0 for the next release (same as the reference implementation).

mcg-web:
Thank you for the information, after more than 3 years of usage in production this is more than stable right now 😁.

serbanbjd:
This fix is breaking maxDepth check, because now in QueryDepth::isEnabled $this->getMaxQueryDepth will return false, and $this->getMaxQueryDepth() !== self::DISABLED will return true

mcg-web:
did not see [this change](https://github.com/webonyx/graphql-php/blob/0.13.x/src/Validator/Rules/QueryDepth.php#L106), this line was forced to integer in previous releases since it does not exists anymore `maxQueryDepth` must be an integer.

vladar:
@mcg-web Mind sending another PR for this?

mcg-web:
done #477 !

Issue #378, opened 2018-10-24
Having fields of interface type


mikelikenike:
Hi, am I not allowed to have fields with a type of a class extended from GraphQL\Type\Definition\InterfaceType? I tried that and ended with this error: "Schema must contain unique named types but contains multiple types named MyType" even though I checked my code and the only place where I create a new MyType is statically in my Types class.

Would the right approach be creating a class extending GraphQL\Type\Definition\UnionType and list the types(which implement my interface type) there?

Thanks

vladar:
This error means that you unintentionally create multiple instances of the same type (either somewhere in your type definitions or in your `resolveType` method). Instead of creating new type instance you must always re-use the same instance of a type (within one schema).

So double check where your types are being created (with the `new` keyword). My first guess is that you do it somewhere in the `resolveType` method of your interface type.

mikelikenike:
The only place I create a new type is in my Types class

`
public class Types {
public static function typeA(): TypeA
    {
        return self::$typea?: (self::$typea = new TypeA());
    }
public static function typeB(): TypeB
    {
        return self::$typeb?: (self::$typeb = new TypeB());
    }
public static function node(): NodeInterface
    {
        return self::$node ?: (self::$node = new NodeInterface());
    }
}
`
Both my types implement the NodeInterface.
`
class NodeInterface extends InterfaceType
{
    public function __construct()
    {
        $config = [
            'name' => 'Node',
            'fields' => [
                'title' => Type::string(),
            ],
            'resolveType' => function ($value, $context, ResolveInfo $info) {
                switch ($value->getType()) {
                    case 'typea':
                        return TypeRegistry::TypeA();
                    case 'typeb':
                        return TypeRegistry::TypeB();
                }
            },
        ];
        parent::__construct($config);
    }
}

public TypeA extends ObjectType
{
public function __construct()
  $config = [
            'name' => 'Type A',
            'interfaces' => [
                TypeRegistry::node(),
            ],
        ];
        parent::__construct($config);
}
}

public TypeB extends ObjectType
{
public function __construct()
  $config = [
            'name' => 'Type B',
            'interfaces' => [
                TypeRegistry::node(),
            ],
        ];
        parent::__construct($config);
}
}
`

In my query class, I want to do something like this:

`
$config = [
            'name' => 'Query',
                'providers' => [
                    'type' => Type::listOf(TypeRegistry::node()),
                ],
            ],
        ];
`

[...]

PR #491, opened 2019-06-07
Warn about orphaned object types

quadrifolia:
Better error when the object type is not available because it is only used as object type within an interface. The "multiple types" error on ($runtimeType !== $this->exeContext->schema->getType($runtimeType->name)) is misleading, because the type is not available in the schema at all.


mfn:
Shouldn't there be a test for this?

quadrifolia:
Probably yes. I just did a quick edit within the github website. It is just a new error message (no type detected) for [this problem](https://github.com/webonyx/graphql-php/issues/378) that took me some time because the existing error message (multiple types detected) led me into the wrong direction.

vladar:
This is useful. But this PR breaks the build (in the coding style) + we need some tests for this

spawnia:
@quadrifolia are you still interested in pursuing this PR?

PR #492, opened 2019-06-10
Add schema validation: Input Objects must not contain non-nullable circular references

spawnia:
Input Objects are allowed to reference other Input Objects. A circular reference occurs
when an Input Object references itself either directly or through subordinated Input Objects.

Circular references are generally allowed, however they may not be defined as an
unbroken chain of Non-Null fields. Such Input Objects are invalid, because there
is no way to provide a legal value for them.

The following examples are allowed:

```graphql example
input Example {
  self: Example
  value: String
}
```

This is fine because a value for `self` may simply be omitted from the arguments.

```graphql example
input Example {
  self: [Example!]!
  value: String
}
```

This also works as `self` can just contain an empty list.

The following examples are invalid:

```graphql counter-example
input Example {
  value: String
  self: Example!
}
```

```graphql counter-example
input First {
  second: Second!
  value: String
}

input Second {
  first: First!
  value: String
}
```

Spec change: https://github.com/graphql/graphql-spec/pull/445
Reference implementation: https://github.com/graphql/graphql-js/pull/1359

I think this might be added to a `0.13` release, since the relevant spec change is just a clarification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 # Changelog
+
+## Unreleased
+- Add schema validation: Input Objects must not contain non-nullable circular references (#492)
+
 ## v0.13.0
 This release brings several breaking changes. Please refer to [UPGRADE](UPGRADE.md) document for details.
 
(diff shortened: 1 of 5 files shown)


vladar:
That's great, thanks!

PR #495, opened 2019-06-12
QueryPlan can now be used on interfaces not only objects.

spyl94:
It's often the case to use interfaces in queries:

```
interface Pet { name: String! }

Query {
 pets: [Pet]
}
```


vladar:
Very good, thanks!

PR #498, opened 2019-06-13
fix query complexity validator on directives

sukano:
During query complexity validation, the directive "skip" is always check and failed when using a custom directive.

```
query {
  currencies @lang(code: "en") {
    id
    iso
    symbol
    name
  }
}
```
In my exemple, the validator searching the argument `if` who doesn't exist in the directive `lang` and return the error : `Argument "if" of required type "Boolean!" was not provided`.

### DIFF
--- a/src/Validator/Rules/QueryComplexity.php
+++ b/src/Validator/Rules/QueryComplexity.php
@@ -206,7 +206,7 @@ class QueryComplexity extends QuerySecurityRule
 
                 return ! $directiveArgsIf;
             }
-            if ($directiveNode->name->value === 'skip') {
+            if (Directive::SKIP_NAME === $directiveNode->name->value) {
                 $directive = Directive::skipDirective();
                 /** @var bool $directiveArgsIf */
                 $directiveArgsIf = Values::getArgumentValues($directive, $directiveNode, $variableValues)['if'];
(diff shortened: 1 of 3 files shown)


vladar:
Could you add a test for this case?

sukano:
done.

vladar:
@mcg-web Could you review this, please?

src/Validator/Rules/QueryComplexity.php
```diff
@@ -206,11 +206,16 @@ static function ($error) {
 
                 return ! $directiveArgsIf;
             }
-            $directive       = Directive::skipDirective();
-            $directiveArgsIf = Values::getArgumentValues($directive, $directiveNode, $variableValues);
+            if ($directiveNode->name->value === 'skip') {
```
mcg-web: can use constant here
```suggestion
            if (Directive::SKIP_NAME === $directiveNode->name->value) {
```

tests/Validator/QuerySecuritySchema.php
```diff
@@ -110,4 +119,24 @@ public static function buildDogType()
 
         return self::$dogType;
     }
+
+    public static function buildFooDirective()
```
simPod:
```suggestion
    public static function buildFooDirective() : Directive
```

PR #506, opened 2019-06-23
Spread ternary expressions across multiple lines

spawnia:
This is already the most common formatting, i just applied it to the rest of the codebase.

### DIFF
--- a/src/Validator/Rules/ProvidedRequiredArgumentsOnDirectives.php
+++ b/src/Validator/Rules/ProvidedRequiredArgumentsOnDirectives.php
@@ -66,13 +66,15 @@ class ProvidedRequiredArgumentsOnDirectives extends ValidationRule
             }
 
             $requiredArgsMap[$def->name->value] = Utils::keyMap(
-                $arguments ? array_filter($arguments, static function (Node $argument) : bool {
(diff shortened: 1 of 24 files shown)


src/Validator/Rules/ProvidedRequiredArgumentsOnDirectives.php
```diff
@@ -66,13 +66,15 @@ static function (FieldArgument $arg) : string {
...
-                        );
-                }) : [],
+                $arguments
+                    ? array_filter($arguments, static function (Node $argument) : bool {
+                        return $argument instanceof NonNullTypeNode &&
+                            (
+                                ! isset($argument->defaultValue) ||
+                                $argument->defaultValue === null
+                            );
+                    })
```
mfn: Are you sure this is supposed to be on the same level as the `?` and `:` and not 3 spaces to the right so it's below the **a**`rray_filter`?

I might have seen a few other places too but before I nitpick on all of them… ;)

spawnia: This is the PHPStorm default style, and it is applied consistently.

vladar:
Thanks!

PR #508, opened 2019-06-23
Nonfunctional changes to improve consistency and clarity

spawnia:


tests/Executor/DeferredFieldsTest.php
```diff
@@ -401,15 +401,15 @@ public function testComplexRecursiveDeferredFields() : void
                 return [
                     'sync'         => [
                         'type'    => Type::string(),
-                        'resolve' => function ($v, $a, $c, ResolveInfo $info) {
+                        'resolve' => function ($val, $args, $context, ResolveInfo $info) {
```
simPod: I think we can use `value` and `arguments` if we are doing this already :)

src/Executor/ReferenceExecutor.php
```diff
@@ -927,28 +928,27 @@ function ($previous, $value) use ($callback) {
...
-     * Complete a list value by completing each item in the list with the
-     * inner type
+     * Complete a list value by completing each item in the list with the inner type.
      *
-     * @param FieldNode[] $fieldNodes
-     * @param mixed[]     $path
-     * @param mixed       $result
+     * @param FieldNode[]         $fieldNodes
+     * @param mixed[]             $path
+     * @param mixed[]|Traversable &$results
```
simPod: is specifying `&` in phpdocs standard thing? I'm not against just don't remember seeing it anywhere else 🤔

spawnia: It is not in the standard. Removed

tests/Executor/DeferredFieldsTest.php
spawnia: `args` is used anywhere, and `$val` is going to be `$rootValue`

PR #509, opened 2019-06-23
PHPStan Level 2

spawnia:
I added in the simple fixes and added an exception/TODO for the strict rule that forbids loose boolean comparisons.

## Discussion

A whole class of errors in PHPStan is a result of PHP's lack of union types.
This commonly happens in the parts of the code that deal with the GraphQL
type system where we can currently use interfaces and lose type safety.

There are about ~30 individual errors caused by that. Given that we want to keep leveling up PHPStan to reap the benefits of the higher static analysis levels, there are a few ways of going forward.

The simples workaround would be to add the individual errors we know to be false positives resulting from the lacking union types to the PHPStan config `ignoreErrors` section in a dedicated block.

Another workaround which would maybe be a bit cleaner would be to add comments to tell PHPStan to ignore the following line, this is not yet possible but we might make it so https://github.com/phpstan/phpstan/issues/786

Finally, we could think about ways to actually cleanly type this thing. This might require breaking changes of this library.


src/Language/AST/Node.php
```diff
@@ -39,6 +39,9 @@ abstract class Node
     /** @var Location */
     public $loc;
 
+    /** @var string */
+    public $kind;
```
mfn: This will be very useful for other libraries using graphql-php, see https://github.com/rebing/graphql-laravel/blob/a458a79f7bad6406f181ac4d1ab371d65aa12c7f/phpstan.neon#L37

phpstan.neon.dist
```diff
@@ -10,6 +10,13 @@ parameters:
         - "~(Method|Property) .+::.+(\\(\\))? (has parameter \\$\\w+ with no|has no return|has no) typehint specified~"
         - "~Variable property access on .+~"
         - "~Variable method call on static\\(GraphQL\\\\Server\\\\ServerConfig\\)~" # TODO get rid of
+        - "~Only booleans are allowed in .*~" # TODO https://github.com/phpstan/phpstan-strict-rules/issues/2
```
simPod: I'm working on this in small PRs just FYI (eg. https://github.com/webonyx/graphql-php/pull/503/files)

tests/Language/ParserTest.php
```diff
@@ -333,7 +333,7 @@ public function testParseCreatesAst() : void
 ');
         $result = Parser::parse($source);
 
-        $loc = static function ($start, $end) {
+        $loc = static function (int $start, int $end) {
```
simPod: we can also add return type here

tests/Type/DefinitionTest.php
```diff
@@ -702,19 +706,24 @@ public function testAllowsShorthandFieldDefinition() : void
...
-        $withArg = $schema->getType('SomeInterface')->getField('withArg');
+        $withArg = $SomeInterface->getField('withArg');
         self::assertEquals(Type::string(), $withArg->getType());
 
         self::assertEquals('arg1', $withArg->args[0]->name);
         self::assertEquals(Type::int(), $withArg->args[0]->getType());
 
-        $testField = $schema->getType('Query')->getField('test');
+        /** @var ObjectType $Query */
+        $Query = $schema->getType('Query');
```
simPod: small `Q`

spawnia: I think that in this case, having the variable name match the type name is fine. This is done in other parts of the lib as well.

simPod: If it's in other parts of the lib, it should be fixed as well. 

camelCase for variable names is reasonably well-established in PHP world and others. There's not reason to break that. The only case I can think of is [naming tests using non-breaking spaces](https://mnapoli.fr/using-non-breakable-spaces-in-test-method-names/) / underscores for better in-tests-documentation readability.

spawnia: While i generally agree, in this case i like matching names better. Standards are rules are fine, but should not be applied dogmatically.

simPod: The point is that if everyone who comes, makes PR with something he likes it will be a mess soon,I don't think there's any dogma at all. 

But makes me wonder why there's no CS rule for this. Will investigate.

spawnia: @simPod right. And right now, variables that contain types are named just like i did here throughout the test suite. Consistency is key.

vladar:
@spawnia I think I am ready to merge this, but can you resolve conflicts, please?

spawnia:
I actually went ahead and "expanded" the union types into the PHPDocs. This works out to satisfy PHPStan.

While a bit tedious, i don't see the GraphQL type system changing much in the near future, so in terms of maintenance, this seems fine.

vladar:
Merging this broke the build (I assume after #501). Can you take a look?

simPod:
@vladar #526

vladar:
Damn, you're fast %) Thanks!

spawnia:
Thanks guys :)

PR #511, opened 2019-06-23
Resolve todo in Boolean coercion, add explanation, update test names to match reference implementation

spawnia:


tests/Type/ScalarSerializationTest.php
```diff
@@ -13,9 +13,9 @@ class ScalarSerializationTest extends TestCase
 {
     // Type System: Scalar coercion
     /**
-     * @see it('serializes output int')
+     * @see it('serializes output as Int')
```
vladar: I suggest not to change @it annotations because they are basically IDs for the reference implementation test cases. You can copy contents of this annotation and quickly find a related test in the reference implementation. That's the only reason for their existence.

vladar: Oh, ignore %) I see that they've changed those in the reference implementation

PR #517, opened 2019-07-10
Upgrade PHPStan to 0.11.12

simPod:


vladar:
@simPod Looks like it broke the build. Can you take a look?

simPod:
Yes, will open PR in about 8h

PR #519, opened 2019-07-12
Remove composer.lock from .gitignore

yura3d:
I think we need to remove `composer.lock` from `.gitignore` because it makes reproducible builds impossible. Even [composer documentation](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control) states it.

I created this PR because today I fall in indicative situation. I cloned my fork of this repo on local machine, typed `composer install` and got the build that can't pass the tests. But I know that this build passed the tests a week before. The reason is phpstan library, that was updated from 0.11.8 to 0.11.12, and this update introduced breaking change.
We had (a week ago) ^0.11.8 version constraint for phpstan in `composer.json` and it allows any available version of 0.11.x (where x >= 8). If we would have `composer.lock` in repository, I would get the right version of phpstan, which was used to prepare the build. But without it I got the last available version of phpstan, which meets the version constraint, but wasn't tested with the build and broke it.

This PR also includes #518 to pass the tests (current master branch is broken).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .phpcs-cache
-composer.lock
 composer.phar
 phpcs.xml
 phpstan.neon
--- a/src/Type/Definition/Directive.php
+++ b/src/Type/Definition/Directive.php
@@ -20,8 +20,8 @@ class Directive
     public const DEPRECATED_NAME      = 'deprecated';
     public const REASON_ARGUMENT_NAME = 'reason';
 
-    /** @var Directive[] */
-    public static $internalDirectives = [];
+    /** @var Directive[]|null */
+    public static $internalDirectives;
 
     // Schema Definitions
 


simPod:
The composer documentation states, that you should commit your `composer.lock` because `... who sets up the **project to use the exact same versions of the dependencies**...`. That is not the case with libraries. They should support all versions between constraints defined within `composer.json`.

Is the problem that higher PHPStan version found more issues than the previous one? That can be easily solved by locking the version in `composer.json` but I have not found it worth it anyway. The real problem here is that in X.Y.Z PHPStan versioning the `Z` is not considered patch version but minor version by the author...

yura3d:
> They should support all versions between constraints defined within `composer.json`.

It sounds greatly, but nobody is insured of errors in new versions of dependencies.

I think any tested build (no matter what it is: a project or library) should install and work correctly without time dependency. How software can be titled stable, when today it builds, but tomorrow will not, and nothing was changed in it?
`composer install` should always get repoducible tested working build.
Want to play with new versions of dependencies? Go ahead with `composer update`.

simPod:
That's how semver works. You have to deal with it and backward compatibility is each maintainer's responsibility.

simPod:
All I'm saying is that I'd rather fix those dev dependencies causing problems than locking all dependencies.

mfn:
I'm with @simPod ; `composer.lock` should not be part of the library; same reasons.

yura3d:
Thanks for your replies. May be I don't understand something.

[doctrine/orm](https://github.com/doctrine/orm), for example, and other doctrine projects like [doctrine/common](https://github.com/doctrine/common), [doctrine/dbal](https://github.com/doctrine/dbal) are also libraries and have a lock in repos. I think due to this, on my working with doctrine for about 5 years I have never got the situation when the builds fails or the tests is broken.

Can you please explain the differences between `orm` and `graphql-php` packages, because of which the first should have a lock, and the second shouldn't? Thanks.

simPod:
The maintainers of those mentioned packages decided to have that for stability of testing packages. It's pretty similar to locking the version tho.

vladar:
This PR also contains some unrelated change to directives?

vladar:
I do understand the intent of this change (to get reproducible builds for tests), but the issue is that our `composer.lock` will be [still ignored](https://getcomposer.org/doc/02-libraries.md#lock-file) by projects using this library.

In other words - we'll get stable builds for tests, but users of the lib won't get the same builds anyway. They will install latest deps as defined in `composer.json`.

So it is actually better for us to follow the same logic by installing latest dependencies so that we could see possible issues earlier.

In order to freeze requirements we should better change versioning in the `composer.json` from `~` to explicit versions when needed.

Issue #309, opened 2018-07-16
throwing GraphQL\Error\UserError: HTTP 400


baversjo:
Hello,

In a previous version from april ( tree https://github.com/webonyx/graphql-php/commit/ec0985619f ),
throwing a `GraphQL\Error\UserError` would return the error in the http response per graphql standard, with HTTP 200.

However, when I upgraded to a recent version of dev-master (https://github.com/webonyx/graphql-php/commit/82b1dbd836 ) throwing a UserError would cause http 400 bad request instead (still with the error in the body).

My understanding is, that per graphql standards, user errors should still return http 200 but with the error in the http body? Is this a bug, or was this change intended behavior in `graphql-php`? I looked through the changes and while there's been a lot, there's nothing obvious that would indicate this change: https://github.com/webonyx/graphql-php/compare/ec0985619f...82b1dbd836

vladar:
This is likely a bug. Can you provide a smallest possible test case on how to reproduce it?

jsiebern:
I can confirm this problem. Reproduction is fairly easy, just use **any** schema / resolver schema (at least for me) and throw an error like so:
```php
throw new \GraphQL\Error\UserErrorUserError('Any error');
```

This will return an http response with the code 400 instead of 200 which makes Apollo (as an example) not pick it up properly!

jsiebern:
The issue seems to be here:

https://github.com/webonyx/graphql-php/blob/392b567f235a528539f47e8c501f2daed029adeb/src/Server/Helper.php#L480

Where it probably should be `empty` instead of `!empty` ?

vladar:
Initially, this server was supposed to mirror express-graphql logic. And they return status code 500 in such case ([here](https://github.com/graphql/express-graphql/blob/master/src/index.js#L330)).

So status 400 is obviously invalid. But the question is if we should switch to Apollo-style and return 200 or be compatible with express-graphql and return 500 (which is also more semantically correct for HTTP in case of execution errors).

jsiebern:
I'd say the context of HTTP protocol it makes sense to send status 4 / 500. But in a GraphQL Context it makes sense to send 200 as the call itself has passed and the graphql implementation expects the client to handle those errors, there's no need to indicate it with another status (which apollo makes quite clear by failing on a not 200 code and not displaying the actual errors).

[...]

PR #524, opened 2019-07-25
Always return http 200

baversjo:
Fixes: https://github.com/webonyx/graphql-php/issues/309

### DIFF
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -484,7 +484,6 @@ class Helper
                     }
                 }
             );
-            $httpStatus = 200;
         } else {
             if (! $result instanceof ExecutionResult) {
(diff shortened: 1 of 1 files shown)


mfn:
If this method only returns `200`, why have it at all?

😉

baversjo:
@mfn it had some error handling in it that I presumed should be kept. To be honest though this PR will probably be rejected as there's might be a bigger plan. However, I had to create a temporary solution as apollo client is unusable without this fix, you cannot handle users gracefully when 400 is returned.

I published the PR as others may find this useful trying to use apollo with this lib.

vladar:
I am going to close this in favor of https://github.com/APIs-guru/graphql-over-http

Don't think it makes sense to introduce a break until the new HTTP spec is ready.

PR #528, opened 2019-08-06
Support no field selection

stevelacey:
I am getting this error on my query that accepts no args and returns a string:

```
Argument 1 passed to GraphQL\\Type\\Definition\\ResolveInfo::foldSelectionSet() must be an instance of GraphQL\\Language\\AST\\SelectionSetNode, null given
```

Here's the query:

```
query Nonce {
  nonce: Nonce
}
```

This should work? Right? This change fixes it.

### DIFF
--- a/src/Type/Definition/ResolveInfo.php
+++ b/src/Type/Definition/ResolveInfo.php
@@ -178,10 +178,12 @@ class ResolveInfo
 
         /** @var FieldNode $fieldNode */
         foreach ($this->fieldNodes as $fieldNode) {
-            $fields = array_merge_recursive(
-                $fields,
-                $this->foldSelectionSet($fieldNode->selectionSet, $depth)
-            );
+            if ($fieldNode->selectionSet) {
+                $fields = array_merge_recursive(
+                    $fields,
+                    $this->foldSelectionSet($fieldNode->selectionSet, $depth)
+                );
+            }
         }
 
         return $fields;


simPod:
For me this already works. Do you have a test?

stevelacey:
@simPod oh really? I'll write one. Where would that test go?

stevelacey:
You're right, this does work, at least according to the `droidName` test I managed to cobble together... I wonder if it's something about how `rebing/graphql-laravel` is setting up the type.

stevelacey:
Ah it's because I'm calling `$info->getFieldSelection()` manually, should something be checked before calling that? Should it not be safe to call that method in this scenario?

mfn:
Can you be more specific about your scenario? You revealed you're using rebing, can you share more code what you're actually doing? And what version of rebing are you using?

simPod:
AFAIK it does not modify any state so should be safe. But it might be that there is no field selection so null is passed and it is not handled. Therefore this PR is IMO actually viable. Let's see what @vladar thinks. 

But test would be nice, maybe ResolveInfoTest? Not sure.

src/Type/Definition/ResolveInfo.php
```diff
@@ -178,10 +178,12 @@ public function getFieldSelection($depth = 0)
 
         /** @var FieldNode $fieldNode */
         foreach ($this->fieldNodes as $fieldNode) {
-            $fields = array_merge_recursive(
-                $fields,
-                $this->foldSelectionSet($fieldNode->selectionSet, $depth)
-            );
+            if ($fieldNode->selectionSet) {
```
simPod:
```suggestion
            if ($fieldNode->selectionSet === null) {
                continue;
            }

            $fields = array_merge_recursive(
                $fields,
                $this->foldSelectionSet($fieldNode->selectionSet, $depth)
            );
```

stevelacey:
I am basically wrapped `getResolver()` so that I could log args and return fields across the entire API, kind of like a middleware, but with access to the parsed AST.

Amongst other things I call `$info->getFieldSelection()` there, on every query and mutation, regardless of return type, which is why it's happened nonsensically on a scalar.

I just got my tests passing with a simple return type check like this:

```
    private function fields(ResolveInfo $info) : array
    {
        if (!$info->returnType instanceof ObjectType) {
            return [];
        }

        return array_keys(array_dot($info->getFieldSelection($depth = PHP_INT_MAX)));
    }
```

but I would imagine it's valuable to make `$info->getFieldSelection()` never fail – I would expect an empty array in this scenario.

PR #533, opened 2019-08-11
Fixing arguments order when getting value from AST

ArnoudThibaut:
Hello,
Given the following `InputObjectType` :
```php
new InputObjectType([
    'name'   => 'order',
    'fields' => [
        'foo'  => ['type' => Type::string()],
        'bar'  => ['type' => Type::string()]
    ],
]);
```
and something like `order: {bar: "ASC", foo: "ASC"}` as value, the return of `GraphQL\Utils\AST::valueFromAST` will be `['foo' => 'ASC', 'bar' = 'ASC']`.

However when the order of arguments has importance, like constructing an `ORDER` SQL clause, this can pose a problem and the return should be `['bar' => 'ASC', 'foo' = 'ASC']`.

The aim of this PR is to fix this by using the order given by the user instead of the order defining during the field configuration.

I used directly `assertSame()` in the test case instead of the helper method `runTestCaseWithVars` because this method use `assertEquals` and this doesn't test the order of the keys.

Tell me if I should replace `assertEquals` by `assertSame` in the helper method, I was not sure about that.

fix : https://github.com/api-platform/core/issues/2808

### DIFF
--- a/src/Utils/AST.php
+++ b/src/Utils/AST.php
@@ -42,8 +42,11 @@ use stdClass;
 use Throwable;
 use Traversable;
 use function array_combine;
+use function array_flip;
 use function array_key_exists;
+use function array_keys;
 use function array_map;
+use function array_replace;
 use function count;
 use function floatval;
 use function intval;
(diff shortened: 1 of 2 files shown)


vladar:
This is not a bug. The specification [clearly states this](https://graphql.github.io/graphql-spec/draft/#sel-EAFTREBAAAFJBPABjzZ):

> **Arguments are unordered**
> 
> Arguments may be provided in any syntactic order and maintain identical semantic meaning.
> 
> These two queries are semantically identical:
> 
> Example № 12
> ```
> {
>   picture(width: 200, height: 100)
> }
> ```
> 
> Example № 13
> ```
> {
>   picture(height: 100, width: 200)
> }
> ```

While I understand what you are trying to achieve, relying on arguments order is not spec-compliant. If we merge this it will make client apps fragile because they will rely on the implementation-specific feature.

See also recent [discussion](https://github.com/graphql/graphql-spec/issues/577) about this topic exactly in the graphql-spec repo. 

So for your use-case of `order` clause the correct model would be a single argument of `List` type. Lists do preserve ordering of elements.

PR #535, opened 2019-08-13
Allow \stdClass for input types

### LINKED ISSUE #534: support \stdClass for input types
Hi, thanks for a great project - it's been very useful.

I've got a custom scalar<sup>1</sup> in my app which represent JSON values that validate against a given JSON Schema. In my scalar's parseValue method I validate the value against my Schema using [opis/json-schema](https://github.com/opis/json-schema). It expects JSON arrays to be arrays and JSON objects to be \stdClass instances.

I've tried decoding my variables JSON with \stdClasses (and casting the root variables object to an associative array - leaving only input types and scalars as \stdClasses). This works great for my custom scalar, but it now means all my inputs are \stdClasses - which I'm ok with, but this library isn't.

[Value::coerceValue](https://github.com/webonyx/graphql-php/blob/v0.13.6/src/Utils/Value.php#L160-L217) expects input types to be associative arrays or Traversables. It returns the content as an array. **Would you consider casting \stdClasses to arrays in this method, if they are passed in?** That way I could safely pass a \stdClass in. The method could then return the array version as it currently does, or cast back to \stdClass before returning (only if it received a \stdClass in the first place) - I'm not worried about this detail, my resolvers could be written to deal with input types in either format.

I'd be happy to do a pull request if you think this is a suitable feature / solution. If you have any other suggestions for how I can get my custom scalars decoded as \stdClasses without touching the inputs, please let me know.

<sup>1</sup>I have multiple custom JSON scalars, each with their own JSON schemas. I find these really useful for things like user preferences - key value pairs where the values are all different types. I just store that JSON in a column in the DB and make sure that it matches the JSON Schema whenever I update it (through GraphQL)

adam1658:
Allow `\stdClass` in `GraphQL\Util\Value::coerceValue`. `\stdClass` will be cast to an associative array, and the coerced value will still be an associative array.

Useful for when you want to parse incoming variables JSON with `\stdClass` (for example, if you have custom scalars that have this requirement, but cannot distinguish between scalars and inputs at the variables level).

Note that the variables map is still required to be an associative array - one way that users can parse variables JSON in this way is to parse in `\stdClass` mode, then cast the map to an array:
```php
$variables = (array)json_decode('{"a": {}, "b": []}', false);
var_dump($variables);
/* array(2) {
  ["a"]=>
  object(stdClass)#1 (0) {
  }
  ["b"]=>
  array(0) {
  }
}
*/
```
Fixes #534.

Please let me know if the implementation, tests or docs changes are incorrect or lacking.

Thanks,
Adam

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 - Add schema validation: Input Objects must not contain non-nullable circular references (#492)
 - Added retrieving query complexity once query has been completed (#316) 
-- Allow input types to be passed in from variables using \stdClass instead of associative arrays
+- Allow input types to be passed in from variables using \stdClass instead of associative arrays (#535)
 
 #### v0.13.5
 - Fix coroutine executor when using with promise (#486) 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 - Add schema validation: Input Objects must not contain non-nullable circular references (#492)
 - Added retrieving query complexity once query has been completed (#316) 
+- Allow input types to be passed in from variables using \stdClass instead of associative arrays
 
 #### v0.13.5
 - Fix coroutine executor when using with promise (#486) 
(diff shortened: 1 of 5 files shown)


vladar:
Cool! Thanks for such a complete PR with docs, tests, etc. It's a rare thing %)

adam1658:
Hi @vladar, when do you expect the next release to be published? I'd love to start making use of this feature.

vladar:
@adam1658 Just published v0.13.7. Check it out

adam1658:
Legend! Thanks so much

PR #539, opened 2019-09-19
Implement BuildClientSchema

### LINKED ISSUE #293: Can I build shema object from introspection string?
I founded that i can build schema object from sdl, but can I build shema object from introspection string?

spawnia:
Resolves #293

Equivalent to https://github.com/graphql/graphql-js/blob/master/src/utilities/buildClientSchema.js

@vladar is there something you wish to change about the implementation? I would do the tests next.


simPod:
@spawnia out of curiosity, why did you decide to go with arrays (https://github.com/webonyx/graphql-php/pull/539/commits/0f7d970fb463598aa718dab76895c640a47a42ca)?

spawnia:
@simPod it fit better with the rest of the library, which also uses associative arrays for other parts that interface with client code.

I prefer `stdClass`, as it is a clean distinction to sequential arrays, but still think that internal consistency of the project is more important than my personal preferences.

vladar:
Cool, let me know when you finish.

spawnia:
@vladar finished the implementation and test suite.

A few minor fixes in the rest of the library came up while implementing the whole test suite. No changes to other tests were required and i don't think there are breaking changes.

I stuck as close as possible to the reference implementation. There were 2 tests i had to skip, related to https://github.com/webonyx/graphql-php/pull/567

src/Type/Definition/FieldArgument.php
```diff
@@ -87,7 +95,7 @@ public function getType() : Type
 
     public function defaultValueExists() : bool
     {
-        return array_key_exists('defaultValue', $this->config);
+        return $this->defaultValueExists;
```
vladar: Looks like this reverts some changes made recently (see https://github.com/webonyx/graphql-php/pull/566#discussion_r339321880). Is it intentional?

vladar:
Awesome, I'll do a full review as soon as I have a chance! Quick note - looks like style checks are failing

src/Type/Definition/FieldArgument.php
spawnia: I did it to align the implementation with `InputObjectField`.

Let's make all the same: https://github.com/webonyx/graphql-php/pull/585

spawnia:
@vladar everything is passing now apart from Scrutinizer complaining about the complexity of `BuildClientSchema`.

I think we should stick close to the reference implementation and keep the class as is. The error should go away once this is merged, right?

vladar:
@spawnia Ignore scrutinizer, we'll remove it from our checks and just use PHPStan. I am ready to merge this but can you please resolve conflicts first?

simPod:
@vladar OT: I marked only Travis as a Required check few weeks ago.

spawnia:
Finally, the pipeline succeeded. @vladar this should be good to go. I have been using this branch successfully in https://github.com/spawnia/sailor

src/Type/Introspection.php
```diff
@@ -184,6 +185,34 @@ public static function getTypes()
...
+     * of the server context, for instance when doing schema comparisons.
+     *
+     * Options:
+     *   - descriptions
+     *     Whether to include descriptions in the introspection result.
+     *     Default: true
+     *
+     * @param array<string, bool> $options
+     *
+     * @return array<string, mixed[]>|null
```
simPod: I'd unify the style to document array (eg. PHPStorm complains when mixing them)

```suggestion
     * @return array<string, array<mixed>>|null
```

src/Type/TypeKind.php
```diff
@@ -6,12 +6,12 @@
...
-    const LIST_KIND      = 6;
-    const NON_NULL       = 7;
+    const SCALAR       = 'SCALAR';
+    const OBJECT       = 'OBJECT';
+    const INTERFACE    = 'INTERFACE';
+    const UNION        = 'UNION';
+    const ENUM         = 'ENUM';
+    const INPUT_OBJECT = 'INPUT_OBJECT';
+    const LIST         = 'LIST';
+    const NON_NULL     = 'NON_NULL';
```
simPod: I wonder what is the reason for this change?

src/Utils/Utils.php
```diff
@@ -272,9 +272,9 @@ public static function groupBy($traversable, callable $keyFn)
     }
 
     /**
-     * @param mixed[]|Traversable $traversable
+     * @param array<mixed, mixed>|Traversable $traversable
```
simPod: Array cannot have `mixed` in keys. Also we can add documentation for Traversable type or change it to `iterable` because `array|Traversable = iterable`

src/Utils/Utils.php
```diff
@@ -272,9 +272,9 @@ public static function groupBy($traversable, callable $keyFn)
...
      *
-     * @return mixed[][]
+     * @return array<mixed, mixed>
```
simPod: Same about key here.

tests/Utils/BuildClientSchemaTest.php
```diff
@@ -0,0 +1,1082 @@
...
+use GraphQL\Type\Introspection;
+use GraphQL\Type\Schema;
+use GraphQL\Utils\BuildClientSchema;
+use GraphQL\Utils\BuildSchema;
+use GraphQL\Utils\SchemaPrinter;
+use PHPUnit\Framework\TestCase;
+use function array_filter;
+
+/**
+ * @see BuildClientSchema
```
simPod: I'd change it to

```suggestion
 * @covers GraphQL\Utils\BuildSchema
```

It will prevent coverage leaking and have same documentation effect.

tests/Utils/BuildClientSchemaTest.php
```diff
@@ -0,0 +1,1082 @@
...
+ */
+class BuildClientSchemaTest extends TestCase
```
simPod: final, should not be allowed to be extended

tests/Utils/BuildClientSchemaTest.php
```diff
@@ -0,0 +1,1082 @@
...
+        $serverSchema         = BuildSchema::build($sdl);
+        $initialIntrospection = Introspection::fromSchema($serverSchema);
+        $clientSchema         = BuildClientSchema::build($initialIntrospection);
+        $secondIntrospection  = Introspection::fromSchema($clientSchema);
+
+        self::assertSame($initialIntrospection, $secondIntrospection);
+    }
+
+    /**
+     * @return array<string, mixed[]>
```
simPod:
```suggestion
     * @return array<string, array<mixed>>
```

src/Type/TypeKind.php
spawnia: This allows reusing the constant values, they are needed in a few places when dealing with introspection.

Also, naming a constant `LIST` is possible in PHP 7+ only, i think. Since we dropped compatibility, there is no reason to keep the inconsistent name.

src/Utils/Utils.php
spawnia: What would be the type for the key then?

simPod: For arrays it can only be `int|string`

tests/Utils/BuildClientSchemaTest.php
spawnia: I am not convinced that using `@covers` is necessary. There seems to be very little gain for a bunch of work.

tests/Utils/BuildClientSchemaTest.php
spawnia: I am firmly opposed to the usage of the `final` keyword in libraries such as this.

src/Type/TypeKind.php
simPod: Oh, were constant names also reserved? :D mkay. I wondered the same about `INTERFACE` for a while.

src/Utils/Utils.php
spawnia: What about `float` or `bool`?

tests/Utils/BuildClientSchemaTest.php
simPod: What do you mean by bunch of work? It's basically just change of one line, isn't it? Can you explain?

I discovered this recently and I like how the coverage does not leak. When I generate line coverage it show much better what I covered in tests and what just "accidentally" runs. But whether something runs or not I know from PHPStan already so personally I prefer not to have unit B covered by testing unit A.

tests/Utils/BuildClientSchemaTest.php
simPod: I understand there are people avoiding `final` but for tests there's no issue, hm? Consumers won't get affected anyway and we accidentally won't extend Tests nor they will get suggested in autocomplete.

src/Utils/Utils.php
simPod: arrays can't have bool or float in keys https://3v4l.org/rQen6 It will always be subset of `int|string`

tests/Utils/BuildClientSchemaTest.php
spawnia: I don't see the problem with "leaking coverage".

src/Utils/Utils.php
```diff
@@ -272,9 +272,9 @@ public static function groupBy($traversable, callable $keyFn)
     }
 
     /**
-     * @param mixed[]|Traversable $traversable
+     * @param iterable $traversable
```
spawnia: Fixed all related type hints in https://github.com/webonyx/graphql-php/pull/613

tests/Utils/BuildClientSchemaTest.php
simPod: The coverage is partially fabricated so the metric is less accurate and makes detection of untested paths harder.

spawnia: The code paths are being tested though, albeit indirectly.

src/Type/Definition/FieldArgument.php
```diff
@@ -32,6 +32,13 @@ class FieldArgument
     /** @var InputType&Type */
     private $type;
 
+    /**
+     * Helps to differentiate when `defaultValue` is `null` and when it was not even set initially
+     *
+     * @var bool
+     */
+    private $defaultValueExists = false;
```
shmax: Why is this making a return?

spawnia: This change is already reverted.

The implementation for those kinds of checks was unified in https://github.com/webonyx/graphql-php/pull/585.

tests/Utils/BuildClientSchemaTest.php
simPod: IMO there's no indirect testing, either you test it or not. Running the code is not testing.  
Whether coverage leaking is issue depends on whether classes it leaks to are part of the unit or not. But those classes used in BuildClientSchema are not part of the unit so coverage should not be leaking to them.

This is just the nitpicking though. I saw the `@see` annotation so having `@covers` there would make more sense to me.  AFAIK we do not use this annotation here yet.

vladar:
Awesome! Thank you for this huge work. And sorry for being late to merge this - last two months were pretty intense for me.

Issue #544, opened 2019-09-26
QueryPlan Field "__typename" is not defined for type ...


oesteve:
QueryPlan throw "Field "__typename" is not defined for type ..." when query has __typename field.

Thanks!

oesteve:
Maybe a solution too simple, I don't know if it's enough https://github.com/webonyx/graphql-php/pull/545

vladar:
Hey, thanks for noticing! This is definitely a bug, but the question is if we should really exclude __typename or just handle it properly and include in the output.

I guess the expectation of most users would be not to have it included, but curious to hear other opinions first (if any). So I will keep your PR open for a while and if we don't get any objections - will merge it.

keulinho:
I think it would be cleaner to include it also to the QueryPlan, as i find it weird to just exclude this field.
But i don't have a special use case that it must be included, IMHO it would just be cleaner if it is included.
An argument for including it is that it is possible to add your own `__typename` with its own resolver and so own, for the sake of consistency i would argue that if the library allows you to add it like a normal field it should be added to the query plan.

vladar:
> An argument for including it is that it is possible to add your own __typename with its own resolver and so own

This is actually not the case. Fields with `__` prefix are fields reserved for introspection, and you can't redefine them (see [in the code](https://github.com/webonyx/graphql-php/blob/master/src/Utils/Utils.php#L534-L554)).

Also, the introspection is not a part of a general data resolution process - it is a separate mechanism that is available in the regular query for convenience.

oesteve:
Hey! I don't have a position for this case ... but if __typename is ignored in queryplan, what do we expect from the getFieldSelection method? will be ignored? will it return true like other types?

mikespub:
Are there any plans to fix this in the future? I noticed several PRs on this, but none of them made it into the repo, and the problem still exists...

Thanks

spawnia:
Fixed as of https://github.com/webonyx/graphql-php/releases/tag/v14.6.4

PR #545, opened 2019-09-27
Fix __typename field in QueryPlan

oesteve:
If query contains__typename field, QueryPlan throw 'Field "__typename" is not defined for type ... '

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fixed error with __typename fields in QueryPlan (#544)
 - **BREAKING/BUGFIX:** Strict coercion of scalar types (#278)
 - **BREAKING:** Removed deprecated directive introspection fields (onOperation, onFragment, onField)
 - **BREAKING:** Removal of `VariablesDefaultValueAllowed` validation rule. All variables may now specify a default value.
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,8 @@
 ## Master
 
+### Breaking: QueryPlan with introspection fields
+Now when in __typename field is present, QueryPlan return the field definition.
+
 ### Breaking (major): dropped deprecations
  - dropped deprecated `GraphQL\Schema`. Use `GraphQL\Type\Schema`.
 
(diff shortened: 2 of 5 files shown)


vladar:
Thanks for the PR. I guess we also need some test for this in [Resolveinfo tests](https://github.com/webonyx/graphql-php/blob/master/tests/Type/ResolveInfoTest.php).

oesteve:
Writing the test It has emerged me a doubt about the behavior of getFieldSelection method https://github.com/webonyx/graphql-php/issues/544#issuecomment-544267254

vladar:
I guess both `getFieldSelection` and `QueryPlan` should behave the same. Technically that would be a breaking change so we also need a note in the CHANGELOG.md about it

One thing I have in mind though is that QueryPlan is supposed to support many use-cases. So ideally there should be an option (e.g `withTypename`) which would allow users to get a full set of fields.

What do you think?

oesteve:
I agree that both methods would have to maintain the same behavior, so ...  backward compatibility may be a reason not to ignore the field in QueryPlan.
On whether it has to be parametizable, I think it may be good not to add more complexity, but it doesn't seem bad. If it seems good to you, I can add them.

src/Type/Definition/ObjectType.php
```diff
@@ -117,6 +118,11 @@ public static function assertObjectType($type)
      */
     public function getField($name)
     {
+        $typeNameDef = Introspection::typeNameMetaFieldDef();
+        if ($name === $typeNameDef->name) {
+            return $typeNameDef;
+        }
+
```
vladar: Somehow I managed to miss that you added changes to `ObjectType`. This is not quite correct because introspection is a completely separate process. `ObjectType` shouldn't know anything about it. Instead, we should literally check for `__typename` field in the `QueryPlan` and `getFieldSelection` method.

oesteve: Ok, no problem! I have updated the QueryPlan and the getFieldSelection method already provides the expected behavior.
I don't know the code in depth, maybe I've been cheeky to write a PR, tell me everything you don't see correctly or feel free to rewrite it.

spawnia:
@oesteve are you still interested in pursuing this PR?

oesteve:
finally I gave another approach to the implementation, for my part it can
be discarded

Thanks for all!

PR #546, opened 2019-09-30
Add test for input object not required field with default value

alquerci:
From the GraphQL specification of June 2018.

> An input field is required if it has a non‐null type and does not have
> a default value. Otherwise, the input object field is optional.

Note
----

I see that this bug has been fixed on the `master` branch as breaking change but not yet released.
It will be nice to backport this patch on the branch `0.13.x` with a BC layer.

I have not time to dig deeper in order to implements the patch but I provide at least a failed test for that.

### DIFF
--- a/tests/Regression/SpecJuin2018/InputObjectRequiredFieldsTest.php
+++ b/tests/Regression/SpecJuin2018/InputObjectRequiredFieldsTest.php
@@ -1,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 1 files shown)


vladar:
Thanks for the test! But 0.13.x must stay backwards-compatible. Version 14.0 is on its way and will be released in the next couple of weeks (hopefully), so the solution would be to upgrade when its out.

alquerci:
> Version 14.0 [...] will be released in the next couple of weeks (hopefully).

Thank you and good luck for this release.

PR #550, opened 2019-10-03
Remove dead code

enumag:
Since PHP5 support is no longer supported we can safely remove this.

Btw. I also noticed this little [todo](https://github.com/webonyx/graphql-php/blob/cd31b00dd33d73a1b5577e2c7204a3233b3d355d/src/Executor/Promise/Adapter/ReactPromiseAdapter.php#L80) but I don't have enough experience with React to implement it. How do you even use Generators with React? It doesn't seem to have coroutines like Amphp.

### DIFF
--- a/src/Deferred.php
+++ b/src/Deferred.php
@@ -56,8 +56,6 @@ class Deferred
         try {
             $cb = $this->callback;
             $this->promise->resolve($cb());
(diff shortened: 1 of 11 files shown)


vladar:
Thanks!

PR #551, opened 2019-10-03
Add AmpPromiseAdapter

enumag:
Had some time today so I implemented a PromiseAdapter for [Amp](https://amphp.org/). Note that I just wrote blindly what seemed to make some sense and it's not tested at all. I'll try to use it in my project when I get to reimplementing my GraphQL part to async. And of course I need to add some unit tests. All of this might take a while (a few months), I'm just posting this here in case someone was looking for it and was willing to help me out with testing.

Code reviews are of course welcome.


enumag:
Added unit tests. I didn't test it in my project yet and won't be able to for a while. It should work though. Whether you merge it now or wait until I give a green from my app is up to you.

vladar:
Let's wait until you confirm it works as expected as I don't have any AMP project to check it.

enumag:
After some fixes it works with the ReferenceExecutor.

When I try to use the ExperimentalExecutor, I'm getting an extremely weird error:

```
Error: Function name must be a string

<projectpath>/vendor/webonyx/graphql-php/src/Experimental/Executor/CoroutineExecutor.php:367
```

What's even more weird is the fact that when I run any of the failing tests by itself, it passes just fine. The failure only occurs when I run the entire suite.

enumag:
Ok, I think I narrowed down the issue and it seems to be a bug in the `CoroutineExecutor`. So let me ping @jakubkulhan here.

If I understand the issue correctly, it seems that `CoroutineExecutor` has a problem when a fieldResolver returns a promise which is already resolved. The problem is in [this part of code](https://github.com/webonyx/graphql-php/blob/34fe9615097be1f1d7f628ef563489973905e759/src/Experimental/Executor/CoroutineExecutor.php#L308-L332) I guess.

Since the promise is resolved already, `$this->done()` is called [here](https://github.com/webonyx/graphql-php/blob/34fe9615097be1f1d7f628ef563489973905e759/src/Experimental/Executor/CoroutineExecutor.php#L323) immediately.

This however means that `CoroutineExecutor::done()` is called before [this](https://github.com/webonyx/graphql-php/blob/34fe9615097be1f1d7f628ef563489973905e759/src/Experimental/Executor/CoroutineExecutor.php#L233) part of code is executed which means that `$this->doResolve` is still null [here](https://github.com/webonyx/graphql-php/blob/34fe9615097be1f1d7f628ef563489973905e759/src/Experimental/Executor/CoroutineExecutor.php#L377-L378).

enumag:
@vladar Can you please tell me what are the expected signatures of the `$onFulfilled` and `$onRejected` callbacks received by `PromiseAdapter::then()`? I keep running into bugs with these. Sometime they return a value sometime they return nothing and now I'm even running into a case where `$onRejected` returned a promise.

vladar:
`onFulfilled` can get any value except promise (i.e. you can get `null` or even `Throwable` if a promise was resolved with that value). `onRejected` should get `Throwable` only. Can you share a reproduction for the case when `onRejected` gets something else? I will look into it.

enumag:
@vladar yeah I thought as much. But what are they supposed to **return**? And what should I do with the returned value?

enumag:
Another extremely weird case is that here is an instance where the argument of `onFulfilled` is [passed by reference](https://github.com/webonyx/graphql-php/blob/0f4434d773885b75ad089371e95347aa8ba43c1f/src/Executor/ReferenceExecutor.php#L654). Why on earth is that? And only in this single case? Am I supposed to do something with the potentially changed value after the callback?

vladar:
You can return anything from them. Promises/A+ spec [defines the behavior](https://promisesaplus.com/#point-41) of the conformant implementation.

vladar:
It is really hard to help without specific examples.

enumag:
Unfortunately specific examples are really hard to provide since I the project is not open-source. :-(

vladar:
> Another extremely weird case is that here is an instance where the argument of `onFulfilled` is passed by reference. Why on earth is that?

Because resolved value can be an array, and we want to modify it in place vs copying all the time. See `completeValue` signature.

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
```diff
@@ -0,0 +1,149 @@
...
+            } else {
+                self::resolveWithCallable($deferred, $onRejected, $reason);
+            }
+        };
+
+        /** @var AmpPromise $adoptedPromise */
+        $adoptedPromise = $promise->adoptedPromise;
+        $adoptedPromise->onResolve($onResolve);
+
+        return new Promise($deferred->promise(), $this);
```
vladar: I am curious why do you need this method at all? Why not just do:

```php
public function then(Promise $promise, ?callable $onFulfilled = null, ?callable $onRejected = null) : Promise
{
        /** @var AmpPromise $adoptedPromise */
        $adoptedPromise = $promise->adoptedPromise;
        return new Promise($adoptedPromise->then($onFulfilled, $onRejected), $this); 
}
```

enumag:
Ok, I reimplemented `then` according to the specification. My project seems to work fine now with both executors (with #596 applied).

I'm kinda considering moving it into the Experimental subnamespace though since it's a new feature and I did have some issues with debugging it. It seems to work ok now but not like I'm an expert on Amphp or GraphQL internals. What do you think?

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
```diff
@@ -0,0 +1,149 @@
...
+
+                return;
+            }
+
+            $deferred->resolve(array_replace($promisesOrValues, $values));
+        };
+
+        all($promises)->onResolve($onResolve);
+
+        return new Promise($deferred->promise(), $this);
```
vladar: If Amp promises are A+ spec-compatible then they should work with our Promise wrapper too (as [spec defines](https://promisesaplus.com/#point-46) anything having `then` method as a Promise).

In this case, this should be as simple as:

```php
public function all(array $promisesOrValues) : Promise
{
    return new Promise(Amp\Promise\all($promisesOrValues), $this);
}
```

If it doesn't accept promises other than AmpPromise instances then your loop makes sense and it should be:
```php
public function all(array $promisesOrValues) : Promise
{
    /** @var AmpPromise[] $promises */
    $promises = [];
    foreach ($promisesOrValues as $key => $item) {
        if ($item instanceof Promise) {
            $promises[$key] = $item->adoptedPromise;
        } elseif ($item instanceof AmpPromise) {
            $promises[$key] = $item;
        }
    }
    return new Promise(Amp\Promise\all($promises), $this);
}
```

But in this case I would suggest opening an issue for Amp indicating that they are not spec-compatible in this regard.

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
enumag: Amphp doesn't have a `then`. It has `onResolve` instead but it only takes one callable and is called regardless of whether the promise was resolved with a success or with a failure.

See https://github.com/amphp/amp/blob/13930a582947831bb66ff1aeac28672fd91c38ea/lib/Promise.php#L24-L25

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
enumag: Amphp doesn't follow either Promises A or Promises A+. See https://github.com/amphp/amp/issues/150.

Another thing is that `Amp\Promise\all()` needs to be called with promises only so we need to filter the promises and then use `array_replace` to merge the resolved values into the array of `$promisesOrValues`.

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
vladar: Ah, I see. I didn't realize they are not conforming to A+ spec and just have their own implementation

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
enumag: In your code you're losing all values which were not provided as promises but as actual values..

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
enumag: Yup. Amphp was created to use coroutines and avoid callbacks. Apparently Promises A+ are not suitable for that:

```
Amp offers a single callback registration mechanism, onResolve(), which is optimized for coroutines. We think that coroutines are superior to callbacks.
```

Source: https://github.com/amphp/amp/issues/150#issuecomment-322245990

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
vladar: Yeah, I didn't realize they don't conform to any spec. This is a bit unfortunate exactly because it makes interoperability harder. But yeah, that's why we have adapters in the first place.

enumag: They made some adjustments to make the interoperability easier such as the fact that Amphp happily accepts and adapts any ReactPHP / A+ promises you throw at it into their promises. It's just that we need to do the opposite here so it's a bit more complicated.

enumag:
PHPStan is failing on travis but I have no idea why. Locally it's perfectly fine.

simPod:
@enumag might be the memory limit? Gets increased in #592.

enumag:
@simPod Yup, higher memory limit helped.

enumag:
Bump. I think this is ready to be merged.

vladar:
Nice job! 🚢

simPod:
@vladar seems like merge broke build, some unit tests are failing

vladar:
Yeah, I fixed those. But what's interesting is that tests were passing initially in CI. Also, can you take a look at PHPStan failures? The problem I have is that it fails for me with 500+ errors for some reason while CI only has a couple of errors. Not sure what's going on there.

simPod:
Probably because it was not rebased and master was far ahead. I'll check.

simPod:
Upgrade PHPStan patch versions. They were some inconsistencies in reported errors between different platforms, fixed in patch versions. Should be all ok now.

vladar:
Awesome! It works for me know. Thanks!

PR #553, opened 2019-10-10
Add message level when using custom warning handler

mcg-web:
I need to make silent `error_trigger` but since it was already refused [here](https://github.com/webonyx/graphql-php/pull/234#issuecomment-362502148) I'm going to implement my own handler make it silent. This will fix the issue when using` isTypeOf` the first entry is null because of the `warnOnce` that trigger error not silently. Thanks to @murtukov while trying to write some documentation on `isTypeOf`...

### DIFF
--- a/src/Error/Warning.php
+++ b/src/Error/Warning.php
@@ -96,22 +96,26 @@ final class Warning
 
     public static function warnOnce(string $errorMessage, int $warningId, ?int $messageLevel = null) : void
     {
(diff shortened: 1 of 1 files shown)


vladar:
> This will fix the issue when using isTypeOf the first entry is null because of the warnOnce that trigger error not silently

Why does it return null? I guess Symfony has a custom error handler that throws an exception when there is a warning? 

Anyways, the change seems reasonable. Thanks!

PR #554, opened 2019-10-11
Modifies the chr() check to properly handle \u0000 codes in the range…

jakevoytko:
… [127, 255] in the lexer

This caused the lexer to output invalid UTF-8 for input like 'pok\u00E9mon'. The output for the é
would be the decimal byte 233, which would indicate that it should be a 4 byte unicode sequence,
but the following bytes didn't have the leading 10 prefix, so the sequence was invalid UTF-8.

### DIFF
--- a/src/Utils/Utils.php
+++ b/src/Utils/Utils.php
@@ -442,9 +442,6 @@ class Utils
      */
     public static function chr($ord, $encoding = 'UTF-8')
     {
-        if ($ord <= 127) {
-            return chr($ord);
-        }
         if ($encoding === 'UCS-4BE') {
             return pack('N', $ord);
         }
--- a/src/Utils/Utils.php
+++ b/src/Utils/Utils.php
@@ -442,7 +442,7 @@ class Utils
      */
     public static function chr($ord, $encoding = 'UTF-8')
     {
-        if ($ord <= 255) {
+        if ($ord <= 127) {
             return chr($ord);
         }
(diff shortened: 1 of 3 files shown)


jakevoytko:
I think additional modification is needed to make it worth with strings like '𝕄𝕖𝕥𝕒𝕝𝕝𝕚𝕔' (encoding `'"\ud835\udd44\ud835\udd56\ud835\udd65\ud835\udd52\ud835\udd5d\ud835\udd5d\ud835\udd5a\ud835\udd54"'`). Consider holding off on reviewing it

src/Utils/Utils.php
```diff
@@ -442,7 +442,7 @@ public static function printSafe($var)
      */
     public static function chr($ord, $encoding = 'UTF-8')
     {
-        if ($ord <= 255) {
+        if ($ord <= 127) {
```
vladar: I guess this check only makes sense when $encoding is `UTF-8`. But we can even try deleting this check altogether since it was a performance optimization that is not required anymore as far as I remember.

jakevoytko: I pushed another commit removing the check entirely per your feedback.

The problem I mentioned about characters like "𝕄𝕖𝕥𝕒𝕝𝕝𝕚𝕔" are a separate problem (UTF-16 surrogates aren't handled in the lexer). I'll mail a patch for that early next week

Issue #35, opened 2016-04-24
Deprecate callbacks in field types


vladar:
Deprecate following field definition form:
```php
[
    'name' => 'Foo',
    'fields' => [
        'bar'  => ['type' => function()  use(&$barType) { return $barType;}],
        'baz'  => ['type' => function()  use(&$bazType) { return $bazType;}],
    ]
]
```
in favor of:
```php
[
    'name' => 'Foo',
    'fields' => function() use (&$barType, &$bazType) {
        return [
            'bar' => ['type' => $barType],
            'baz' => ['type' => $bazType],
        ];
    }
]
```

ooflorent:
Is there any reason why? What's the motivation behind this change?

vladar:
This seems redundant. Also first notation adds some performance overhead because every field type has to be checked with `is_callable` and then validated against base `Type` class.

Pretty much everything you can do with first notation can be done with second as well. Also that's how reference js implementation works. It turns out that maintaining this library is way easier when there is little deviation from reference implementation (at least until things settle down with GraphQL itself).

But if you have any use-cases which require this feature - please describe them. That's pretty much the reason why I created these issues - to start a discussion if somebody needs them %)

mcg-web:
:+1: for this change! it will be much more easier to implements dynamics fields without having to used hacks.

vladar:
Done in https://github.com/webonyx/graphql-php/commit/c11f25794a16863ad727c844ca9083bf1cb3c3d1

decebal:
must compliment you guys for referincing this issue in the actual error message! :1st_place_medal:

Alexandru-Dobre:
Is there any chance the first notation could be permitted once again ?
We have a library that uses this and json models in order to build a loopback-like php graphql api factory. 
Having to move the callback to the fields rather than the type of each field would greatly complicate things since the field is linked to the middleware-connector pipeline during field construction. Here's a link to the concerned code : https://github.com/ampize/alambic/blob/master/src/Alambic/Alambic.php

vladar:
@Alexandru-Dobre We can discuss it, but first I would like to figure out why it is impossible to change the code of your library for this notation (as it is also the only way reference js implementation allows you to define fields). I am pretty sure there should be a way to do this.

What will happen if you change [these lines](https://github.com/ampize/alambic/blob/master/src/Alambic/Alambic.php#L287-L297) to:
```php
$typeArray = [
    'name' => $type['name'],
[...]

PR #557, opened 2019-10-17
Lazy loading of types

shmax:
Implements #425 
Proof of concept of full support for lazy loading of types.

In a nutshell, you can substitute a `callable` any place that you would ordinarily supply a `Type`, and it won't be executed by the library until it is actually needed.

For example, you can create a `Types` class that looks like this:
```php
class Types {
	public static function get(string $classname)
	{
		return function() use ($classname) {
			return static::fromCache($classname);
		};
	}

	protected static function fromCache(string $classname) {
		$parts = explode("\\", $classname);
		$cacheName = strtolower(preg_replace('~Type$~', '', $parts[count($parts) - 1]));
		$type = null;

		if (!isset(self::$types[$cacheName])) {
			if (class_exists($classname)) {
				$type = new $classname();
			}

			self::$types[$cacheName] = $type;
		}

		$type = self::$types[$cacheName];

		if (!$type) {
			throw new Exception("Unknown graphql type: " . $classname);
		}
		return $type;
	}

    public static function currency() { return static::get(CurrencyType::class);}
    // ... I explicitly stub out these type methods for the benefit of intellisense and [stanning](https://github.com/phpstan/phpstan), but they're not necessary
}
    
```

And then your field/arg definitions look the same as usual:

```php
...
	'currency' => [
		'type' => Types::currency(), // this won't load until it is specifically needed!
		'description' => 'A currency record',
```

It may be difficult to imagine this being of any appreciable benefit for small or toy programs, but in my real project I dynamically generate my mutation types (using [this](https://github.com/shmax/graphql-php-validation-toolkit)) and doing even a small mutation triggers a flood of activity and the instantiation of `656` types(!).

With this change, that number drops to `1`.

I have run this in the profiler:
[without lazy loading](https://blackfire.io/profiles/be6a1083-7620-4043-a4d0-219add8d8f01/graph)
[with lazy loading](https://blackfire.io/profiles/50800a5d-0b42-4e3e-a40b-6c9fd87f86d2/graph)

For me, this equates to a savings of around a quarter of a second and 3MB of memory.

So, what do you think? It's only around 50 lines of mostly trivial code. If you think it looks promising I'll start in on tests and documentation and further polishing.


vladar:
Thanks for the PR. Ironically we had this in place before and removed in https://github.com/webonyx/graphql-php/issues/35

I realize that true lazy types are a MUST but one other option worth exploring is to allow defining types using SDL instead (and maybe always require type loader to be defined):

```php
$type = new ObjectType([
    'name' => 'Foo',
    'fields' => function() {
        return [
            'bar' => [
                'type' => 'String!',
            ],
        ];
    }
]);
```

This will likely require a significant rework but it provides some other benefits as well in addition to real lazy-loading (less typing, also automatically ensure there is only one instance of a type in schema).

I guess the approach we choose should be weighted carefully, so I encourage everyone interested to post their opinions here.

mcg-web:
That exactly the way we has designed the GraphQL bundle and this really ease types declaration. I'm 100% for this change! This will not be a big change since we can make both declaration lives together by requiring type loader only if type is string instead of an object.

shmax:
Thanks much for the quick feedback!

> (less typing, also automatically ensure there is only one instance of a type in schema)

I don't personally see "less typing" happening, at least not with my coding standard--I would never type string literals into my field values; too much room for human error. I would just wind up using a constant, anyway, so it would just look like this:

```
...
	'currency' => [
		'type' => Types::currency, // a static constant on Types, "currency"
		'description' => 'A currency record',  

```

> also automatically ensure there is only one instance of a type in schema

I believe your library already throws an exception if the same name pops up twice. 

All that said, I think it would be pretty trivial to also support string values. If you look at my `Types::get` method, you can see it already operates on a string, so it would just be a matter of routing to the type loader, as you suggested.

vladar:
> I don't personally see "less typing" happening

Compare `[String!]!` and `Type::nonNull(Type::listOf(Type::nonNull(Type::string())))`

shmax:
> Compare [String!]! and Type::nonNull(Type::listOf(Type::nonNull(Type::string())))

Well, that's a built-in type, and it wouldn't benefit from lazy-loading, right? Certainly we should consider if any of this makes sense for SDL, but I'm not sure I see the wisdom in mixing the two things together. And you still have the human error angle in your approach.

mcg-web:
See how with [some light checks](https://github.com/overblog/GraphQLBundle/blob/master/src/Resolver/TypeResolver.php#L93-L107) we can make this less error prone.

shmax:
> FYI Blackfire profiles seem to be unavailable for global access (maybe require a login)

Ah, sorry, I'm still sort of new to Blackfire. They should be public, now.

shmax:
> See how with some light checks  we can make this less error prone.

I appreciate the thought, but it doesn't seem wise to me to be repeating the work that is already performed by the lexer. 

But why would I be parsing and lexing? I define my schemas in PHP instead of SDL because they tend to be dynamic and I want what little safety can be supplied by the language checker itself.

We may have different ideas about what it means to pass a string to the type loader--to me, it's just any identifier that has meaning to you; it's up to you to interpret it. In my sample, the string is the namespace of the class itself (and I instantiate it directly when the time comes), so there is little room for human error.

If I wanted to use SDL, I would use SDL. Trying to mix it in with schemas defined with PHP idioms seems like a bad idea to me; for one thing, you're now adding more work back into the equation in the form of additional parsing and lexing that doesn't need to be there.

But what about full SDL? Does any of this make sense? I don't know, yet; but I'll find out.

shmax:
Hi @vladar and @mcg-web, I'm working away on tests and cleanup, but I have a question about this:

https://github.com/webonyx/graphql-php/blob/master/src/Type/Definition/FieldDefinition.php#L180

This is annotated as `Type` for the static checker, but then you have an entire test class that tries writing different kooky values to it (for example, [here](https://github.com/webonyx/graphql-php/blob/master/tests/Type/ValidationTest.php#L1316)).

This seems to weaken the effectiveness of the static type checking. If you're serious about `FieldDefinition::$type` being a `Type` (or with this PR, a `callable`), would it not make more sense to throw an error in the `FieldDefinition` constructor, rather than waiting for the validation step?

Please let me know your thoughts so I can make the appropriate changes to the constructor and tests.

shmax:
I took a stab at writing some tests, and fixed enough of the scrutinizer noise for it to pass.

Regarding SDL string values for types: it may or may not be a good idea, but unless I'm mistaken it can't replace function-based loading. At best it would have to be an alternative option to what I'm proposing, here. If we can agree on that, then I'd just as soon let someone else figure it out if they can come up with a good use case for it. 

In short, it's a different feature than this one, and the two ideas seem to be orthogonal to one another, so it's not necessary to solve both of them at the same time. 

I also had a look at full SDL-defined schemas, and as far as I can tell they don't overlap with lazy type loading, either.

So, please tell me what we need to do move forward with this. Thanks!

vladar:
> This is annotated as Type for the static checker, but then you have an entire test class that tries writing different kooky values to it (for example, here).

Those fields are effectively user input. The idea was that we didn't want to validate user input (provided as config arrays) on every request.

Instead, you choose when to validate them (you can run `schema->validate()` in the dev environment only + maybe during some build/deploy step)

Static type checking doesn't help much when it comes to validating config arrays.

vladar:
> but unless I'm mistaken it can't replace function-based loading

Why do you think so? The idea is that typeloader remains the same but you specify field types as SDL expressions vs object references (or functions returning those object references)

> In short, it's a different feature than this one, and the two ideas seem to be orthogonal to one another, so it's not necessary to solve both of them at the same time.

For me those are mutually exclusive. Mostly because it is a pain to maintain multiple ways of defining type relations. Especially when people start mixing different styles together.

And I see how SDL-based types fix the lazy-loading (as well as some other things mentioned above). Obviously it has it's own downsides (parsing overhead, no strict typing, etc)

The question is about trade-offs we choose.

> So, please tell me what we need to do move forward with this. Thanks!

 It is highly likely that we'll go the route you propose (as it's easier and backwards compatible). But I'd like to hear more feedback from people using the library before making a decision.

shmax:
> Those fields are effectively user input. The idea was that we didn't want to validate user input (provided as config arrays) on every request.
> 
> Instead, you choose when to validate them (you can run `schema->validate()` in the dev environment only + maybe during some build/deploy step)

It's not user input, though. If inline schemas are an alternative to SDL, and SDL is not user input, then neither are inline schemas. This kind of "validation" is ideally handled by the PHP language, or static checking. 
 
> Static type checking doesn't help much when it comes to validating config arrays.

It does help. You have to raise your PHP stan level to 3, and use array shapes in your doc blocks (The feature isn't quite ready for prime time; for some reason it doesn't support multiline array definitions, which is probably a deal-breaker for the kind of complexity we have). I've been experimenting with it in a separate branch, and it seems to work well.

But that doesn't matter in this case. Ideally, I should be able to use a native PHP typehint of type `Type` for my `Type::resolveLazyType` method, but I can't at the moment because you allow all those weird values through. If you threw an exception early I wouldn't have to worry about it.

Granted, this can probably wait until we work on PHPStan level 3, which I wouldn't personally want to tackle without the array shape stuff (which I'm currently investigating; will let you know).

shmax:
> Why do you think so? 

I assume we're talking about something like this, right?

```
type: "MySpecialLazyLoadedType"
```

This could work for trivial cases, but the problem here is that if you require me to know the name of a type, then it's not truly being lazy-loaded. The power of inline schemas is that they are programmatic. We don't necessarily know the fields of our type until we create it, and we don't necessarily know the _name_ until we create it.

> It is highly likely that we'll go the route you propose (as it's easier and backwards compatible). But I'd like to hear more feedback from people using the library before making a decision.

Wonderful! Completely understandable, and there's no real rush, here (I have my fork to work with, after all). But is there anyone we should ping?

vladar:
> It does help. You have to raise your PHP stan level to 3, and use array shapes in your doc blocks

This is interesting! But there is a bunch of projects not using PHPStan who still use this library (for example in the old-school CMS world). So given the wide range of use-cases, this is not really an option for the validation at the library level.

> We don't necessarily know the fields of our type until we create it, and we don't necessarily know the name until we create it.

This doesn't sound right to me. Lazy types are somewhat similar to PHP classes autoloading. If your clients know what they are querying (relying on a static schema), how does it happen that you do not know it?

> But is there anyone we should ping?

I guess we'll have to create a separate issue (with options listed here), mention it in the next release notes to collect some feedback.

shmax:
> This is interesting! But there is a bunch of projects not using PHPStan who still use this library (for example in the old-school CMS world). So given the wide range of use-cases, this is not really an option for the validation at the library level.

Right, you can't assume people are using PHPStan, but we still have native PHP typehints. I should be able to just typehint my return as `Type`, but I currently can't because of the architectural decision to not do any kind of runtime type-checking until the validation stage. I'm not going to press the issue right now--let's save it for PHPStan level 3, which will force us to reevaluate a lot of things (for example, `OutputType` does not itself implement `AbstractType`, so you can't really use it in places expected `Input`).

> This doesn't sound right to me. Lazy types are somewhat similar to PHP classes autoloading. If your clients know what they are querying (relying on a static schema), how does it happen that you do not know it?

Clients don't generally use the names of types in their queries (there are exceptions, such as specifying query variables). They use the names of fields and args. Where do they get the names of the fields and args? Well, they inspect the API, which is generated from an introspection query, which forces all types to load.

The concept is similar to--but not identical--to PHP auto-loading. In auto-loading, there is no wiggle room--the requested name must match the name of the class, which must match the name of the file (of course, you can write your own auto-loader to do something different). With `GraphQl` types, the name is just one of many properties on the type. It doesn't have to match the name of the class (although that's what will happen if you allow it to be inferred)--you can use anything you like. Since it is up to the author of the inline schema to come up with a name, it is programmatic. There is no rule that says it must match the name of the class, or even that your type must subclass anything.

Hopefully you can agree with all that so far. In summary:

1. Inline schemas are programmatic
2. The name is just one of many properties that will be decided by the developer.
3. There are no hard rules about what the name should be.
Therefore, a truly lazy-loaded type can't require advance knowledge of the name anymore than it can require advance knowledge of the fields.

I wrote an entire library around automatically generating types: https://github.com/shmax/graphql-php-validation-toolkit

Lazy loading these generated types is simply not possible with your string-based idea.

shmax:
By the way, I have a PR open to support multiline array shape syntax: https://github.com/phpstan/phpdoc-parser/pull/33

Seems to be doing okay. The next roadblock is that slevomatic/code-sniffer has not done a release of the array shape support (or any release since March), and the maintainer is being [bizarre about it](https://github.com/slevomat/coding-standard/issues/750#issuecomment-545531086)

src/Type/Definition/Type.php
```diff
@@ -314,10 +315,25 @@ public static function isAbstractType($type) : bool
...
+    }
+
+    /**
+     * @param callable|Type $type
+     */
+    public static function resolveLazyType($type)
+    {
+        if (is_callable($type)) {
+            return $type();
+        }
```
vladar: I guess we need to throw if loaded value is not a type similar to how we had it in the past:

https://github.com/webonyx/graphql-php/blob/c11f25794a16863ad727c844ca9083bf1cb3c3d1/src/Type/Definition/Type.php#L202-L207

src/Type/Definition/UnionType.php
```diff
@@ -57,7 +58,7 @@ public function isPossibleType(Type $type) : bool
         if ($this->possibleTypeNames === null) {
             $this->possibleTypeNames = [];
             foreach ($this->getTypes() as $possibleType) {
-                $this->possibleTypeNames[$possibleType->name] = true;
+                $this->possibleTypeNames[Type::resolveLazyType($possibleType)->name] = true;
```
vladar: We don't need it. `getTypes` always returns resolved types

src/Type/Schema.php
```diff
@@ -353,15 +355,18 @@ private function loadType($typeName)
...
                 sprintf(
-                    'Type loader is expected to return valid type "%s", but it returned %s',
+                    'Type loader is expected to return a callable or valid type "%s", but it returned %s',
                     $typeName,
                     Utils::printSafe($type)
                 )
             );
         }
+
+        $type = Type::resolveLazyType($type);
```
vladar: Not sure if we should allow `typeLoader` to return `callable`. It is basically useless because when it's called - we already need the actual type. Do you have any reasons to allow it?

src/Type/Schema.php
```diff
@@ -411,6 +416,8 @@ private function getPossibleTypeMap()
         if ($this->possibleTypeMap === null) {
             $this->possibleTypeMap = [];
             foreach ($this->getTypeMap() as $type) {
+                $type = Type::resolveLazyType($type);
```
vladar: `getTypeMap` should also always return a map of fully resolved types. This should be removed. Any errors would indicate that `getTypeMap` implementation must be adjusted

src/Utils/TypeInfo.php
```diff
@@ -157,6 +157,9 @@ public static function extractTypes($type, ?array $typeMap = null)
         if ($type instanceof WrappingType) {
             return self::extractTypes($type->getWrappedType(true), $typeMap);
         }
+
+        $type = Type::resolveLazyType($type);
```
vladar: I would enforce type to be already resolved in here. So this should be removed. If we correctly resolve types in `UnionType::getTypes`, `ObjectType::getInterfaces`, `FieldDefinition::getType`, etc then this method will always deal with fully loaded types (and it should)

shmax:
@vladar Thanks for all the comments! If you don't mind, I think I would like to wait for the PHPStan Level 3 branch to merge, then I'll do another pass over this and clean everything up.

src/Type/Definition/ListOfType.php
```diff
@@ -4,22 +4,29 @@
...
 
     public function toString() : string
     {
-        return '[' . $this->ofType->toString() . ']';
+        return '[' . $this->getOfType()->toString() . ']';
+    }
+
+    public function getOfType()
+    {
+        return Type::resolveLazyType($this->ofType);
```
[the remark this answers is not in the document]
shmax: Right, I'm hoping this will all tighten up after level 3. Will revisit...

src/Executor/ReferenceExecutor.php
```diff
@@ -940,9 +940,13 @@ private function completeAbstractValue(AbstractType $returnType, $fieldNodes, Re
     {
         $exeContext  = $this->exeContext;
         $runtimeType = $returnType->resolveType($result, $exeContext->contextValue, $info);
+
         if ($runtimeType === null) {
             $runtimeType = self::defaultTypeResolver($result, $exeContext->contextValue, $info, $returnType);
         }
+
+        $runtimeType = is_callable($runtimeType) ? $runtimeType() : $runtimeType;
```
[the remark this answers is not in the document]
shmax: Sounds good. 👍

src/Type/Definition/FieldDefinition.php
```diff
@@ -129,7 +129,7 @@ public static function defineFieldMap(Type $type, $fields)
                     );
                 }
 
-                $fieldDef = self::create(['name' => $name, 'type' => $field]);
+                $fieldDef = self::create(['name' => $name, 'type' => Type::resolveLazyType($field)]);
```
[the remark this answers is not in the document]
shmax: You're right! Seems to still work just fine...

src/Type/Definition/ListOfType.php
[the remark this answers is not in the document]
shmax: I think we're okay, here--the lazy load method is type-hinted with `Type`...

src/Type/Definition/Type.php
shmax: Ah, no, actually I just type-hinted the return as `Type`, so we can let natural language-level checking take care of this for us.

src/Type/Definition/UnionType.php
shmax: Right again! 🔥

src/Type/Schema.php
shmax: Sorry, I don't understand--what are you suggesting?

src/Type/Schema.php
shmax: Right again! removed...

src/Utils/TypeInfo.php
shmax: Right!

tests/Type/ValidationTest.php
```diff
@@ -1346,28 +1328,6 @@ public function testRejectsWithReleventLocationsForANonOutputTypeAsAnObjectField
         );
     }
 
-    // DESCRIBE: Type System: Interface fields must have output types
-
-    /**
-     * @see it('rejects an Object implementing a non-type values')
-     */
-    public function testRejectsAnObjectImplementingANonTypeValues() : void
```
shmax: Removed this because now that `ObjectType::getInterfaces` calls `Schema::resolveType`, anything that is not a `Type` will violate the type hint and throw a natural error.

shmax: @vladar I was thinking; if you really want this `validate` stuff to work the way it used to, maybe what I could do is change the type hint to an `assert`, and I could throw a crafted exception. I could then catch that in the validation code, and create an `Error` as before. Let me know what you think.

src/Type/Schema.php
vladar: I suggest disallowing `typeLoader` to return `callable`. It should always return a specific type.

src/Type/Definition/FieldArgument.php
```diff
@@ -82,7 +85,13 @@ public static function createMap(array $config) : array
      */
     public function getType() : Type
     {
-        return $this->type;
+        $type = Schema::resolveType($this->type);
+        assert(
```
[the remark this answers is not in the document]
vladar: The assertion stuff is really cool! We could probably utilize it more in future in the lib (as it makes sense to have stricter checks in dev which could be disabled and have 0 cost in prod)

tests/Type/ValidationTest.php
vladar: Yeah, I am curious which error message and line/string will users get without this? Because right now it should be obvious what is wrong with typedef. But if you get some cryptic error message in an unrelated place of code - the DX is very different

src/Type/Definition/FieldDefinition.php
```diff
@@ -179,7 +181,10 @@ public function getArg($name)
      */
     public function getType() : Type
     {
-        return $this->type;
+        $type = Schema::resolveType($this->type);
+        assert($type instanceof OutputType, new Error('Expected type of FieldDefinition to implement OutputType but got: ' . Utils::printSafe($type)));
```
vladar: I would suggest using `InvariantViolation` instead of an `Error` exception for assertions.

vladar:
Not sure if it's ready for review? Please ping me when you are done.

src/Type/Definition/FieldArgument.php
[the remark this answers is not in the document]
simPod: The only problem with those is that it persuades phpstan or any other static analysis about something, that is not real in prod runtime. 

Eg. `assert($x !== null)` narrows `int|null` into `int` for phpstan but in reality `null` value can be passed through it.

I still don't know how to approach it. I'm rather using Assertion libraries that throw exception everytime assertion fails currently in my job . But it's not suitable way here either as it would add some cost.

tests/Type/ValidationTest.php
shmax: Fair enough. I might start another PR to try to advance some of this assert stuff, which should smooth this out and make phpstan level 4 easier to do.

shmax:
> Not sure if it's ready for another review pass? Please ping me when you are done.

Oh, sorry, I didn't realize I was expected to do that. Let me do another round of fixes today, and I'll let you know. Thanks!

src/Type/Definition/FieldArgument.php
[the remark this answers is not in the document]
shmax: You're right, of course. Asserts fall into the same family of precautions as unit tests and static checking: they aren't on the job in a production environment. But in development they're fantastic--they create a strong relation between the static type and the run-time type. 

It's true that there is some room for divergence on production, but remember that these things are designed to catch programming errors, not user errors, so in principle there's only potential for danger here for code that you not only didn't test, but _didn't even try_ in your development environment.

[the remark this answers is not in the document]
simPod: I encountered situations where these assertions lied even in tested and dev-run code though 😶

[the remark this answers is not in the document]
shmax: Then I suppose we'll have to watch for reports of such cases and adjust our methodology accordingly.

src/Type/Definition/FieldDefinition.php
shmax: Fixed in 932f5dfa75d76125adb8de3609132e3d99b85ebe.

(sorry for the slow progress--I moved to a new computer this week, and work has been hectic)

shmax:
Sorry for the slow progress--I'm having some trouble getting my environment up and running on my new machine--but I'm still working on it. More soon...

vladar:
So what's the state of this PR at the moment?

shmax:
> So what's the state of this PR at the moment?

Sorry again for the lack of progress--the holidays happened, and now I'm traveling for work and living out of a suitcase, etc, etc.

I had promised to try to restore the "validation" stuff to its former glory, but if that can wait then I suppose there's nothing stopping you from merging this as-is (once I fix the conflicts). Up to you, really...

shmax:
@simPod I wonder if I can trouble you for some insight on why the tests are failing? They seem fine when run by travis, but in the other contexts (github actions?) they bomb out, and many of them don't even make it as far as the tests.

simPod:
@shmax sure, will look into it.

simPod:
@shmax exactly what I was saying here https://github.com/webonyx/graphql-php/pull/557#discussion_r349496079 ;)

`zend.assertions` is not set in GA. Relying on `assert()` behaviour is not safe even in tests in my opinion.

shmax:
> zend.assertions is not set in GA. Relying on assert() behaviour is not safe even in tests in my opinion.

So what is different about the setup? Why does it work in travis but not in github actions?

simPod:
Dunno how Travis sets up its PHP. You could try enabling the option manually but there's no way to ensure it through code so it will always be fragile.

shmax:
Well, this bit is what enables the exception-throwing for tests: 

https://github.com/webonyx/graphql-php/pull/557/files#diff-98c12a693941ae3c4f50872403416471R9

Are we sure that this config file is being loaded?

simPod:
> Are we sure that this config file is being loaded?

Yes, definitely. And `zend.assertions=1` can be set only through php.ini so cannot do it via phpunit config.

shmax:
@simPod I configured `zend.assertions` as you suggested, and that was exactly the problem--everything is working much better, now. Thank you! By the way, I must say I'm very impressed by the speed of the ci--you guys did a fabulous job on that.

shmax:
@simPod I don't want to abuse the favor, but any idea what this last error is about? https://github.com/webonyx/graphql-php/pull/557/checks?check_run_id=379209909#step:8:315

simPod:
@shmax yes, it's #605

shmax:
> @shmax yes, it's #605

Ah, understood, thanks. 👍

simPod:
@shmax it's fixed in master, rebase should do the trick

shmax:
> @shmax it's fixed in master, rebase should do the trick

Yee-haw, all green! Thanks so much...

shmax:
Hey guys, sorry for the super long delay. I've fixed all the stanning, linting, and test issues with this PR. In the interest of speeding things along, I've removed all the native `assert` stuff; I still feel confident it's the way to go, and not having it required me to loosen some of the compound PHPStan annotations, but I know @simPod doesn't like it and I don't want to fight that battle now. 

I restored the tests to their original condition, with one exception: I had to stick with the removal of `ValidationTest::testRejectsAnObjectImplementingANonTypeValues`. There's a native type hint of `Type` on `Schema::resolveType`, and for this test to continue to function the way it did before I would have to remove the type hint. I hope you agree that weakening an important native typehint for the benefit of a separate validation step would be a step backward, but if you really want to insist on it I can keep working on it.

If it's not a blocker, then I think this is ready for final review. Thanks much for your patience.

src/Type/Definition/FieldArgument.php
```diff
@@ -4,10 +4,13 @@
 
 namespace GraphQL\Type\Definition;
 
+use GraphQL\Error\Error;
 use GraphQL\Error\InvariantViolation;
 use GraphQL\Language\AST\InputValueDefinitionNode;
+use GraphQL\Type\Schema;
 use GraphQL\Utils\Utils;
 use function array_key_exists;
+use function assert;
```
simPod, line 13: Might be unused. Wonder why it was not caught by CS check 🤔

src/Type/Definition/FieldDefinition.php
```diff
@@ -7,7 +7,9 @@
 use GraphQL\Error\Error;
 use GraphQL\Error\InvariantViolation;
 use GraphQL\Language\AST\FieldDefinitionNode;
+use GraphQL\Type\Schema;
 use GraphQL\Utils\Utils;
+use function assert;
```
simPod, line 12: Might be unused. Wonder why it was not caught by CS check 🤔

src/Type/Definition/InputObjectField.php
```diff
@@ -7,8 +7,10 @@
 use GraphQL\Error\Error;
 use GraphQL\Error\InvariantViolation;
 use GraphQL\Language\AST\InputValueDefinitionNode;
+use GraphQL\Type\Schema;
 use GraphQL\Utils\Utils;
 use function array_key_exists;
+use function assert;
```
simPod, line 13: Might be unused. Wonder why it was not caught by CS check 🤔

src/Type/Definition/NonNull.php
```diff
@@ -4,13 +4,25 @@
...
 
-    public function __construct(NullableType $type)
+    /**
+     * code sniffer doesn't understand this syntax. Pr with a fix here: waiting on https://github.com/squizlabs/PHP_CodeSniffer/pull/2919
+     * phpcs:disable
+     * @param  (NullableType&Type)|callable $type
+     */
+    public function __construct($type)
     {
+        assert($type instanceof NullableType || is_callable($type), new InvariantViolation('NonNull constructor expects an instance of NullableType, or a callable'));
```
simPod, line 24: This can be dropped, should be guarded by PHPStan

src/Type/Definition/ObjectType.php
```diff
@@ -147,7 +150,7 @@ public function hasField($name)
     public function getFields()
     {
         if ($this->fields === null) {
-            $fields       = $this->config['fields'] ?? [];
+            $fields       = $this->config['fields'] ?: [];
```
simPod, line 153: The short ternary operator is not strict and disallowed by CS, can you mb rebase onto master to see if it gets caught?

https://github.com/doctrine/coding-standard/pull/57

src/Type/Definition/ResolveInfo.php
```diff
@@ -40,7 +40,7 @@ class ResolveInfo
      * Expected return type of the field being resolved.
      *
      * @api
-     * @var OutputType&Type
+     * @var Type
```
simPod, line 43: Wasn't `OutputType&Type` more accurate?

src/Type/Definition/Type.php
```diff
@@ -280,29 +283,14 @@ public static function isAbstractType($type) : bool
 
     /**
      * @param mixed $type
-     *
-     * @return mixed
      */
-    public static function assertType($type)
+    public static function assertType($type) : Type
```
simPod, line 287: Personally I'd drop this method. Wonder what others think.

For me types eg. here are sufficient:

```php
    /**
     * @param callable():Type|Type $type
     */
    public function __construct($type)
    {
        $this->ofType = is_callable($type) ? $type : Type::assertType($type);
    }
```

src/Type/Definition/UnionType.php
```diff
@@ -53,7 +55,7 @@ public function __construct(array $config)
 
     public function isPossibleType(Type $type) : bool
     {
-        if (! $type instanceof ObjectType) {
+        if (! ($type instanceof ObjectType && $type instanceof Type)) {
```
simPod, line 58: Why this `$type instanceof ObjectType` is not enough? `ObjectType extends Type` so it seems to me that 2nd part will always be true if 1st part is.

src/Type/Schema.php
```diff
@@ -359,12 +364,27 @@ private function loadType(string $typeName) : ?Type
...
+    /**
+     * @param Type|callable():Type $type
+     */
+    public static function resolveType($type) : Type
+    {
+        if (is_callable($type)) {
+            return $type();
+        }
+
+        /**
```
simPod, line 382: Is this comment necessary?

tests/Type/LazyTypeLoaderTest.php
```diff
@@ -0,0 +1,395 @@
...
+use GraphQL\Type\Definition\InterfaceType;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Schema;
+use PHPUnit\Framework\TestCase;
+use stdClass;
+use Throwable;
+use function lcfirst;
+
+class LazyTypeLoaderTest extends TestCase
```
simPod, line 20:
```suggestion
final class LazyTypeLoaderTest extends TestCase
```

tests/Type/LazyTypeLoaderTest.php
```diff
@@ -0,0 +1,395 @@
...
+    /** @var callable */
+    private $typeLoader;
+
+    /** @var string[] */
+    private $calls;
+
+    /** @var Type[] */
+    private $loadedTypes = [];
+
+    protected function _lazyLoad(string $name)
```
simPod, line 54: missing return typehint

tests/Type/LazyTypeLoaderTest.php
```diff
@@ -0,0 +1,395 @@
...
+                        'args' => [
+                            'input'           => Type::nonNull($this->postStoryMutationInput),
+                            'clientRequestId' => Type::string(),
+                        ],
+                    ],
+                ];
+            },
+        ]);
+
+        $this->typeLoader = function ($name) : ?callable {
```
simPod, line 176:
```suggestion
        $this->typeLoader = function (string $name) : ?callable {
```

tests/Type/LazyTypeLoaderTest.php
simPod: why not simply this?

```suggestion
    private function lazyLoad(string $name)
```

I see there's no sniff for that, I'll try to introduce it...

src/Type/Definition/FieldArgument.php
```diff
@@ -77,12 +80,9 @@ public static function createMap(array $config) : array
         return $map;
     }
 
-    /**
-     * @return InputType&Type
```
simPod, line 81: This should stay, no?

src/Type/Definition/ObjectType.php
simPod: https://github.com/webonyx/graphql-php/pull/645 :)

src/Type/Definition/FieldArgument.php
shmax: Right you are. fdcce6e

src/Type/Definition/FieldDefinition.php
shmax: f89c842

src/Type/Definition/InputObjectField.php
shmax: 730e117

src/Type/Definition/NonNull.php
shmax: 48106ac

src/Type/Schema.php
shmax: Evidently not. e1b363e

tests/Type/LazyTypeLoaderTest.php
shmax: Sure, if you like. 1e48d6d

tests/Type/LazyTypeLoaderTest.php
shmax: 51caeef 

> why not simply this?

No objection. c085169

tests/Type/LazyTypeLoaderTest.php
shmax: 9f2cf34

src/Type/Definition/FieldArgument.php
shmax: I'd like that, but without some kind of assert my only other option is a cast, which sort of defeats the point of checking.

shmax:
@simPod @vladar Hey guys, please let me know what I have to do to move forward with this PR. I'm about two weeks away from the point where my main application can't progress further without this, so let me know whatever it is you don't like, and I'll snip it out. I aim to please!

src/Type/Definition/FieldArgument.php
simPod: so can we assert / ignore in static analysis for now? This can be IMO resolved better through templates later

src/Type/Definition/ListOfType.php
```diff
@@ -4,24 +4,38 @@
 
 namespace GraphQL\Type\Definition;
 
+use GraphQL\Type\Schema;
+use function is_callable;
+
 class ListOfType extends Type implements WrappingType, OutputType, NullableType, InputType
 {
     /** @var Type */
```
simPod, line 12: this can also be callable according to modified constructor

src/Type/Definition/InputObjectField.php
```diff
@@ -50,12 +51,9 @@ public function __construct(array $opts)
         $this->config = $opts;
     }
 
-    /**
-     * @return Type&InputType
```
simPod, line 54: should be kept in order not to lower type coverage

src/Type/Definition/FieldDefinition.php
```diff
@@ -174,12 +175,9 @@ public function getArg($name)
         return null;
     }
 
-    /**
-     * @return OutputType&Type
```
simPod, line 178: should be kept in order not to lower type coverage

src/Type/Schema.php
```diff
@@ -57,7 +57,7 @@ class Schema
      */
     private $resolvedTypes = [];
 
-    /** @var array<string, array<string, ObjectType>>|null */
+    /** @var array<string, array<string, ObjectType|UnionType>> */
```
simPod, line 60: This seems it can still be null

simPod: Also the UnionType is missing in the getter method return type

src/Type/Definition/NonNull.php
```diff
@@ -4,12 +4,22 @@
 
 namespace GraphQL\Type\Definition;
 
+use GraphQL\Error\InvariantViolation;
+use GraphQL\Type\Schema;
+use function assert;
```
simPod, line 9: I don't see this used in the file

src/Type/Definition/ListOfType.php
```diff
@@ -4,24 +4,38 @@
...
     public $ofType;
 
-    public function __construct(Type $type)
+    /**
+     * @param callable():Type|Type $type
+     */
+    public function __construct($type)
     {
-        $this->ofType = $type;
+        $this->ofType = is_callable($type) ? $type : Type::assertType($type);
```
spawnia, line 20:
```suggestion
        $this->ofType = is_callable($type)
            ? $type
            : Type::assertType($type);
```

This is a bit of a pet peeve of mine, i find one-liner ternary expressions wildly unreadable.

src/Type/Definition/Type.php
spawnia: It seems to be a leftover from `graphql-js`, in JS you really can't ever be sure. Agreed that those assertions aren't really needed as PHP itself has runtime typechecks in function signatures.

src/Type/Schema.php
```diff
@@ -339,15 +341,18 @@ private function loadType(string $typeName) : ?Type
 
         $type = $typeLoader($typeName);
 
-        if (! $type instanceof Type) {
+        if (! $type instanceof Type && ! is_callable($type)) {
```
spawnia, line 344: How about we first check if a `callable` is given, call it, then assert that either `$type` or the result of calling the `callable` is a valid `Type`?

src/Type/Definition/ListOfType.php
shmax: If you feel that strongly about it then it should be enforced by the linter. Since it's not, and since I personally prefer the one-liner, I'd rather not descend into subjective style nits, if it's all the same to you.

mfn: FWIW I'm with @spawnia 

Too easy to miss the else part when reading the code.

shmax: I'm not going to get into some popularity vote over this. You are free to open your own PR after the fact and do whatever kind of style gymnastics you want and I wont' interfere, but I have enough trouble with this PR without bikeshedding.

spawnia: `php-cs-fixer` doesn't seem to have a configuration rule for it. Would be nice to automate for sure.

src/Type/Definition/ListOfType.php
shmax: True enough! a8f870d

src/Type/Definition/InputObjectField.php
shmax: Unfortunately, it's not that simple. Before this PR, you were able to get away with this annotation, because `$this->type` received its write from an anonymous, [un-typed array iteration](https://github.com/webonyx/graphql-php/blob/master/src/Type/Definition/InputObjectField.php#L47), so it didn't argue with you about the type, and it would take your word for it on the return type, here. Now that I am returning the result of `Schema::resolveType`, the types must agree. If you can suggest a simple fix I'm all ears, but my long-term plan was to re-introduce native asserts in a follow-up PR.

src/Type/Definition/FieldDefinition.php
shmax: As discussed [previously](https://github.com/webonyx/graphql-php/pull/557#discussion_r437488563), nothing I can do here without asserts.

src/Type/Definition/InputObjectField.php
simPod: I definitely prefer `assert()` rather than discarding type annotation and losing type coverage.

src/Type/Schema.php
shmax: > This seems it can still be null

Where do you see it receiving null?

> Also the UnionType is missing in the getter method return type

👀 e73353c

src/Type/Definition/NonNull.php
shmax: Right you are. d214aa5

src/Type/Schema.php
simPod:
```php
    public function testNull()
    {
        $schema = new Schema([]);
        self::assertNotNull($schema->possibleTypeMap);
    }
```

this fails because `Schema::$possibleTypeMap` can be null but it's not listed in `@var` annotation.

src/Type/Schema.php
shmax: Sure, if you prefer. 07b9297

src/Type/Definition/InputObjectField.php
shmax: Great, I agree, but let me get into that in a follow-up PR and I'll really do it justice all across the board.

src/Type/Schema.php
shmax: `possibleTypeMap` is private, is it not?

simPod: Yes. This was just an example to show the issue.

src/Type/Definition/InputObjectField.php
simPod: If you promise to return the types... 😁

shmax: Scout's honor! 🤞

src/Type/Schema.php
shmax: But it doesn't show the issue. It's private, so your assert wouldn't even get a chance to fire.

src/Type/Definition/FieldArgument.php
shmax: My next PR will be a full native assert overhaul across the board. You can consider that the price of merging this PR.

src/Type/Schema.php
simPod: But it's still null for class members after instantiation. So there's no way to eg. statically analyses that the value is set. It's bug of this kind https://3v4l.org/jMf4f

shmax: > But it's still null for class members after instantiation. 

By that logic, literally all class members that are not initialized at declaration time should have `null` as part of their type, and I don't think you want that. I should also mention that as of PHP 7.4, if you try to access a typehinted member before it has been initialized you will get an exception of the form "must not be accessed before initializtion"; in other words, the default value is not `null`, it is simply undefined.

src/Type/Schema.php
spawnia: It's an extra level of safety. Thanks

src/Type/Schema.php
spawnia: I think lazily initialized properties such as this one should be annotated as nullable, consider https://github.com/webonyx/graphql-php/blob/f6839dd6559bc0f891fdb1af1e1bc9c00ad32aae/src/Type/Schema.php#L409

When we switch to PHP 7.4 and use typed properties, we will have to choose one of those variants:

```php
private int $y;

if (! isset($this->y))
```

```php
private ?int $y;

if ($this->y === null)
```

I prefer the first one, because it disallows `$this->y = null`. So to summarize, i think the type annotation is good, and we might switch the check in the lazy access to use `isset()`.

src/Type/Definition/ListOfType.php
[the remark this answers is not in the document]
spawnia: This is fixed as of [https://github.com/webonyx/graphql-php/pull/557/commits/07b92977552db16eccdfb875b6c59dc0c7868271 ](https://github.com/webonyx/graphql-php/pull/557/commits/07b92977552db16eccdfb875b6c59dc0c7868271)

src/Type/Definition/FieldArgument.php
spawnia: Why do you need to remove this, though? As a user of this library + PHPStan, i would appreciate the precise type, event if there is a slim chance that it might be wrong (not sure how that might even happen).

Adding an `assert()` later seems like a great idea, too - i think we can have both.

src/Type/Schema.php
shmax: > So to summarize, i think the type annotation is good, and we might switch the check in the lazy access to use isset().

Are you arguing for or against adding the `null` type? A shift towards `isset` seems to be in favor of not using it...

src/Type/Definition/InputObjectField.php
spawnia: PHP will not allow us to achieve full type-soundness all the way through. That privilege is reserved for other languages like Haskell 😉

In some cases, we should just be pragmatic and accept that a malicious or careless user *might* cause a type annotation to lie, but still offer the precise type hint for the 99% case.

src/Type/Schema.php
spawnia: Non-null, as it sits in your PR 👍

src/Type/Definition/InputObjectField.php
shmax: > Typed property GraphQL\Type\Schema::$possibleTypeMap must not be accessed before initialization

I appreciate your input, but I think we can do a little better than just jamming the wrong annotations on there--after all, the annotations we have now _are_ correct for the moment--and ignoring the static checker. If you really, really just want to see that `Type&InputType` up there as some form of documentation, then I can do a stan-style cast. But really, I'd rather just save my energy for a proper refactor in a follow-up PR, so let me hear it from @simPod before I do anything else, here...

src/Type/Definition/FieldArgument.php
shmax: > Why do you need to remove this, though?

Because if I restore it, it conflicts with the known return type of `Schema::resolveType`, which is `Type`. I believe the best way to make everybody happy is going to be with native asserts, but setting those up properly will involve more work and more discussion, and I really think it warrants a dedicated PR.

src/Type/Definition/ResolveInfo.php
shmax: It gets a write from `$fieldDefinition->getType()`, which resolves to `Type`. I will handle this more gracefully in my follow-up PR.

src/Type/Definition/Type.php
shmax: It's only being used in one place; I will nuke it in my next PR.

src/Type/Definition/UnionType.php
shmax: That's a good point! I wonder what I was thinking...

shmax: Simplified in 88db0b3, but I still don't remember what led me to do this 🤔

src/Type/Definition/FieldArgument.php
simPod: TBH the type doc can be IMO kept as is and the type incompatibility within this method ignored until the next PR

src/Type/Schema.php
simPod: > By that logic, literally all class members that are not initialised at declaration time should have null as part of their type

exactly

I created wrong example before, this is the bug we're introducing here https://3v4l.org/5Fvv3

We have not yet dropped php 7.1, we're not dropping 7.3 anytime soon that said for following few years there won't be typed properties in this lib.

Also any type checker will complain about this code `if ($this->possibleTypeMap === null) {`  because we're wrongly introducing statement that it can never be `null`. I must admit that I really can't understand while we all know the `null` is property's possible type why we're discussing whether we document it or not. Especially considering the fact it was correct before.

spawnia: That conflates *declaration* with *assignment*. We should change the check to `if (! isset($this->possibleTypeMap)) {`.

There is value in declaring properties as non-null: Assigning `null` is not allowed. We already have a way of knowing if a property is always initialized (if it is assigned a value at declaration).

Those semantics don't really change from PHP 7.1 to PHP 7.4, they are just reiffied by the actual property types.

> By that logic, literally all class members that are not initialised at declaration time should have null as part of their type

simPod: But this is not about the actual check. That's about introducing bug within the class scope.

1) Schema can be instantiated with invalid state.
2) Whenever working within the class with `$this->possibleTypeMap` there's higher risk of breaking something because `null` declaration is missing so we can't leverage type analysis to catch accessing the empty value as shown in my 3v4l.
3) This https://3v4l.org/Ospoj is ofc nice but we will won't have typed properties any time in the near future so there's no point writing the code like we have them already.

We will have this for few years
```php
    /** @var int */
    private $y;
```

Not this

```php
    /** @var int|null */
    private $y;
```

Therefore this is unsafe and bug https://phpstan.org/r/21100d5a-c0d0-4e74-8eba-884961ab7d29 while this is still bug https://phpstan.org/r/58a50994-2ba0-405c-8ed5-5c7c57bd5470 but safely caught.

When we drop PHP 7.3 support, we can start working with typed properties.

spawnia: > we can't leverage type analysis to catch accessing the empty value

That is a limitation of PHPStan, see https://github.com/phpstan/phpstan-strict-rules/issues/88

It is unfortunate the tools don't support it, but the semantics of this definition are quite clear to me:

```php
    /** @var int */
    private $y;
```

It is obvious that `$y` isn't an `int` all the time, but the definition tells me *if* it is initialized, it may only be assigned to an `int`.

Consider the following bug: https://3v4l.org/PoOhI

shmax: > b02d8ce

Actually, I'm a little baffled. I have the commit locally and pushed to this branch, but git is telling me that "Everything up-to-date", and I don't see it here. 😕

src/Type/Definition/NonNull.php
```diff
@@ -4,12 +4,21 @@
...
 class NonNull extends Type implements WrappingType, OutputType, InputType
 {
-    /** @var NullableType&Type */
+    /** @var callable|(NullableType&Type) */
     private $ofType;
 
-    public function __construct(NullableType $type)
+    /**
+     * code sniffer doesn't understand this syntax. Pr with a fix here: waiting on https://github.com/squizlabs/PHP_CodeSniffer/pull/2919
+     * phpcs:disable
```
simPod, line 18: can we disable only incompatible sniff, not all?

shmax: Sure. cfb3d3a

src/Type/Definition/InputObjectField.php
spawnia: There are a bunch of fine solutions, but i would like to keep at least some reference to what this type should be. If you want a technically correct type, you can still do something like:

```php
/**
 * @return Type that should implement InputType
 */
```

src/Type/Definition/ListOfType.php
```diff
@@ -4,24 +4,38 @@
...
+        return '[' . $this->getOfType()->toString() . ']';
+    }
+
+    public function getOfType()
+    {
+        return Schema::resolveType($this->ofType);
     }
 
+    /**
+     * @return ObjectType|InterfaceType|UnionType|ScalarType|InputObjectType|EnumType
```
spawnia, line 34: That type is only accurate if you call this function like `getWrappedType(true)`. If not, the wrapped type can be another wrapping type.

tests/Type/LazyTypeLoaderTest.php
```diff
@@ -0,0 +1,395 @@
...
+
+    /** @var string[] */
+    private $calls;
+
+    /** @var Type[] */
+    private $loadedTypes = [];
+
+    private function lazyLoad(string $name) : callable
+    {
+        return function () use ($name) {
```
spawnia, line 56:
```suggestion
        return function () use ($name): ?Type {
```

src/Type/Definition/InputObjectField.php
shmax: You win. I added a temporary cast. 7aba47f

spawnia: Thank you

src/Type/Definition/ListOfType.php
shmax: Fair enough! 3551138

tests/Type/LazyTypeLoaderTest.php
shmax: Thanks much! 72ccaa3

src/Type/Schema.php
```diff
@@ -339,15 +341,22 @@ private function loadType(string $typeName) : ?Type
...
                 sprintf(
-                    'Type loader is expected to return valid type "%s", but it returned %s',
+                    'Type loader is expected to return a callable or valid type "%s", but it returned %s',
                     $typeName,
                     Utils::printSafe($type)
                 )
             );
         }
+
+        $type = self::resolveType($type);
```
spawnia, line 358: No longer needed.

src/Type/Schema.php
```diff
@@ -339,15 +341,22 @@ private function loadType(string $typeName) : ?Type
 
         $type = $typeLoader($typeName);
 
+        if (is_callable($type)) {
+            $type = self::resolveType($type);
```
spawnia, line 345:
```suggestion
            $type = $type();
```

src/Type/Schema.php
```diff
@@ -359,12 +368,24 @@ private function loadType(string $typeName) : ?Type
...
+        // Default type loader simply falls back to collecting all types
         $typeMap = $this->getTypeMap();
 
         return $typeMap[$typeName] ?? null;
     }
 
+    /**
+     * @param Type|callable():Type $type
+     */
+    public static function resolveType($type) : Type
```
spawnia, line 380: I think this function can be removed, as this check and conversion happens above. The crafted error is nicer than having this fail at the function boundary.

tests/Type/ValidationTest.php
spawnia: I think this is fine to leave out for now. By using stacktraces, users should still be able to figure out the error just fine. `assert()` will improve this, too :+1:

src/Type/Schema.php
spawnia: Ah, just saw this is used in a bunch of places. I am fine with keeping it, but we should avoid double-checking for `is_callable()` in this class.

shmax: Why would I remove this? It provides a hook to narrow the type, encapsulates the actual work, and in any case is currently called 27 times in the code base. 

I'm doing my best to hang in there, but this PR has been open since October, and I am rapidly reaching the point where my main application can't progress without this feature. Any chance we can save major refactoring suggestions like this for some other day?

src/Type/Schema.php
```diff
@@ -339,15 +341,22 @@ private function loadType(string $typeName) : ?Type
 
         $type = $typeLoader($typeName);
 
+        if (is_callable($type)) {
+            $type = self::resolveType($type);
+        }
+
         if (! $type instanceof Type) {
```
spawnia, line 348: We can never reach this point if we go through the type-safe `resolveType()` before. I think we can remove this typecheck, it seems to be a leftover from `graphql-js`. We can just let PHP do the type checking for us.

src/Type/Schema.php
spawnia: Nevermind, i think we can remove the `is_callable()` check here.

src/Type/Schema.php
shmax: True! Thank you... 375fd91

src/Type/Schema.php
shmax: >  if we go through the type-safe resolveType()

You don't go through it if the type is not callable.

src/Type/Schema.php
spawnia: I was wrong with my first comment, this function should stay. There is redundant code in this class that should go instead, see https://github.com/webonyx/graphql-php/pull/557#discussion_r439626066

src/Type/Schema.php
shmax: >  we can remove the is_callable() check her

I would love that, and if this were my personal code base it would already be done, but we have a lot of (sort of bizarre) tests that want to demonstrate what happens when bad types are passed through the system, and making that change would cause them to fail and open up a huge can of worms.

Not to sound like a broken record, but in my next PR I'm going to overhaul all of this, and while I'm at it I'll see if I can get away with killing those tests.

src/Type/Schema.php
spawnia, line 348:
```suggestion
        $type = self::resolveType($type);
```

That line is all we need. No additional checks for `is_callable()`, no checks if `$type` is indeed a type.

spawnia: https://github.com/webonyx/graphql-php/pull/557/files/72ccaa301778869cafbd35d2ce6746b21872b4cf#r439629614

src/Type/Schema.php
spawnia: https://github.com/webonyx/graphql-php/pull/557/files/72ccaa301778869cafbd35d2ce6746b21872b4cf#r439629614

src/Type/Schema.php
shmax: Can't do it without breaking a lot of tests. 😢 Will revisit in next PR.

shmax: ?

spawnia: GitHub won't let me PR from this repo to your branch, but here is how i propose to fix this: https://github.com/webonyx/graphql-php/commit/a705ad46c83fdcc866448d707adb92f45527ea0a

shmax: Sure, looks lovely, but the last time I messed with the tests I got push-back from @vladar . So if it's okay with him, it's okay with me, but I don't want to do anything until I hear it from him.

src/Type/Schema.php
spawnia: I understand what you mean @vladar

There are two levels of lazy loading going on here:

1. Calling `$typeLoader($typeName)`
2. Checking if the result of that is yet another callable and calling that again

Why would 2. be necessary? As you noted, the two calls happen right after each other.

src/Type/Schema.php
spawnia: Most importantly, we should be consistent and always throw either a `TypeError` or a crafted `InvariantViolation`. As of now, this PR does a mix of both, depending on whether the type was lazy or not.

My opinion is that we can let the errors happen right at PHP's type boundaries. In contrast to JavaScript, PHP gives us those checks essentially for free and includes a readable error message and a stacktrace. Since those type errors work the same everywhere, they should be quite easy to understand.

shmax: I'm well aware of both flavors. My goal is to get this PR merged with minimal impact to the existing tests, because once we start removing all the "invariant" stuff then we start changing the functionality of the current "validate" functionality, and before you know it I'll have lost another 3 months to this PR. For the love of all that is holy, please just let it rest for right now and I swear I will revisit all of this stuff in my next PR and we can all have a great time arguing about this stuff then.

spawnia:
```suggestion
        if (is_callable($type)) {
            $type = $type();
        }

        if (! $type instanceof Type) {
```

Let's go with this then. That way, we throw a consistent error.

shmax: If it will satisfy you, then fine: 6a1fad4

src/Type/Schema.php
shmax: Is this what you guys had in mind? 484d2dd

I dunno, I guess I'm okay with it, but I kind of like the idea that executing a lazy type is entirely the library's responsibility, and all the client has to do is deliver a callable or a type, but I suppose I can go either way. Let me know what you think.

spawnia: @shmax thanks for showing the impact of changing the approach in that commit.

I can see how it can more convenient to implement if the type loading is lazy in all circumstances. It allows using the same lazy type reference in any place that expects a type, including the type loader.

The performance impact should be pretty minimal, there is one extra call to check `is_callable()` for every type loaded during a request. I think the extra convenience in development and the simplicity are worth it.

@vladar can you make a final call?

shmax:
@vladar I think we're ready for another review when you're ready.

src/Type/Schema.php
simPod: BTW I think this is not valid even with php 7.4+. If property is null after construction, it should be marked as nullable. We won't be able to fully leverage static analysis over this if we introduce this bug.

spawnia: It is not `null` though, it is uninitialized. It just so happens that untyped properties return `null` when accessed, whereas typed properties throw.

simPod: Can you share such a snippet that is valid with psalm?

shmax: > We won't be able to fully leverage static analysis over this if we introduce this bug.

What bug? Semantically, Uninitialized variables are not `null`, they are undefined. Yes, for the moment, in PHP they take the `null` value, but this is an implementation detail that can change (as we will see in PHP 7.4). It doesn't matter what happens in the constructor, it doesn't matter when they get their first write, and it doesn't even matter that they are members (the same principle holds for locally declared variables); all that matters is that you should not try to access it until it has been initialized. True, PHPStan does not currently catch usage of uninitialized variables, but that's a limitation, not a feature.

I feel like we're going in circles, here. What is it going to take to convince you?

simPod: The thing is that it is a feature https://github.com/vimeo/psalm/issues/2650

https://psalm.dev/articles/psalm-supports-php-7-4#warn-about-properties-not-initialised-within-constructors

shmax: I don't understand; what impact does that have on this discussion? The sample you shared doesn't seem to have anything to do with nullable types.

simPod: We're doing this https://psalm.dev/r/5c9d1dcdd0. This will always emit errors.

shmax: This is _enforcement_ of not accessing uninitialized varaibles. It's great, and I wish PHPStan did it, too (no doubt it will, someday), but it does not change or conflict with what @spawnia and I are trying to tell you.

simPod: I guess I still don't see how do we make the code analysable without ignoring such errors while maintaining this suggested style of accessing variables.

shmax: We don't currently have any errors. We are correctly using `isset` to ensure that we don't access a member that we know can be undefined. The code is happy. The functionality is happy. The static checker is happy. The only small wrinkle is that we're missing a nice-to-have, which is the ability for our checker to catch usage of uninitialized variables, but even if the checker never gets around to it, it will happen natively for us all in PHP 7.4.

simPod: So as we target to max level of static analysis, we will have to ignore errors reported by code analysis in the future to bypass code style like this. And that will result in rendering the code less error prone because of unnecessary ignores in our checks.

I'm not saying it does not work. I'm saying this will lower our code quality as we will have to disable our future checks. Static checker is happy because most of it's functionality is disabled...

shmax: > we will have to ignore errors reported by code analysis in the future to bypass code style like this

No. The checker will not produce any errors here, even at max level. The worst you can say about it is that it will fail to catch usage of uninitialized variables. That's regrettable, but it will either appear in time, or eventually be mooted by PHP 7.4. It's not just cause to run around through our code making all of our members nullable, which absolutely WILL weaken our code (by inviting null writes that are not appropriate).

I can't spend the rest of my life arguing about this. Take one more day to sleep on it, and if you're still not convinced I'll change this one instance just to get us moving again, and we'll resume the fight some other day.

spawnia: Psalm won't even allow nullable typed properties to be uninitialized after construction: https://psalm.dev/r/a982c0ab46. It's not like this approach is foolproof: https://psalm.dev/r/3342e252f2 fails https://3v4l.org/jBLFS

Lazy property initialization is useful for performance. I also want to be able to use typed properties once they are available, and to write idiomatic code. Changing the type hint to allow something wrong (assigning `null`) to work around the limitations of tools we are using seems silly.

A great benefit of non-null types is that once we know the property was initialized, we know it's value is good. When we make every property nullable, our code is going to be littered with null-checks.

Hopefully, the tooling will grow to catch more real errors related to possibly uninitialized properties. Until then, we can ignore rules that don't help us to achieve the result we want. Luckily for us, it's not like there is some kind of type police going around that is going to beat us for it 👮‍♀️

shmax: > When we make every property nullable, our code is going to be littered with null-checks.

Thank you 👍

shmax: > Psalm won't even allow nullable typed properties to be uninitialized after construction

That strikes me as a kind of hammy check--there is no requirement in any other statically typed language I know of to initialize your members in the constructor. Who says that's appropriate in every case? I suspect they did it this way because it was simpler than tracking the state of the variable through the entire tree during analysis.

simPod: So as I said it basically means we will have to suppress errors from our static analysis tooling. I tried to avoid that to keep code quality automated as much as possible.

shmax: > So as I said it basically means we will have to suppress errors from our static analysis tooling. 

What errors?

simPod: I linked it before, this error case

https://psalm.dev/r/5c9d1dcdd0  
https://github.com/vimeo/psalm/issues/2650

shmax: Yes, that is psalm. They took a very conservative approach to checking for uninitialized members (the hammy constructor-time check we've already discussed). PHPStan doesn't do it that way, and it never will. We don't currently have any static checking errors around our current approach, and we have no reason to expect any in the future.

shmax:
@vladar Yee-haw! Thanks for reviewing. I'm available to do touch-ups, and I'll get started on the assert stuff I've been yapping about. Did you have any final input on https://github.com/webonyx/graphql-php/pull/557#discussion_r439725210 ?

vladar:
Yeah, I see your point of doing this now. I guess it makes sense to restore the original version allowing callbacks. Mind creating another follow-up PR for this?

src/Type/Schema.php
vladar: Yeah, I see how this improves DX, let's restore it in a follow-up PR!

shmax:
Sure, will work on it this weekend. Thanks for your support.

unclecheese:
Love this. Thank you, everyone. Can someone tell me, is there a performance benefit to storing the type definitions in separate files and adding a `require_once()` call to the cache fetch? Seems like if your query only needs one or two types, it would save a lot of parsing time.

shmax:
> Love this. Thank you, everyone. Can someone tell me, is there a performance benefit to storing the type definitions in separate files and adding a require_once() call to the cache fetch? Seems like if your query only needs one or two types, it would save a lot of parsing time.

That's an interesting idea--I haven't gotten that far, myself, but I love the idea. As our first lazy load test pilot, why don't you give it a try and let us know?

unclecheese:
Yup, will do. My schema is 28,000 lines of generated PHP, so if it has an effect, I'm sure I'll see it. :)

shmax:
> save a lot of parsing time.

By the way, are you not using opcache?

unclecheese:
I haven't gotten that far yet. Clearly that would make a big difference, but I'm just hacking away on a Vagrant box ATM, and I'm not sure what it has configured.

shmax:
Well, it would probably be easier than refactoring your code, and it would benefit everything. All the cool kids are doing it...

PR #558, opened 2019-10-18
Lock PHPStan version

simPod:
PHPStan does not have typical semver versioning and often introduces new checks in "patch" versions (those are not considered patch by author).

This change prevents checks in CI/clean composer installs to fail when new PHPStan version is released (same goal was attempted to be achieved here https://github.com/webonyx/graphql-php/pull/519)


mfn:
Is there any intention by phpstan to go semvar?

At this point it's not clear yet, it's still a pre-1 version thus semver wouldn't apply either way (at least, that's my understanding).

simPod:
@mfn there was akward discussion on the topic https://github.com/phpstan/phpstan/issues/2163 

There are BC breaks in "patch" versions. I don't complain about it, yet we should be locked against it.

mfn:
Thanks for the link, appreciated!

vladar:
Nice, thank you!

PR #566, opened 2019-10-27
Phpstan Level 3

shmax:
### PHPStan Level 3!

This is already a really tight, beautifully put together code base, so there wasn't that much to do, but there were a few points of interest:

* I wrote PHPStan extensions to direct inferred type for some static methods, namely `isInputType`, `isOutputType`, and `isCompositeType`
* It turns out I was able to bring in native PHP type hinting in a lot of places, so a few of the tests wound up in the position of needlessly validating stuff that would get caught by PHP's natural checking, so I removed them


src/Type/Definition/FieldArgument.php
```diff
@@ -28,14 +28,17 @@ class FieldArgument
     /** @var mixed[] */
     public $config;
 
-    /** @var InputType */
+    /** @var InputType&Type */
     private $type;
 
-    /** @var bool */
-    private $defaultValueExists = false;
```
shmax: I removed this because you can do the same logic by checking the config bucket, but there was and still is a potential bug, here--`defaultValue` is public with no setter, so if no default value is passed at construction time but is set after the fact, then `defaultValueExists` will still return `false`.

tests/Type/ValidationTest.php
```diff
@@ -1308,20 +1314,6 @@ private function schemaWithObjectFieldOfType($fieldType) : Schema
         ]);
     }
 
-    /**
-     * @see it('rejects an empty Object field type')
-     */
-    public function testRejectsAnEmptyObjectFieldType() : void
```
shmax: No longer needed! PHP itself will throw an error if type is not provided and not `Type`.

tests/Type/ValidationTest.php
```diff
@@ -1340,21 +1332,6 @@ public function testRejectsANonOutputTypeAsAnObjectFieldType() : void
         }
     }
 
-    /**
-     * @see it('rejects a non-type value as an Object field type')
-     */
-    public function testRejectsANonTypeValueAsAnObjectFieldType()
-    {
-        $schema = $this->schemaWithObjectFieldOfType($this->Number);
```
shmax: No longer necessary. PHP will throw.

tests/Type/ValidationTest.php
```diff
@@ -1669,21 +1646,6 @@ private function schemaWithInterfaceFieldOfType($fieldType)
         ]);
     }
 
-    /**
-     * @see it('rejects an empty Interface field type')
-     */
-    public function testRejectsAnEmptyInterfaceFieldType() : void
-    {
-        $schema = $this->schemaWithInterfaceFieldOfType(null);
```
shmax: Not needed.

tests/Utils/BreakingChangesFinderTest.php
```diff
@@ -1379,7 +1379,10 @@ public function testShouldDetectIfADirectiveArgumentWasRemoved() : void
                     'name'      => 'DirectiveWithArg',
                     'locations' => [DirectiveLocation::FIELD_DEFINITION],
                     'args'      => FieldArgument::createMap([
-                        'arg1' => ['name' => 'arg1'],
+                        'arg1' => [
+                            'name' => 'arg1',
+                            'type' => Type::string()
```
shmax: This missing type property was caught by PHP's natural checking.

mfn:
> Multiline array shape syntax

🤔 wow, #TIL 

When I look through the PR, 99% (just a feeling) of what I saw regarding the type changes (ecept the multiline array) looks great.

Any reason this would need to be part of this "waiting" PR?

Looks to me the majority of the changes (i.e. "no brainer changes") could be made independently already, no?

shmax:
> Looks to me the majority of the changes (i.e. "no brainer changes") could be made independently already, no?

That's a good point. I'll see if I can separate that stuff out tomorrow morning...

mfn:
> Multiline array shape syntax

Btw, its there a "standard" for this? Does any IDE out there support this? Quickly tried PhpStorm, doesn't seem so.

Neither the referenced PR nor the [issue](https://github.com/phpstan/phpstan/issues/2497) seem to reference the origin of this format.

Thanks

shmax:
> Btw, its there a "standard" for this? Does any IDE out there support this? Quickly tried PhpStorm, doesn't seem so.

I don't know. The [original PR](https://github.com/phpstan/phpdoc-parser/pull/30) that brought it into phpstan/phpdoc-parser didn't mention anything; best I could find is [this issue](https://github.com/phpstan/phpdoc-parser/issues/21), which mentions a similar feature in psalm's docblocks.

Digging further, I found [this](https://github.com/vimeo/psalm/blob/fafb7ee7e17a670302b877c1286d6aafe8d51a72/docs/annotating_code/type_syntax/array_types.md#object-like-arrays), which describes the feature as a "special format for arrays". The PHPDoc proposed [PSR-5 standard](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#arrays) doesn't have this feature.

Phan also has [something similar](https://github.com/phan/phan/blob/e4457bfb0dcf269147a8fc9e33015958f7a4306f/src/Phan/Language/Type/ArrayShapeType.php), though I didn't bother digging down to find its source (interestingly, phan seems to [credit the idea](https://github.com/phan/phan/blob/e4457bfb0dcf269147a8fc9e33015958f7a4306f/src/Phan/Language/Type/ArrayShapeType.php#L23) to phpdoc).

I guess we'll have to consider this sort of a graphql/phan/psalm-only feature.

simPod:
I think it was first introduced in psalm https://psalm.dev/docs/annotating_code/type_syntax/array_types/#object-like-arrays. 

PHPStorm is kinda slow in implementing this, I think I even created ticket for them. But eg. ticket for generic collection support is there in top voted tickets for ages already.

mfn:
> PHPStorm is kinda slow in implementing this

Interesting. When PSR-12 was about to be ratified, I couldn't change my shirt faster than they started implementing it.

Guess having something "official" can boost this.

Sorry for OT and @shmax thanks for digging deeper, obviously I did a bad job at researching 👍

shmax:
Removed the array shape stuff for now. Will open a new PR if this gets merged.

src/Executor/ReferenceExecutor.php
```diff
@@ -995,7 +995,7 @@ private function completeAbstractValue(AbstractType $returnType, $fieldNodes, Re
      * @param mixed|null              $contextValue
      * @param InterfaceType|UnionType $abstractType
      *
-     * @return ObjectType|Promise|null
+     * @return string|Type|Promise|null
```
simPod:
```suggestion
     * @return Promise|Type|string|null
```

src/Experimental/Executor/Collector.php
```diff
@@ -48,7 +48,7 @@ class Collector
     /** @var FieldNode[][] */
     private $fields;
 
-    /** @var string[] */
+    /** @var array<string,bool> */
```
simPod:
```suggestion
    /** @var array<string, bool> */
```

src/Experimental/Executor/CoroutineExecutor.php
```diff
@@ -73,22 +74,22 @@ class CoroutineExecutor implements Runtime, ExecutorImplementation
     /** @var string|null */
     private $operationName;
 
-    /** @var Collector */
+    /** @var ?Collector */
```
simPod: Not sure how `?` is supported in @var annotations, `|null` is afaik more standard way.

src/GraphQL.php
```diff
@@ -287,7 +288,7 @@ public static function getStandardTypes() : array
      * Replaces standard types with types from this list (matching by name)
      * Standard types not listed here remain untouched.
      *
-     * @param Type[] $types
+     * @param array<string,ScalarType> $types
```
simPod:
```suggestion
     * @param array<string, ScalarType> $types
```

src/Language/DirectiveLocation.php
```diff
@@ -56,10 +56,8 @@ class DirectiveLocation
 
     /**
      * @param string $name
-     *
-     * @return bool
      */
-    public static function has($name)
+    public static function has($name) : bool
```
simPod: you can also typehint $name and drop the doc block completely

src/PhpStan/Type/Definition/Type/IsCompositeTypeStaticMethodTypeSpecifyingExtension.php
```diff
@@ -0,0 +1,44 @@
...
+use PHPStan\Analyser\Scope;
+use PHPStan\Analyser\SpecifiedTypes;
+use PHPStan\Analyser\TypeSpecifier;
+use PHPStan\Analyser\TypeSpecifierAwareExtension;
+use PHPStan\Analyser\TypeSpecifierContext;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\StaticMethodTypeSpecifyingExtension;
+
+class IsCompositeTypeStaticMethodTypeSpecifyingExtension implements StaticMethodTypeSpecifyingExtension, TypeSpecifierAwareExtension
```
simPod:
```suggestion
final class IsCompositeTypeStaticMethodTypeSpecifyingExtension implements StaticMethodTypeSpecifyingExtension, TypeSpecifierAwareExtension
```

same for others

also, does this have to be in src? shouldn't we move it to tests?

src/Server/Helper.php
```diff
@@ -146,7 +146,7 @@ public function parseRequestParams($method, array $bodyParams, array $queryParam
      * Checks validity of OperationParams extracted from HTTP request and returns an array of errors
      * if params are invalid (or empty array when params are valid)
      *
-     * @return Error[]
+     * @return array<int,RequestError>
```
simPod:
```suggestion
     * @return array<int, RequestError>
```

src/Type/Definition/IDType.php
```diff
@@ -66,14 +64,13 @@ public function parseValue($value)
     }
 
     /**
-     * @param Node         $valueNode
      * @param mixed[]|null $variables
      *
-     * @return string|null
+     * @return string
```
simPod: would move to return type

src/Type/Definition/IntType.php
```diff
@@ -99,14 +99,13 @@ public function parseValue($value)
     }
 
     /**
-     * @param Node         $valueNode
      * @param mixed[]|null $variables
      *
-     * @return int|null
+     * @return int
```
simPod: return type

src/Type/Definition/ListOfType.php
```diff
@@ -22,12 +22,7 @@ public function toString() : string
         return '[' . $this->ofType->toString() . ']';
     }
 
-    /**
-     * @param bool $recurse
-     *
-     * @return ObjectType|InterfaceType|UnionType|ScalarType|InputObjectType|EnumType
```
simPod: I'm not sure but in some places this annotation makes sense as Type is too wide. I'll look at it more deeply later.

src/Type/Definition/StringType.php
```diff
@@ -67,14 +67,13 @@ public function parseValue($value)
     }
 
     /**
-     * @param Node         $valueNode
      * @param mixed[]|null $variables
      *
-     * @return string|null
+     * @return string
```
simPod: return type

src/Type/Definition/Type.php
```diff
@@ -33,8 +35,8 @@ abstract class Type implements JsonSerializable
     public const FLOAT   = 'Float';
     public const ID      = 'ID';
 
-    /** @var Type[] */
-    private static $standardTypes;
+    /** @var array<string,ScalarType|CustomScalarType>> */
```
simPod:
```suggestion
    /** @var array<string, ScalarType|CustomScalarType>> */
```

src/Type/Definition/Type.php
```diff
@@ -199,7 +187,7 @@ public static function getInternalTypes()
     }
 
     /**
-     * @param Type[] $types
+     * @param array<string,ScalarType> $types
```
simPod:
```suggestion
     * @param array<string, ScalarType> $types
```

src/Type/Definition/UnionType.php
```diff
@@ -22,13 +22,16 @@ class UnionType extends Type implements AbstractType, OutputType, CompositeType,
     /** @var ObjectType[] */
     private $types;
 
-    /** @var ObjectType[] */
-    private $possibleTypeNames;
+    /** @var array<string, bool>|null */
+    private $possibleTypeNames = null;
```
simPod: null is default value, no need to specify

src/Type/Introspection.php
```diff
@@ -39,7 +39,7 @@ class Introspection
     const TYPE_FIELD_NAME      = '__type';
     const TYPE_NAME_FIELD_NAME = '__typename';
 
-    /** @var Type[] */
+    /** @var array<string,mixed> */
```
simPod:
```suggestion
    /** @var array<string, mixed> */
```

src/Type/Schema.php
```diff
@@ -57,8 +57,8 @@ class Schema
      */
     private $resolvedTypes = [];
 
-    /** @var Type[][]|null */
-    private $possibleTypeMap;
+    /** @var array<string,Type[]> */
```
simPod:
```suggestion
    /** @var array<string, Type[]> */
```

src/Type/Schema.php
```diff
@@ -390,26 +390,25 @@ private function defaultTypeLoader($typeName)
...
     {
         $possibleTypeMap = $this->getPossibleTypeMap();
 
-        return isset($possibleTypeMap[$abstractType->name]) ? array_values($possibleTypeMap[$abstractType->name]) : [];
+        return array_values($possibleTypeMap[$abstractType->name] ?? []);
     }
 
     /**
-     * @return Type[][]
+     * @return array<string,Type[]>
```
simPod:
```suggestion
     * @return array<string, Type[]>
```

src/Type/Schema.php
```diff
@@ -390,26 +390,25 @@ private function defaultTypeLoader($typeName)
...
      */
     private function getPossibleTypeMap()
     {
-        if ($this->possibleTypeMap === null) {
-            $this->possibleTypeMap = [];
+        if (empty($this->possibleTypeMap)) {
```
simPod: empty is not safe

```suggestion
        if (count($this->possibleTypeMap) === 0) {
```

src/Utils/BreakingChangesFinder.php
```diff
@@ -269,9 +269,6 @@ private static function isChangeSafeForObjectOrInterfaceField(
         return false;
     }
 
-    /**
```
simPod: this should imo at least stay

src/Utils/BreakingChangesFinder.php
```diff
@@ -472,8 +469,6 @@ public static function findValuesRemovedFromEnums(
      * breaking or dangerous changes in the newSchema related to arguments
      * (such as removal or change of type of an argument, or a change in an
      * argument's default value).
-     *
-     * @return string[][]
```
simPod: this should imo at least stay

src/Utils/BreakingChangesFinder.php
```diff
@@ -579,9 +574,6 @@ static function ($arg) use ($newTypeFieldArgDef) {
         ];
     }
 
-    /**
-     * @return string[][]
-     */
```
simPod: this should imo at least stay

src/Utils/BreakingChangesFinder.php
```diff
@@ -619,9 +611,6 @@ static function (InterfaceType $interface) use ($oldInterface) : bool {
         return $breakingChanges;
     }
 
-    /**
-     * @return string[][]
-     */
```
simPod: this should imo at least stay

src/Utils/TypeInfo.php
```diff
@@ -340,11 +342,12 @@ static function ($arg) use ($node) {
...
                 $this->argument            = $argDef;
                 $this->defaultValueStack[] = $argDef && $argDef->defaultValueExists() ? $argDef->defaultValue : Utils::undefined();
-                $this->inputTypeStack[]    = Type::isInputType($argType) ? $argType : null;
+                $this->inputTypeStack[]    = $argType;
                 break;
 
             case $node instanceof ListValueNode:
-                $listType = Type::getNullableType($this->getInputType());
+                $type     = $this->getInputType();
+                $listType = $type !== null ? Type::getNullableType($type) : null;
```
simPod: cognitive load can be decreased

```suggestion
                $listType = $type === null ? null : Type::getNullableType($type);
```

src/Validator/Rules/FieldsOnCorrectType.php
```diff
@@ -84,6 +84,9 @@ private function getSuggestedTypeNames(Schema $schema, $type, $fieldName)
             $interfaceUsageCount  = [];
 
             foreach ($schema->getPossibleTypes($type) as $possibleType) {
+                if (! $possibleType instanceof ObjectType) {
+                    continue;
+                }
```
simPod:
```suggestion
                }

```

src/Validator/Rules/UniqueInputFieldNames.php
```diff
@@ -16,10 +17,10 @@
 
 class UniqueInputFieldNames extends ValidationRule
 {
-    /** @var string[] */
+    /** @var array<string,NameNode> */
```
simPod:
```suggestion
    /** @var array<string, NameNode> */
```

src/Validator/Rules/UniqueInputFieldNames.php
```diff
@@ -16,10 +17,10 @@
...
     public $knownNames;
 
-    /** @var string[][] */
+    /** @var array<array<string,NameNode>> */
```
simPod:
```suggestion
    /** @var array<array<string, NameNode>> */
```

src/Experimental/Executor/CoroutineExecutor.php
shmax: I suppose it's a matter of opinion--there are other `?`-style instances in the code base that had nothing to do with me, and it is a bit more concise, but I have no preference, so I'll go with your idea.

src/PhpStan/Type/Definition/Type/IsCompositeTypeStaticMethodTypeSpecifyingExtension.php
shmax: It doesn't really have anything to do with tests, and I think you probably want it to be distributed to consumers of the package so they can use and check the public method in their own code. That said, there may be a better way to handle this...

src/Type/Definition/ListOfType.php
shmax: Let me know if you can figure it out--in the meantime, I did the same basic `Type` conversion on `ofType` and was able to add a typehint and remove an assert, which is nice.

src/Type/Definition/Type.php
shmax: Heh, I'm dumb--I just figured out that I can commit these suggestions automatically. Cool.

src/PhpStan/Type/Definition/Type/IsCompositeTypeStaticMethodTypeSpecifyingExtension.php
simPod: php world has usualy a bit weird standard folder structure -> `src` and `tests`. Compared to java's apache standard it's a bit limiting

But yup, I though it was not supposed to be consumed 🤔 is it?

`final` should be there in any case.

src/Utils/BreakingChangesFinder.php
shmax: Ah, right--forgot to restore this when I undid the array shapes stuff.

src/Type/Definition/Type.php
simPod: Haha, that's why I do it that way, to make ur life easier! :D

shmax:
@simPod thanks very much for the thorough review! I think I addressed all your comments. Ready for the next round...

src/Utils/BreakingChangesFinder.php
shmax: Yep, fixed, thanks

src/Type/Definition/NonNull.php
```diff
@@ -4,65 +4,24 @@
...
-     */
-    public static function assertNullType($type)
-    {
-        Utils::invariant(
-            $type instanceof self,
-            'Expected ' . Utils::printSafe($type) . ' to be a GraphQL Non-Null type.'
-        );
-
-        return $type;
+        /** @var Type&NullableType $nullableType*/
```
shmax: This is a hack to get around a parsing bug in `phpcode_sniffer`--it can't handle the "Type&NullableType" form in the docblock, so I have to "cast" it here to appease `PHPStan`.

shmax:
Sorry for all the extra commits in this PR--I had to keep fighting with it to make Scrutinizer happy. By the way, are you guys still all-in on Scrutinizer? Trying to keep its black-box code analyzer happy (its got plenty of its own bugs) is a serious pain. We could probably configure it to use PHPStan instead of its own checker; I'm willing to investigate the idea if anyone is interested...

src/PhpStan/Type/Definition/Type/IsCompositeTypeStaticMethodTypeSpecifyingExtension.php
shmax: I did a run into a problem--these files were souring the coverage results. They don't execute at run time, so it doesn't really make sense for them to be factored in when computing coverage. For now I took your advice and moved them into test.

Next I did a little research--apparently what people do is create a dedicated package for their phpstan extensions, and if people want to use these extensions then they add that package to their `composer.json` and add it to the `includes` section of their `phpstan.neon`.

simPod: Yea, it uses to be in another repository. Extension installer can be also easily leveraged then.

docs/type-system/scalar-types.md
```diff
@@ -100,7 +100,7 @@ class EmailType extends ScalarType
      * @return string
      * @throws Error
      */
-    public function parseLiteral($valueNode, array $variables = null)
+    public function parseLiteral(Node $valueNode, array $variables = null)
```
simPod:
```suggestion
    public function parseLiteral(Node $valueNode, ?array $variables = null)
```

examples/01-blog/Blog/Type/Scalar/UrlType.php
```diff
@@ -48,7 +48,7 @@ public function parseValue($value)
      * @return null|string
      * @throws Error
      */
-    public function parseLiteral($valueNode, array $variables = null)
+    public function parseLiteral(Node $valueNode, array $variables = null)
```
simPod:
```suggestion
    public function parseLiteral(Node $valueNode, ?array $variables = null)
```

phpstan.neon.dist
```diff
@@ -24,6 +28,20 @@ parameters:
...
+        tags:
+            - phpstan.typeSpecifier.staticMethodTypeSpecifyingExtension
+    -
+        class: GraphQL\Tests\PhpStan\Type\Definition\Type\IsOutputTypeStaticMethodTypeSpecifyingExtension
+        tags:
+            - phpstan.typeSpecifier.staticMethodTypeSpecifyingExtension
+    -
+        class: GraphQL\Tests\PhpStan\Type\Definition\Type\IsCompositeTypeStaticMethodTypeSpecifyingExtension
+        tags:
+            - phpstan.typeSpecifier.staticMethodTypeSpecifyingExtension
```
simPod: missing EOL

src/Language/Token.php
```diff
@@ -81,18 +81,10 @@ class Token
      */
     public $prev;
 
-    /** @var Token */
+    /** @var ?Token */
```
simPod:
```suggestion
    /** @var Token|null */
```

src/Language/Token.php
```diff
@@ -81,18 +81,10 @@ class Token
...
-    /**
-     * @param string     $kind
-     * @param int        $start
-     * @param int        $end
-     * @param int        $line
-     * @param int        $column
-     * @param mixed|null $value
-     */
-    public function __construct($kind, $start, $end, $line, $column, ?Token $previous = null, $value = null)
+    public function __construct(string $kind, int $start, int $end, int $line, int $column, ?Token $previous = null, $value = null)
```
simPod: missing docs for `$value`. should be `mixed` I guess

src/Language/Visitor.php
```diff
@@ -392,8 +392,6 @@ public static function removeNode()
 
     /**
      * @param callable[][] $visitors
-     *
-     * @return callable[][]
```
simPod: should keep

src/Type/Definition/EnumType.php
```diff
@@ -71,12 +71,10 @@ public function getValue($name)
         return $lookup[$name] ?? null;
     }
 
-    /**
-     * @return ArrayObject<string, EnumValueDefinition>
```
simPod: can keep?

src/Type/Definition/EnumType.php
```diff
@@ -71,12 +71,10 @@ public function getValue($name)
...
-     */
-    private function getNameLookup()
+    private function getNameLookup() : ArrayObject
     {
         if (! $this->nameLookup) {
+            /** @var ArrayObject<string, EnumValueDefinition> $lookup */
```
simPod: and this can be maybe drop when return type doc is kept?

src/Type/Definition/FieldDefinition.php
```diff
@@ -30,15 +30,15 @@ class FieldDefinition
      * Callback for resolving field value given parent value.
      * Mutually exclusive with `map`
      *
-     * @var callable
+     * @var ?callable
```
simPod: `?` -> `|null`

src/Type/Definition/FieldDefinition.php
```diff
@@ -30,15 +30,15 @@ class FieldDefinition
...
      */
     public $resolveFn;
 
     /**
      * Callback for mapping list of parent values to list of field values.
      * Mutually exclusive with `resolve`
      *
-     * @var callable
+     * @var ?callable
```
simPod: ? -> |null

src/Type/Definition/IDType.php
```diff
@@ -53,8 +53,6 @@ public function serialize($value)
     /**
      * @param mixed $value
      *
-     * @return string
```
simPod: can move to return type

src/Type/Definition/IDType.php
```diff
@@ -66,14 +64,11 @@ public function parseValue($value)
     }
 
     /**
-     * @param Node         $valueNode
      * @param mixed[]|null $variables
      *
-     * @return string|null
```
simPod: can move to return type

src/Type/Definition/IntType.php
```diff
@@ -99,14 +99,11 @@ public function parseValue($value)
     }
 
     /**
-     * @param Node         $valueNode
      * @param mixed[]|null $variables
      *
-     * @return int|null
```
simPod: can move to return type

src/Type/Definition/ObjectType.php
```diff
@@ -62,7 +62,7 @@ class ObjectType extends Type implements OutputType, CompositeType, NullableType
     /** @var ObjectTypeExtensionNode[] */
     public $extensionASTNodes;
 
-    /** @var callable */
+    /** @var ?callable */
```
simPod: `?` -> `|null`

src/Type/Definition/Type.php
```diff
@@ -55,105 +57,85 @@ abstract class Type implements JsonSerializable
...
-        return self::getStandardType(self::ID);
-    }
-
-    /**
-     * @param string $name
-     *
-     * @return (IDType|StringType|FloatType|IntType|BooleanType)[]|IDType|StringType|FloatType|IntType|BooleanType
-     */
-    private static function getStandardType($name = null)
+    public static function id() : ScalarType
```
simPod: `IDType` should stay here, no?

```suggestion
    public static function id() : IDType
```

src/Type/Definition/Type.php
```diff
@@ -55,105 +57,85 @@ abstract class Type implements JsonSerializable
...
+        return static::$standardTypes[self::ID];
     }
 
     /**
-     * @return StringType
-     *
      * @api
      */
-    public static function string()
+    public static function string() : ScalarType
```
simPod:
```suggestion
    public static function string() : StringType
```

src/Type/Definition/Type.php
```diff
@@ -55,105 +57,85 @@ abstract class Type implements JsonSerializable
...
+        return static::$standardTypes[self::STRING];
     }
 
     /**
-     * @return BooleanType
-     *
      * @api
      */
-    public static function boolean()
+    public static function boolean() : ScalarType
```
simPod:
```suggestion
    public static function boolean() : BooleanType
```

src/Type/Definition/Type.php
```diff
@@ -55,105 +57,85 @@ abstract class Type implements JsonSerializable
...
+        return static::$standardTypes[self::BOOLEAN];
     }
 
     /**
-     * @return IntType
-     *
      * @api
      */
-    public static function int()
+    public static function int() : ScalarType
```
simPod:
```suggestion
    public static function int() : IntType
```

src/Type/Definition/Type.php
```diff
@@ -55,105 +57,85 @@ abstract class Type implements JsonSerializable
...
+        return static::$standardTypes[self::INT];
     }
 
     /**
-     * @return FloatType
-     *
      * @api
      */
-    public static function float()
+    public static function float() : ScalarType
```
simPod:
```suggestion
    public static function float() : FloatType
```

src/Type/Introspection.php
```diff
@@ -39,7 +39,7 @@ class Introspection
     const TYPE_FIELD_NAME      = '__type';
     const TYPE_NAME_FIELD_NAME = '__typename';
 
-    /** @var Type[] */
+    /** @var array<string, mixed> */
```
simPod:
```suggestion
    /** @var array<string, Type> */
```

src/Type/Schema.php
```diff
@@ -340,10 +341,8 @@ public function hasType($name)
 
     /**
      * @param string $typeName
-     *
-     * @return Type
      */
-    private function loadType($typeName)
+    private function loadType($typeName) : ?Type
```
simPod:
```suggestion
    private function loadType(string $typeName) : ?Type
```

src/Type/Schema.php
```diff
@@ -373,10 +372,8 @@ private function loadType($typeName)
 
     /**
      * @param string $typeName
-     *
-     * @return Type
      */
-    private function defaultTypeLoader($typeName)
+    private function defaultTypeLoader($typeName) : ?Type
```
simPod:
```suggestion
    private function defaultTypeLoader(string $typeName) : ?Type
```

src/Utils/TypeInfo.php
```diff
@@ -40,37 +40,39 @@
...
 use function sprintf;
 
 class TypeInfo
 {
     /** @var Schema */
     private $schema;
 
-    /** @var SplStack<OutputType> */
+    /** @var array<(OutputType&Type)|null> */
     private $typeStack;
```
simPod: if it's not stack anymore, can be renamed. `$types`?

src/Utils/TypeInfo.php
```diff
@@ -40,37 +40,39 @@
...
-    /** @var SplStack<FieldDefinition> */
+    /** @var array<FieldDefinition> */
     private $fieldDefStack;
 
-    /** @var SplStack<mixed> */
+    /** @var array<mixed> */
     private $defaultValueStack;
 
-    /** @var Directive */
+    /** @var ?Directive */
```
simPod: `?` -> `|null`

src/Utils/TypeInfo.php
```diff
@@ -40,37 +40,39 @@
...
     private $directive;
 
-    /** @var FieldArgument */
+    /** @var ?FieldArgument */
```
simPod: `?` -> `|null`

src/PhpStan/Type/Definition/Type/IsCompositeTypeStaticMethodTypeSpecifyingExtension.php
shmax: Sounds good. Do you think we can host that here under the `webonyx` space?

tests/Type/ValidationTest.php
```diff
@@ -178,7 +178,9 @@ public function testRejectsTypesWithoutNames() : void
         $this->assertEachCallableThrows(
             [
                 static function () {
-                    return new ObjectType([]);
+                    $empty = [];
+
+                    return new ObjectType($empty);
```
simPod: maybe could stay the way it was

src/Type/Schema.php
```diff
@@ -453,11 +449,9 @@ public function isPossibleType(AbstractType $abstractType, ObjectType $possibleT
      *
      * @param string $name
      *
-     * @return Directive
-     *
      * @api
      */
-    public function getDirective($name)
+    public function getDirective($name) : ?Directive
```
simPod:
```suggestion
    public function getDirective(string $name) : ?Directive
```

src/Type/Definition/NonNull.php
```diff
@@ -4,65 +4,24 @@
 
 namespace GraphQL\Type\Definition;
 
-use GraphQL\Utils\Utils;
-
 class NonNull extends Type implements WrappingType, OutputType, InputType
 {
-    /** @var NullableType */
+    /** @var Type&NullableType */
```
simPod: usually you are using it inversely 
```suggestion
    /** @var NullableType&Type */
```

src/Type/Definition/NonNull.php
```diff
@@ -4,65 +4,24 @@
...
         return $this->getWrappedType()->toString() . '!';
     }
 
-    /**
-     * @param bool $recurse
-     *
-     * @return Type
-     */
-    public function getWrappedType($recurse = false)
+    public function getWrappedType(bool $recurse = false) : Type
```
simPod:
```suggestion
    public function getWrappedType(bool $recurse = false) : NullableType
```

src/Utils/AST.php
```diff
@@ -588,7 +588,7 @@ public static function typeFromAST(Schema $schema, $inputTypeNode)
      *
      * @param string $operationName
      *
-     * @return bool
+     * @return bool|string
```
simPod: can move to return type

src/Type/Definition/NonNull.php
shmax: I don't think I can do that, as it would break the contract with the `WrappingType` interface.

src/Utils/AST.php
shmax: How so? Does PHP support union types?

src/PhpStan/Type/Definition/Type/IsCompositeTypeStaticMethodTypeSpecifyingExtension.php
shmax: Another option would be to just get rid of the `Type::is...` family of methods. Most of them just wrap a vanilla call to `instanceof`, and the more elaborate ones just add a call to `getNamedType`, which is arguably obfuscation.

src/Utils/AST.php
simPod: Ah, my bad. seen `|null`

[but soon!](https://github.com/php/php-rfcs/pull/1)

src/Utils/TypeInfo.php
shmax: Hmm, I dunno--I didn't change the run time type of this in any way, so it's still whatever it used to be. Do you still want me to rename it?

src/Type/Definition/EnumType.php
shmax: I get fresh errors about conflict with `$this->nameLookup`... maybe easier to jusdt keep it for now. Furthermore, I had to remove the annotated return type, or the style checker complains...

src/Language/Token.php
shmax: Sorry, what is it exactly you want me to do?

simPod:
I have to reply via email, github broke Comment input. Hope it will work.



the constructor should have type documentation for `$value`. Should be something like

```php
/**
 * @param mixed $value 
 */
```

src/Language/Token.php
simPod: It's because there's check whether you are missing type declarations or not (currently disabled). And there would be no way to tell whether it's `mixed` or just forgoten. So `mixed` should be declared explicitely.

examples/01-blog/Blog/Type/Scalar/UrlType.php
```diff
@@ -48,7 +48,7 @@ public function parseValue($value)
      * @return null|string
```
simPod: also can move this to return type if we're adding types already :)

src/Error/Error.php
```diff
@@ -330,6 +330,8 @@ public function getExtensions()
      * @deprecated Use FormattedError::createFromException() instead
      *
      * @return mixed[]
+     *
+     * @codeCoverageIgnore
```
simPod: what is the reason for these?

src/Utils/TypeInfo.php
```diff
@@ -363,52 +363,42 @@ static function ($arg) use ($node) {
...
+                    $this->enumValue = $enumType->getValue($node->value);
                 }
                 $this->enumValue = $enumValue;
                 break;
         }
     }
 
     /**
-     * @return Type
+     * @return (Type & OutputType) | null
```
simPod: you don't have spaces elsewhere

```suggestion
     * @return (OutputType&Type)|null
```

src/Utils/TypeInfo.php
```diff
@@ -363,52 +363,42 @@ static function ($arg) use ($node) {
...
-            return $this->typeStack[count($this->typeStack) - 1];
-        }
-
-        return null;
+        return $this->typeStack[count($this->typeStack) - 1] ?? null;
     }
 
     /**
-     * @return Type
+     * @return (Type & CompositeType) | null
```
simPod: you don't have spaces elsewhere

```suggestion
     * @return (CompositeType&Type)|null
```

src/Utils/TypeInfo.php
```diff
@@ -437,57 +427,37 @@ private static function getFieldDefinition(Schema $schema, Type $parentType, Fie
...
-            return $this->defaultValueStack[count($this->defaultValueStack) - 1];
-        }
-
-        return null;
+        return $this->defaultValueStack[count($this->defaultValueStack) - 1] ?? null;
     }
 
     /**
-     * @return ScalarType|EnumType|InputObjectType|ListOfType|NonNull|null
+     * @return (Type & InputType) | null
```
simPod: you don't have spaces elsewhere

```suggestion
     * @return (InputType&Type)|null
```

src/Validator/Rules/KnownArgumentNames.php
```diff
@@ -38,7 +39,7 @@ public function getVisitor(ValidationContext $context)
                 if ($argumentOf instanceof FieldNode) {
                     $fieldDef   = $context->getFieldDef();
                     $parentType = $context->getParentType();
-                    if ($fieldDef && $parentType) {
+                    if ($fieldDef && $parentType instanceof Type) {
```
simPod:
```suggestion
                    if ($fieldDef !== null && $parentType instanceof Type) {
```

src/Utils/TypeInfo.php
shmax: I'll have to take another look at this, but basically the challenge for all this is that I have to satisfy:

1. phpstan/phpdoc-parser
2. code sniffer
3. scrutinizer

In this case,  scrutinizer's static checker is the problem--if I restore the preferred syntax, then it will falsely report incompatibility with the member type. I'll see about opening a bug report with them...

src/Error/Error.php
shmax: Well, they're deprecated and unused, so they are falsely impacting the code coverage score. This was just a convenient way for me to increase the overall average coverage score so that Scrutinizer would pass.

src/Utils/TypeInfo.php
simPod: IMO we can ignore scrutinizer and report problem to them

shmax: Okay. I worked really hard to get Scrutinizer to pass, so let's save this for last and I will report the problem to them first thing after work. If they don't respond and there are no other review comments, I'll change this back.

simPod: Sure there's no point to force it. I'd rather have a sniff for this in the future anyway.

shmax: Absolutely. I'll dig deep into this tonight--there are at least two Scrutinizer bugs I'm aware of, and one in code sniffer. I'll follow up on those, then I'll see if there's an appropriate sniff...

examples/01-blog/Blog/Type/Scalar/UrlType.php
shmax: No can do; PHP will complain about violating the `parseLiteral` signature defined in `LeafScalar`.

src/Type/Schema.php
```diff
@@ -390,26 +387,25 @@ private function defaultTypeLoader($typeName)
      *
      * This operation requires full schema scan. Do not use in production environment.
      *
-     * @param InterfaceType|UnionType $abstractType
+     * @param UnionType|InterfaceType $abstractType
```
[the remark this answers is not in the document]
shmax: No clue! I changed it back.

src/Utils/TypeInfo.php
shmax: I created a dummy repo with a minimal reproduction case of this issue:

https://scrutinizer-ci.com/g/shmax/scrutinizer-bugs/inspections/fea47b85-e320-4ec1-aff5-cb978a0a10bc/code-structure/operation/MyClass%3A%3AgetType

Scrutinizer's github repo is no more, but I reported it through their chat-bot, which is evidently the only way to report issues, these days.

src/Error/Error.php
simPod: Oh I didn't see they're deprecated. In that case I think it's ok. 

But generally Travis should always pass, scrutinizer is nice to have. If coverage decreases it doesn't necessarily be bad/issue. It's just a simple metric anyways.

shmax: And if we're not all that interested in scrutinizer's static checking, I would highly recommend replacing it with PHPStan. Trying to satisfy a checker that is black box, not open source, not available to run locally, and full of bugs is a nightmare.

src/Type/Definition/NonNull.php
shmax: @simPod I forked `PHP_CodeSniffer`(https://github.com/squizlabs/PHP_CodeSniffer) so I can debug this and possibly open a PR with a fix, but I'll be damned if I can figure out how to run tests for it from within PHPStorm (the `contributor.md` recommends running a global `phpunit` command from shell). Any ideas?

simPod: Yup, I have it forked as well

Set test framework (don't forget the PHP interpreter as well)

image

phpunit is dev dependency there, so no need for global.

Then run tests and youre good to go 
image

src/Language/Token.php
simPod: also, null is included in `mixed` afaik, so `mixed` should me enough

src/Experimental/Executor/CoroutineExecutor.php
simPod: Just tested in Jetbrains IDE. It doesn't consider `?` a null on property

src/Type/Definition/NonNull.php
shmax: Ah, I must have missed the config file; I'll try it after work. Great, thanks so much!

src/Type/Definition/Type.php
shmax: I'll have to revisit this; if you change each of these to the obvious relevant type it causes other stuff to break, mostly related to the fact that we allow people to override the built-in types. I'll take another crack at it...

src/Language/Token.php
shmax: removed null

src/Language/Token.php
shmax: done

src/Type/Schema.php
shmax: done

src/Type/Definition/NonNull.php
shmax: I wasn't able to come up with a quick fix for this, but I'm going to keep working on it. In the meantime, hopefully it won't hold up this PR.

src/Type/Definition/EnumType.php
shmax: Unfortunately, not--code sniffer complains about it being "unecessary". Sigh.

src/Type/Definition/IDType.php
shmax: Unfortunately, not; that violates the contract with `LeadScalar`. But I was able to typehint the return of `parseValue`.

src/Type/Definition/IntType.php
shmax: Can't; violates `LeafScalar`. But I added return type hint to `parseValue`.

src/Type/Definition/Type.php
shmax: Can't, unfortunately; users are allowed to replace any of the core built-int types with their own `CustomScalarType`, so without further refactoring of that system there's nothing more I can do at this point.

src/Type/Definition/Type.php
shmax: Don't think there's anything I can do, as this can also be a `CustomScalarType`...

src/Type/Definition/Type.php
shmax: Don't think there's anything I can do, as this can also be a `CustomScalarType`...

src/Type/Definition/Type.php
shmax: Don't think there's anything I can do, as this can also be a `CustomScalarType`...

src/Type/Introspection.php
shmax: Can't do this one, unfortunately--$map also stores `FieldDefinition` (which seems odd)

src/Type/Schema.php
shmax: That works! Thanks...

shmax:
I've addressed all open comments (thanks, @simPod !). At this point all tests and checkers pass (even Scrutinizer). There is still a lot to do, but I think we're at a good stopping point here, and I will resume in PHPStan lvl 4.

Would love some more reviewers! @vladar @spawnia

src/Type/Introspection.php
simPod: What about in the meanwhile then?

```suggestion
    /** @var array<string, FieldDefinition|Type> */
```

src/Utils/TypeInfo.php
shmax: Temporarily necessary because of Scrutinizer. I'll start investigating moving off of their lame static checker in a different branch

simPod:
_If @vladar thinks it's good to go, I'd squash commits before merge._

src/Utils/TypeInfo.php
shmax: Still not a peep from Scrutinizer team.

shmax:
> If @vladar thinks it's good to go, I'd squash commits before merge.
Sure, happy to squash, but doesn't Github have the `Squash and Merge` button?

src/Type/Introspection.php
shmax: Oh, that's a good idea! I'll try that when I get home...

shmax: Done, thanks!

shmax: Had to revert it--Scrutinizer (rightfully) complains. I'll sort this out in more depth in future phpstan levels.

src/Utils/TypeInfo.php
vladar: I would say we never actually bothered about Scrutinizer anyway. Mostly used this for test coverage reports. So feel free to drop its static analysis in favor of PHPStan.

src/Utils/TypeInfo.php
```diff
@@ -339,11 +338,12 @@ static function ($arg) use ($node) {
                 }
                 $this->argument            = $argDef;
                 $this->defaultValueStack[] = $argDef && $argDef->defaultValueExists() ? $argDef->defaultValue : Utils::undefined();
-                $this->inputTypeStack[]    = Type::isInputType($argType) ? $argType : null;
+                $this->inputTypeStack[]    = $argType;
```
vladar: I think this is wrong. The issue is that `isInputType` additionally checks that the whole chain of possibly wrapped types are also input types. Without this check `[ObjectType]` will qualify as an `InputType` (because `ListOf` wrapper is both Output and Input type). We are probably lacking some tests for this.

src/Utils/TypeInfo.php
```diff
@@ -363,52 +363,42 @@ static function ($arg) use ($node) {
                     $inputFieldType = $inputField ? $inputField->getType() : null;
                 }
                 $this->defaultValueStack[] = $inputField && $inputField->defaultValueExists() ? $inputField->defaultValue : Utils::undefined();
-                $this->inputTypeStack[]    = Type::isInputType($inputFieldType) ? $inputFieldType : null;
+                $this->inputTypeStack[]    = $inputFieldType;
```
vladar: Same here

src/Validator/Rules/FieldsOnCorrectType.php
```diff
@@ -84,6 +84,10 @@ private function getSuggestedTypeNames(Schema $schema, $type, $fieldName)
...
```
vladar: `$schema->getPossibleTypes()` is supposed to always return an array of `ObjectType`, so this seems redundant.

tests/Executor/DeferredFieldsTest.php
```diff
@@ -529,14 +530,14 @@ public function testComplexRecursiveDeferredFields() : void
...
-            ['!dfd for: ', ['deferredNest', 'nest', 'deferred']],
-            ['!dfd nest for: ', ['deferredNest', 'deferredNest']],
+            ['!dfd for: ', 'nest', 'deferredNest', 'deferred'],
+            ['!dfd for: ', 'deferredNest', 'deferred'],
+            ['!dfd for: ', 'deferredNest', 'nest', 'deferred'],
+            ['!dfd nest for: ', 'deferredNest', 'deferredNest'],
 
             ['deferredNest', 'deferredNest', 'sync'],
             ['deferredNest', 'deferredNest', 'deferred'],
-            ['!dfd for: ', ['deferredNest', 'deferredNest', 'deferred']],
```
vladar: This whole file change seems wrong. Nesting here is important for sane maintenance. We should bail out of typing tests or maybe just set the type of path as `array<mixed>`. Please revert this.

tests/Type/DefinitionTest.php
```diff
@@ -515,18 +539,6 @@ public function testIdentifiesOutputTypes() : void
...
-     */
-    public function testProhibitsNestingNonNullInsideNonNull() : void
-    {
-        $this->expectException(InvariantViolation::class);
-        $this->expectExceptionMessage(
-            'Expected Int! to be a GraphQL nullable type.'
-        );
-        Type::nonNull(Type::nonNull(Type::int()));
-    }
-
```
vladar: This might still be useful to check (even if we have core exception vs a custom one). Because one day we can change something in the `NonNull` type signature and there won't be any test to fail.

tests/Type/DefinitionTest.php
```diff
@@ -1585,84 +1597,6 @@ public function testDoesNotAllowIsDeprecatedWithoutDeprecationReasonOnEnum() : v
...
-            $typeStr = Utils::printSafe($badType);
-            try {
-                Type::nonNull($badType);
-                self::fail(sprintf('Nulls should not accept %s', $typeStr));
-            } catch (InvariantViolation $e) {
-                self::assertEquals(sprintf('Expected %s to be a GraphQL nullable type.', $typeStr), $e->getMessage());
-            }
-        }
-    }
-
```
vladar: Same for all of these tests, they might be still useful. Or am I missing something?

src/Language/AST/DocumentNode.php
```diff
@@ -9,6 +9,6 @@ class DocumentNode extends Node
     /** @var string */
     public $kind = NodeKind::DOCUMENT;
 
-    /** @var NodeList|DefinitionNode[] */
+    /** @var NodeList|array<DefinitionNode> */
```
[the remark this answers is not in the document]
simPod: @vladar it's in progres in master branch. Sneak peak: https://arnaud.le-blanc.net/post/phpstan-generics.html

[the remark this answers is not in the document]
vladar: Oh that would be awesome

src/Type/Definition/FieldArgument.php
shmax: I don't intend to fight this battle in this PR, but there _are_ ways to have read-only properties. For example:

```
class FieldArgument {
   private $_defaultValue;

   public function __get($name) {
       switch($name) {
          case 'defaultValue`:
             return $_defaultValue;
       } 
   }
}
```

src/Type/Definition/ListOfType.php
shmax: > it returns any Type except instances of WarppingType

Is that true? Couldn't the `ofType` be another `ListOfType`?

src/Utils/TypeInfo.php
shmax: I'm working on it!

src/Executor/ReferenceExecutor.php
```diff
@@ -1023,7 +1023,7 @@ private function defaultTypeResolver($value, $contextValue, ResolveInfo $info, A
         $possibleTypes           = $info->schema->getPossibleTypes($abstractType);
         $promisedIsTypeOfResults = [];
         foreach ($possibleTypes as $index => $type) {
-            $isTypeOfResult = $type->isTypeOf($value, $contextValue, $info);
+            $isTypeOfResult = $type instanceof ObjectType ? $type->isTypeOf($value, $contextValue, $info) : null;
```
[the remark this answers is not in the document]
shmax: Right! I was able to tighten all that up by using `ObjectType`, thanks.

src/Experimental/Executor/CoroutineExecutor.php
```diff
@@ -143,10 +149,11 @@ public static function create(
     private static function resultToArray($value, $emptyObjectAsStdClass = true)
     {
         if ($value instanceof stdClass) {
-            $array = [];
-            foreach ($value as $propertyName => $propertyValue) {
+            $array = (array) $value;
+            foreach ($array as $propertyName => $propertyValue) {
```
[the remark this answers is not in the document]
shmax: > Any specific reason for this change?

PHPStan complains about iterating over `stdClass`. I opened an [issue about it](https://github.com/phpstan/phpstan/issues/2541), but was shot down. I personally feel that this is one case where PHPStan is crossing over into style-checking territory, but I decided not to press the issue at this time. Feel free to comment on the issue.

> may produce another implicit iteration

Not entirely sure what you mean, but I stepped through it a few times and didn't notice anything unusual.

src/Experimental/Executor/CoroutineExecutor.php
```diff
@@ -937,8 +949,8 @@ private function resolveTypeSlow(CoroutineContext $ctx, $value, AbstractType $ab
 
         $selectedType = null;
         foreach ($possibleTypes as $type) {
-            $typeCheck = yield $type->isTypeOf($value, $this->contextValue, $ctx->resolveInfo);
-            if ($selectedType !== null || $typeCheck !== true) {
+            $typeCheck = yield $type instanceof ObjectType ? $type->isTypeOf($value, $this->contextValue, $ctx->resolveInfo) : null;
```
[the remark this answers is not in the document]
shmax: Nice! Fixed in bd650ab

src/Type/Definition/Type.php
```diff
@@ -10,12 +10,14 @@
 use GraphQL\Type\Introspection;
 use GraphQL\Utils\Utils;
 use JsonSerializable;
+use phpDocumentor\Reflection\Types\Nullable;
```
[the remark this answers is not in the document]
shmax: I don't know. Removed this and a few other unused in 0807bb5

src/Type/Definition/Type.php
```diff
@@ -32,8 +34,8 @@ abstract class Type implements JsonSerializable
     public const FLOAT   = 'Float';
     public const ID      = 'ID';
 
-    /** @var Type[] */
-    private static $standardTypes;
+    /** @var array<string, ScalarType|CustomScalarType>> */
```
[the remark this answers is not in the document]
shmax: Good eye! Fixed in e1e5b08

src/Type/Definition/WrappingType.php
```diff
@@ -6,10 +6,5 @@
 
 interface WrappingType
 {
-    /**
-     * @param bool $recurse
-     *
-     * @return ObjectType|InterfaceType|UnionType|ScalarType|InputObjectType|EnumType
-     */
-    public function getWrappedType($recurse = false);
+    public function getWrappedType(bool $recurse = false) : Type;
```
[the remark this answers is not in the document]
shmax: > This method returns any type that is not a WrappingType

Not sure that's true; if `$recurse` is passed in as `false`, then you could get, for example, another `ListOfType` back.

src/Type/Schema.php
```diff
@@ -392,24 +374,23 @@ private function defaultTypeLoader($typeName)
...
 
     /**
-     * @return Type[][]
+     * @return array<string, Type[]>
      */
     private function getPossibleTypeMap()
     {
-        if ($this->possibleTypeMap === null) {
-            $this->possibleTypeMap = [];
+        if (count($this->possibleTypeMap) === 0) {
```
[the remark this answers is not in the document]
shmax: Ah, good point. Restored in bebaad9

src/Utils/TypeInfo.php
shmax: Sounds good. Restored in 3b428f7

src/Utils/TypeInfo.php
shmax: Sure. We can revisit some other time. reverted in 78295d7

src/Validator/Rules/FieldsOnCorrectType.php
shmax: Yes, thank you. Removed in 801033f

tests/Executor/DeferredFieldsTest.php
shmax: Yeah, I must have done that late at night. reverted to what is on master in 18ed4af

tests/Type/DefinitionTest.php
shmax: Do you mind if we revisit that some other time? I'm sure I can modify this test to catch the core exception, but I would rather not dive down that rabbit hole right now if there is not any pressing need for it.

tests/Type/DefinitionTest.php
shmax: `NonNull` constructor now has a `NullableType` type hint in its constructor, so PHP simply won't allow you to get it wrong (unless you configure your error reporting to ignore such errors). I've removed the asserts from the constructor, so these explicit exceptions are not thrown, and there is nothing to catch.

So, at this point, any tests along these lines would not be testing any internal logic in the framework--they would be testing PHP's type hinting, which seems like a waste of resources to me.

src/Type/Definition/WrappingType.php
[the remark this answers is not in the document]
vladar: You are right

tests/Type/DefinitionTest.php
vladar: OK, seems minor anyway

vladar:
Merged! Thanks for this exceptionally useful PR!

shmax:
Woot! Thanks for merging, and thanks all for reviewing!

PR #567, opened 2019-10-31
Only include built-in types that are actually used in Introspection result

spawnia:
graphql-js currently behaves that way, should we adapt to that as well?

I took a look at the latest spec, it does not say anything about that. I think it is not a problem to keep them in, just wanted to bring it up for discussion.


vladar:
I don't remember exactly why but there was a reason for this. Do you have any specific situation where this is important?

spawnia:
I stumbled upon this while porting the tests for `BuildClientSchema` in #539, i figured we could ease maintenance by sticking close to the reference implementation.

vladar:
I am going to close this for now. But if we face real-world issues caused by this, we can re-consider.

PR #569, opened 2019-11-02
Use phpstan instead of scrutinizer's own checker

shmax:
Experimenting with replacing Scrutinizer's proprietary checker with PHPStan.

This doesn't work yet--supposedly any tool that implements the `general-checkstyle` format can replace Scrutinizer's own checks, but while Scrutinizer does parse the new data, it doesn't seem to display it in its report. Please let me know if you have any ideas...

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -20,6 +20,7 @@ build:
                         analysis:
                           file: phpstan-checkstyle.xml
                           format: 'general-checkstyle'
+                    - php-scrutinizer-run
 
 tools:
     external_code_coverage:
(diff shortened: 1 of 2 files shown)


vladar:
Cool, thanks! Will merge after #566 to avoid broken builds %)

simPod:
Also just noticed it seems to be possible to set check as not required.

image

shmax:
Oh, sorry, this isn't finished yet--I only opened a PR because the branches don't seem to run checks...

shmax:
> These reports are often handy

Absolutely; I'm trying to figure out how to generate the reports from our tools instead of Scrutnizer's black box crap.

shmax:
Superseded by https://github.com/webonyx/graphql-php/pull/583 👍

PR #571, opened 2019-11-02
fix phpdoc annotation for GraphQL\Type\Definition\ObjectType::$resolveFieldFn

mavimo:
The field should be nullable (as per definition in constructor) but the phpDoc annotation definition do not report the nullable type.


vladar:
Thanks for the PR! I would have merged it but we just merged a broader scope PR #566 which addressed this.

PR #572, opened 2019-11-05
Remove config validation section

mfn:
At first I wanted to improve it but then I realized it's basically a duplicate of https://github.com/webonyx/graphql-php/blob/master/docs/type-system/schema.md#schema-validation

This "duplicate" likely exactly lead to this situation that there's now outdated information.

TL;DR: I think simply removing it here is enough as there's only one clear location now discussing schema validation

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -1,13 +1,3 @@
-# Config Validation
-Defining types using arrays may be error-prone, but **graphql-php** provides config validation
-tool to report when config has unexpected structure. 
-
-This validation tool is **disabled by default** because it is time-consuming operation which only 
-makes sense during development.
-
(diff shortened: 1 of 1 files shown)


docs/best-practices.md
```diff
@@ -16,4 +6,4 @@ Technically you can create several instances of your type (for example for tests
 will throw on attempt to add different instances with the same name.
 
 There are several ways to achieve this depending on your preferences. We provide reference 
-implementation below that introduces TypeRegistry class:
\ No newline at end of file
+implementation below that introduces TypeRegistry class:
```
mfn: Ugh, this was done by Githubs online editor 🤷‍♀️

vladar:
Makes sense, thanks!

PR #575, opened 2019-11-11
Shorten the composer scripts commands

spawnia:
That makes them quicker to type and saves time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,5 +62,5 @@ jobs:
       php: 7.1
       env: STATIC_ANALYSIS
       install: travis_retry composer install --prefer-dist
-      script: composer static-analysis
+      script: composer stan
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ For smaller contributions just use this workflow:
 * Fork the project.
 * Add your features and or bug fixes.
 * Add tests. Tests are important for us.
-* Check your changes using `composer check-all`.
+* Check your changes using `composer check`.
 * Add an entry to the [Changelog's Unreleases section](CHANGELOG.md#unreleased).
 * Send a pull request.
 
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -62,5 +62,5 @@ jobs:
       php: 7.1
       env: STATIC_ANALYSIS
       install: travis_retry composer install --prefer-dist
-      script: composer static-analysis
+      script: composer stan
```
simPod: I'd say `phpstan` Technically phpcs is also static analysis.

spawnia: We have `lint` for that.

I have seen that shortcut used in a few places now, it seems to be common.

simPod: Yes, not desired though :) https://twitter.com/OndrejMirtes/status/1189846443500154880

(As for lint, it's also wrong IMO.)

spawnia: I don't think naming the shortcuts just like the underlying tool is the greatest idea, as that couples them tightly.

simPod: They can still come and go, be easily replaced etc. I don't see any coupling here 🤔.

I have not found generic name for phpstan nor phpcs so far. Both are linting and static analysis tools. So personally I came to conflusion to call them their names, rather than using inaccurate ones.

spawnia: That is a fine way of looking at it, too.

In the sake of saving time, how about we just go with the PR as is? :wink:

simPod: @spawnia don't consider this a PR block 👋. Just wanted to express that the change in general goes in a good direction but in my opinion and by wording definition is not correct and might be misleading.

simPod: BTW we often use `cs`/`codestyle` for phpcs on php-only projects where there are no other style checks.

vladar:
I am pragmatic when it comes to questions like this. Shorter is better even if there is some ambiguity %)

PR #580, opened 2019-11-15
add automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix src --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/src/Error/Error.php
+++ b/src/Error/Error.php
@@ -105,8 +105,8 @@ class Error extends Exception implements JsonSerializable, ClientAware
 
         // Compute list of blame nodes.
         if ($nodes instanceof Traversable) {
-            $nodes = iterator_to_array($nodes);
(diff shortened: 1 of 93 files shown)


simPod:
Those functions are already imported if you check imports block on top of each file. They are already not looked up locally and the optimalization is achieved. Also the code is more readable when import is used rather than FQN.

draco2003:
Perfect! no worries than. 
Apologize for my automation not catching that prior to the PR.

simPod:
If you're interested, you can check out https://github.com/doctrine/coding-standard. It's the standard used here and probably the best out there. It has sniff for this.

PR #581, opened 2019-11-17
Simplify deferred/promise implementation

### LINKED ISSUE #573: How to correctly handle multiple deferred / promise?
In a resolver I need to access:
- `$model->relation1` => so I use a dataloader
- however after `relation1` is loaded, I need another relation _on that one_
- `$relation1->relation2` => again via dataloader

There is incentive using the dataloader also for `relation2` because this may be required in other resolvers too (without relation1)

I came up with a resolver looking like this:
```php
'resolve' => function (Model $model, $_) {
    $dataloader = $this->getDataLoaderFor('relation1');
    $dataloader->batch($model->relation1_id);

    $deferred = new Deferred(function () use ($model, $dataloader) {
        return $dataloader->load($model->relation1_id);
    });

    $syncPromise->then(function (Relation1 $relation1) use () {
        $dataloader = $this->getDataLoaderFor('relation2');
        $dataloader->batch($relation1->relation2_id);

        return new Deferred(function () use ($relation1, $dataloader) {
            return $dataloader->load($relation1->relation2_id);
        });
    });

    return new Deferred(function () use ($syncPromise) {
        return $syncPromise;
    });
},
```
I know closures are verbose in PHP already 😄 however I realized when I need _dependent_ deferred using `->then()`, I need to wrap it in yet another `Deferred` because `->then()` returns a `SyncPromise` which is not handled by webonxy.

Is this "amount of code" expected? Is the approach correct / good? Can I simplify the above?

Note: it's not about reducing LoC, I could do that sparing temporary variables. It's more about the approach, because currently it's:
- `Deferred`
- `->then()` returning a `SyncPromise`
- contains a closure which returns a `Deferred`
- to return the `SyncPromise` I need to wrap it in a `Deferred`

I guess I would have expected I could return `$deferred->then(…)` directly without wrapping it again.

Thanks!

vladar:
Merge `Deferred` and `SyncPromise`. This simplifies things a bit and allows certain new features (like Deferred chaining):

```php
$resolver = function() {
    return Deferred::create(function() {
        return 'foo';
    })->then(function($foo) {
        return $foo . ' bar';
    });
}
```

The caveats are:
1. The sequence of nested deferred calls might be suboptimal and don't produce the desired batching sequence.
2. We can't `wait` for a specific promise anymore. The whole chain of promises has to be resolved before `wait` returns (as we have a single queue now). 

So I encourage anyone using `deferred` resolvers to try this branch and report some feedback.

@mcg-web Can you check if this will affect your DataLoader implementation somehow?

CC @mfn @spawnia

Fixes #573


mfn:
Wow, thanks for the quick actions. Unfortunately it will take me a while, seems due to dependencies I can't just upgrade to this commit and need to fiddle around manually.

Will keep you updated, also regarding the batch sequence order.

src/Executor/Promise/Adapter/SyncPromise.php
```diff
@@ -47,6 +47,20 @@ public static function runQueue() : void
         }
     }
 
+    public function __construct(?callable $executor = null)
+    {
+        if (! $executor) {
```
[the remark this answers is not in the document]
spawnia: I think the way it is written now communicates the intent more clearly.

src/Deferred.php
```diff
@@ -5,58 +5,16 @@
...
-        $queue = self::getQueue();
-        while (! $queue->isEmpty()) {
-            /** @var self $dequeuedNodeValue */
-            $dequeuedNodeValue = $queue->dequeue();
-            $dequeuedNodeValue->run();
-        }
-    }
-
-    public function then($onFulfilled = null, $onRejected = null)
+    public static function create(callable $executor) : self
```
spawnia: Why is this function added in addition to the constructor?

src/Executor/Promise/Adapter/SyncPromiseAdapter.php
```diff
@@ -24,19 +23,19 @@ class SyncPromiseAdapter implements PromiseAdapter
...
     }
 
     /**
      * @inheritdoc
      */
     public function convertThenable($thenable)
     {
-        if (! $thenable instanceof Deferred) {
+        if (! $thenable instanceof SyncPromise) {
             throw new InvariantViolation('Expected instance of GraphQL\Deferred, got ' . Utils::printSafe($thenable));
```
spawnia:
```suggestion
            throw new InvariantViolation('Expected instance of ' . SyncPromise::class . ', got ' . Utils::printSafe($thenable));
```

src/Executor/Promise/Adapter/SyncPromise.php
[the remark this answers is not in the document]
simPod: Ultimately, we want to enable PHPStan check that allows only booleans in conditions. Implicitly coalescing objects to booleans is not good practice, increases WTF factor and can have unexpected results. Therefore it's best to get rid of it and have strict boolean comparisons.

src/Executor/Promise/Adapter/SyncPromiseAdapter.php
simPod: True 👍, also using sprintf would increase readability 🤔

src/Deferred.php
simPod: As for me, it helps with fluent call chaining readability.

`(new A())->doSth()`  
vs
`A::create()->doSth()`

Not sure it was the reason here tho but I like it.

src/Executor/Promise/Adapter/SyncPromiseAdapter.php
spawnia: I intensily dislike think `sprintf`, i think it makes it hard to know what goes into a message, is prone to errors and makes the code brittle and harder to change.

src/Executor/Promise/Adapter/SyncPromise.php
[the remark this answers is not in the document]
spawnia: Good point, i am fine with that. Upping the level of static correctness guarantees seems worth it.

spawnia:
> Wow, thanks for the quick actions. Unfortunately it will take me a while, seems due to dependencies I can't just upgrade to this commit and need to fiddle around manually.
> 
> Will keep you updated, also regarding the batch sequence order.

Had a dependency issue as well and could resolve it by aliasing the reference:

```json
"webonyx/graphql-php": "dev-simplify-deferred as 0.13.8"
```

src/Deferred.php
vladar: Yes, that's the only reason

src/Executor/Promise/Adapter/SyncPromiseAdapter.php
vladar: We don't expose `SyncPromise` in docs or anywhere in error messages because it may lead to confusion.

If you create `SyncPromise` without an `executor` function in the constructor and return from your resolver - it will never resolve (actually throw while trying). The whole point of `Deferred` was to ensure it never happens and that any resolver creates at least one `executor` to start the promise chain.

So I just don't want to expose `SyncPromise` to users (in a hope to avoid confusion).

vladar:
> I think Deferred extends SyncPromise is a BC break as result of $deferred instanceof SyncPromise with this change changes from false to true so might need to document it.

I agree that we should add a note about this in Upgrade.md

> Do we need Deferred at all?

We do for now. First for BC, second to enforce `executor` function in the constructor (without it there is no place where we could start resolving a chain of promises).

src/Executor/Promise/Adapter/SyncPromiseAdapter.php
spawnia: I can get behind that reasoning, how about we put that into the DocBlock of the `SyncPromise` class? A one-liner above this error message would also be helpful.

vladar: Yep, will do when I have a minute, thanks!

mcg-web:
> Can you check if this will affect your DataLoader implementation somehow?

yes this will have an impact on DataLoader but a positive impact with this change it will make the implementation easier since we dealing with a unique object instead of two.

spawnia:
@vladar i tried this branch in Lighthouse and ran our test suite against it, seems to be working fine.

src/Executor/Promise/Adapter/SyncPromise.php
```diff
@@ -163,4 +177,9 @@ public function then(?callable $onFulfilled = null, ?callable $onRejected = null
 
         return $tmp;
     }
+
```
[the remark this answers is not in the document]
vladar: Both `then` and `catch` return `self` actually. Added those typehints

src/Deferred.php
```diff
@@ -5,58 +5,22 @@
...
-            $dequeuedNodeValue = $queue->dequeue();
-            $dequeuedNodeValue->run();
-        }
-    }
-
-    public function then($onFulfilled = null, $onRejected = null)
+    /**
+     * @param callable() : mixed $executor
+     */
+    public static function create(callable $executor) : self
```
simPod: Are we fine with the fact that child class cannot override it to create self instance? This won't be possible:

```php

class A extends Deferred
{
    public static function create(callable $executor) : self
    {
       // something
        return parent::create($executor);
    }
}
```

I'm fan of final classes, just asking.

vladar: It is not meant to be extended. But if you really want to, you can still override the constructor (you shouldn't, really).

esamattis:
Hi! I been testing this with WPGraphQL. It's bit behind with graphql-php versions but this change in particular does not cause any issues with our tests.

I'm working on a field level caching solution for WPGraphQL at the moment and this change would help a lot with that since I need to tap into the promise chains so I can capture the resolved values.

Here's simplified version of what I'm trying to do

```php
add_filter('graphql_resolve_field', function( $defer, $field_key ) {

	return $defer->then(function($result) use ($field_key) {
		write_cache( $field_key, wp_json_encode($result));
		return $result;
	});


}, 100, 2);
```

Where `$defer` is is `GraphQL\Deferred`

This does not work with the latest graphql-php stable release because the `then` method actually changes the type of `$defer` which seems break some instanceof check in downstream.

This PR fixes that issue nicely 👌

PR #582, opened 2019-11-17
Make query batching implicit

simPod:
Query batching is IMO client's thing and server should support it implicitly. Looking at the code, the only thing the _query batching feature toggle_ did was throwing exception when batch request was received and batching was not enabled. 

Also, Apollo [states](https://www.apollographql.com/docs/react/networking/network-layer/#query-batching): 

> Batching works only with server that support batched queries (for example graphql-server). Batched requests to servers that don’t support batching will fail.

StandardServer supports it. The flag had no other behaviour so I came to conclusion was just adding unnecessary complexity. WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - **BREAKING:** Removed deprecated directive introspection fields (onOperation, onFragment, onField)
 - **BREAKING:** Removal of `VariablesDefaultValueAllowed` validation rule. All variables may now specify a default value.
 - **BREAKING:** renamed `ProvidedNonNullArguments` to `ProvidedRequiredArguments` (no longer require values to be provided to non-null arguments which provide a default value).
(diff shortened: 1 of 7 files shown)


vladar:
Query batching is something Apollo-specific. They introduced it initially because they didn't utilize fragments properly and had to run multiple queries instead of a single one composed of fragments. There is no spec or standard for it.

Query batching doesn't play well with complexity scoring and API throttling. So people who do not want to support it shouldn't care about possible overhead and security risks incurred.

So for me, this feature is a hack. Hopefully temporary. And `setQueryBatching` option is a good place to produce deprecation warnings in the future if it becomes unnecessary.

However, if it gets standardized we can revisit this.

vladar:
Pretty much related: https://lab.wallarm.com/graphql-batching-attack/

PR #584, opened 2019-11-18
Add support GitHub actions

Smolevich:
- Create ci file with configuration

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -15,6 +15,10 @@ jobs:
       matrix:
         php: [7.1, 7.2, 7.3, 7.4]
         env: ['EXECUTOR= DEPENDENCIES=--prefer-lowest', 'EXECUTOR=coroutine DEPENDENCIES=--prefer-lowest', 'EXECUTOR=', 'EXECUTOR=coroutine']
+        include:
+          - os: ubuntu-latest
+            php: 7.1
+            env: EXECUTOR=coroutine
     steps:
     - uses: actions/checkout@v1
     - name: Install PHP
@@ -40,13 +44,12 @@ jobs:
     - name: Code coverage
       run: |
         ./vendor/bin/phpunit --coverage-php /tmp/coverage/clover_executor.cov
-        EXECUTOR=coroutine ./vendor/bin/phpunit --coverage-php /tmp/coverage/clover_executor-coroutine.cov
+        $ENV ./vendor/bin/phpunit --coverage-php /tmp/coverage/clover_executor-coroutine.cov
         ./vendor/bin/phpcov merge /tmp/coverage --clover /tmp/clover.xml
         wget https://github.com/scrutinizer-ci/ocular/releases/download/1.5.2/ocular.phar
         php${{matrix.php}} ocular.phar code-coverage:upload --format=php-clover /tmp/clover.xml
       env:
-        COVERALLS_RUN_LOCALLY: 1
-        COVERALLS_REPO_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        ENV: ${{ matrix.env }}
     - name: CODING_STANDARD
       if: matrix.php == 7.1
       run: |
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -51,10 +51,10 @@ jobs:
     - name: Coding Standard
       if: matrix.php == 7.1
       run: |
-        composer install --prefer-dist
+        composer update --prefer-dist
         composer lint
     - name: PHPStan
       if: matrix.php == 7.1
       run: |
-        composer install --prefer-dist
+        composer update --prefer-dist
         composer stan
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -5,6 +5,7 @@ on:
     branches:
     tags:
   pull_request:
+  
 jobs:
   build:
     runs-on: ${{ matrix.operating-system }}
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -138,6 +138,12 @@ jobs:
     - name: Install Dependencies
       run: composer install ${DEPENDENCIES}
 
+    - name: Update xdebug version
+      run: |
+        sudo pecl channel-update pecl.php.net
+        sudo pecl install xdebug-2.9.0
+      continue-on-error: true
+
     - name: Show php version
       run: php -v
 
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -144,9 +144,6 @@ jobs:
         sudo pecl install xdebug-2.9.0
       continue-on-error: true
 
-    - name: Show php version
-      run: php -v
-
     - name: Code coverage
       run: |
         ./vendor/bin/phpunit --coverage-php /tmp/coverage/clover_executor.cov
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -143,7 +143,8 @@ jobs:
 
     - name: Code coverage
       run: |
-        pecl install xdebug-2.7.0
+        pear config-set http_proxy $(printenv http_proxy)
+        pecl install xdebug-2.8.1
         ./vendor/bin/phpunit --coverage-php /tmp/coverage/clover_executor.cov
         EXECUTOR=coroutine ./vendor/bin/phpunit --coverage-php /tmp/coverage/clover_executor-coroutine.cov
         ./vendor/bin/phpcov merge /tmp/coverage --clover /tmp/clover.xml
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -141,7 +141,7 @@ jobs:
     - name: Update xdebug version
       run: |
         sudo pecl channel-update pecl.php.net
-        sudo pecl install xdebug-2.8.1
+        sudo pecl install xdebug-2.7.0
 
     - name: Show php version
       run: php -v
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -30,12 +30,12 @@ jobs:
         echo $ENV
         ENV composer update --prefer-dist
       env:
-        ENV: $ {{ matrix.env }}
+        ENV: ${{ matrix.env }}
     - name: Run unit tests
       run: |
         ENV ./vendor/bin/phpunit --group default,ReactPromise
       env:
-        ENV: $ {{ matrix.env}}
+        ENV: ${{ matrix.env}}
     - name: Run style-fixer
       run: php${{ matrix.php }} vendor/bin/php-cs-fixer fix --dry-run
     - name: Code coverage
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Install dependencies
       run: |
         export $ENV
-        composer update --prefer-dist
+        composer update ${DEPENDENCIES}
       env:
         ENV: ${{ matrix.env }}
     - name: Run unit tests
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Run unit tests
       run: |
         ${ENV} ./vendor/bin/phpunit --group default,ReactPromise
-      with:
+      env:
         ENV: $ {{ matrix.env}}
     - name: Run style-fixer
       run: php${{ matrix.php }} vendor/bin/php-cs-fixer fix --dry-run
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -37,8 +37,6 @@ jobs:
         ./vendor/bin/phpunit --group default,ReactPromise
       env:
         ENV: ${{ matrix.env}}
-    - name: Run style-fixer
-      run: php${{ matrix.php }} vendor/bin/php-cs-fixer fix --dry-run
     - name: Code coverage
       run: |
         ./vendor/bin/phpunit --coverage-php /tmp/coverage/clover_executor.cov
@@ -54,9 +52,9 @@ jobs:
       if: matrix.php == 7.1
       run: |
         composer install --prefer-dist
-        ./vendor/bin/phpcs
+        composer lint
     - name: STATIC_ANALYSIS
       if: matrix.php == 7.1
       run: |
         composer install --prefer-dist
-        ./vendor/bin/phpcs
+        composer stan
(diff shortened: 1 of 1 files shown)


Smolevich:
[Link](https://github.com/Smolevich/graphql-php/runs/307958888) on ci

shmax:
Lookin' good. You might consider disabling phpstan's progress bar with the [no-progress](https://github.com/phpstan/phpstan/commit/1a9bacccb57b9091f9f5fb891d34d8913724a179) option,

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
...
+  push:
+    branches:
+    tags:
+  pull_request:
+
+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
simPod: can it run on something with smaller footprint, alpine ideally

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
+name: CI
+
+on:
+  push:
+    branches:
+    tags:
+  pull_request:
+
```
simPod: too many lines here

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
...
+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php: [7.1, 7.2, 7.3, 7.4]
+        env: ['EXECUTOR= DEPENDENCIES=--prefer-lowest', 'EXECUTOR=coroutine DEPENDENCIES=--prefer-lowest', 'EXECUTOR=', 'EXECUTOR=coroutine']
```
simPod: Would it be possible to make one per line?

```yaml
env:
  - arg1
  - arg2
```
?

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
...
+        include:
+          - os: ubuntu-latest
```
simPod: does os have to be repeated here?

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
...
+            php: 7.1
+            env: EXECUTOR=coroutine
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install PHP
+      uses: shivammathur/setup-php@master
```
simPod: This should be bound, master is variable

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
...
+      with:
+        php-version: ${{ matrix.php  }}
+        extension-csv: dom, mbstring
+        coverage: xdebug
+    - name: Show php version
```
simPod: This step can be dropped

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
...
+      run: php${{ matrix.php }} -v && composer -V
+    - name: Install dependencies
+      run: |
+        export $ENV
+        composer update --prefer-dist
```
simPod: Shouldn't this leverage DEPENDENCIES env var?

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
...
+      uses: shivammathur/setup-php@master
+      with:
+        php-version: ${{ matrix.php  }}
+        extension-csv: dom, mbstring
+        coverage: xdebug
+    - name: Show php version
+      run: php${{ matrix.php }} -v && composer -V
+    - name: Install dependencies
+      run: |
+        export $ENV
```
simPod: What's in $ENV? Why is it needed?

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
...
+        ./vendor/bin/phpunit --coverage-php /tmp/coverage/clover_executor.cov
+        ./vendor/bin/phpunit --coverage-php /tmp/coverage/clover_executor-coroutine.cov
+        ./vendor/bin/phpcov merge /tmp/coverage --clover /tmp/clover.xml
+        wget https://github.com/scrutinizer-ci/ocular/releases/download/1.5.2/ocular.phar
+        php${{matrix.php}} ocular.phar code-coverage:upload --format=php-clover /tmp/clover.xml
+      env:
+        ENV: ${{ matrix.env }}
+    - name: CODING_STANDARD
```
simPod:
```suggestion
    - name: Coding Standard
```

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,64 @@
...
+      if: matrix.php == 7.1
+      run: |
+        composer install --prefer-dist
+        composer lint
+    - name: STATIC_ANALYSIS
```
simPod:
```suggestion
    - name: PHPStan
```

.github/workflows/ci-build.yml
Smolevich: It's not docker image. It is virtual environment, it should one of three values: macos, windows and ubuntu. See [docs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners)

.github/workflows/ci-build.yml
Smolevich: Why?

.github/workflows/ci-build.yml
Smolevich: because env include more than one environment variable, export used for setting environment variables

.github/workflows/ci-build.yml
simPod: lines 8-10 should be only one line

.github/workflows/ci-build.yml
simPod: Oh. Then fix version to `ubuntu-18.04` so the builds are the same overtime

.github/workflows/ci-build.yml
simPod: And isn't it already set? 🤔

.github/workflows/ci-build.yml
Smolevich: no

.github/workflows/ci-build.yml
Smolevich: fixed

.github/workflows/ci-build.yml
Smolevich: Pair of environment variables from level `jobs.build.strategy.matrix.env` will be mapped in env variable with name `ENV` inside step

.github/workflows/ci-build.yml
Smolevich: Ok, i replace on `ubuntu-18.04`

.github/workflows/ci-build.yml
Smolevich: yes, it needed

.github/workflows/ci-build.yml
Smolevich: ok, i fix it

.github/workflows/ci-build.yml
Smolevich: fixed

Smolevich:
@simPod, do you need this [mechanism](https://github.com/shivammathur/setup-php#cache-dependencies)?

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,60 @@
...
+      uses: shivammathur/setup-php@1.5.2
+      with:
+        php-version: ${{ matrix.php  }}
+        extension-csv: dom, mbstring
+        coverage: xdebug
+    - name: Install dependencies
+      run: |
+        export $ENV
+        echo ${DEPENDENCIES}
+        composer update ${DEPENDENCIES}
```
simPod: Why update? 🤔

simPod:
@Smolevich caching composer deps should speed up the build

Smolevich:
> @Smolevich caching composer deps should speed up the build

I added

Smolevich:
@simPod, https://github.com/Smolevich/graphql-php/runs/312457435 step with sending code covarage failed, can you tell me how it works in travis? Maybe @vladar knows?

simPod:
@Smolevich do you have scrutinizer with your fork added? I suppose not, that's why 404.

Smolevich:
> @Smolevich do you have scrutinizer with your fork added? I suppose not, that's why 404.

What do you mean say about as scrutinizer?

simPod:
@Smolevich you have to register to scrutinizer and add your fork there

Smolevich:
@simPod, I add scrutinizer integration

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,70 @@
...
+      id: composer-cache
+      run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+    - name: Cache dependencies
+      uses: actions/cache@v1
+      with:
+        path: ${{ steps.composer-cache.outputs.dir }}
+        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+        restore-keys: ${{ runner.os }}-composer-
+    - name: Install dependencies
+      run: |
```
simPod: Can this be separated into multiple commands? Kind of weird how it's hidden

image

image

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,70 @@
...
+        php${{matrix.php}} ocular.phar code-coverage:upload --format=php-clover /tmp/clover.xml
+      env:
+        ENV: ${{ matrix.env }}
+    - name: Coding Standard
+      if: matrix.php == 7.1
+      run: |
+        composer update --prefer-dist
+        composer lint
+    - name: PHPStan
+      if: matrix.php == 7.1
```
simPod: Currently, PHPStan is run 4 times in all these

image

I'd prefer to have only PHPStan job, same for CS and I guess Coverage

.github/workflows/ci-build.yml
Smolevich: Do you want it to be in different steps?

simPod: Perhaps? I mean something like this

image

So commands are visible at first sight. Currently, significant `composer update` is hidden under `Run export $ENV` which is not significant for contributors.

.github/workflows/ci-build.yml
Smolevich: @simPod, i added only one step for  PHPStan job, CS and Coverage

.github/workflows/ci-build.yml
Smolevich: Is it not actual

simPod: install should be good enough as composer.lock is not versioned in libraries

Smolevich: https://github.com/webonyx/graphql-php/blob/master/.travis.yml#L27 current travis build uses `composer update`

.github/workflows/ci-build.yml
simPod: I see but eg. unit tests are still run for CS.

https://github.com/Smolevich/graphql-php/runs/328856568

I believe we have to split it into more jobs. We should have similar structure as is in Travis now. To have a job for 1. running tests, then another job to run 2. codesniffer and another job to run 3. phpstan.

It would also be nice to reuse some parts of the config to keep it dry, this might help https://confluence.atlassian.com/bitbucket/yaml-anchors-960154027.html

Your work is definitely appreciated but I'd love to do this right so even though this works, Travis build is currently still superior. Also another problem is that it's hard to get inspired elsewhere, I don't know of any project that has good Actions setup.

.github/workflows/ci-build.yml
simPod: Oh, my bad! Then lets keep it and don't change it 👍

.github/workflows/ci-build.yml
Smolevich: Export needed because env might multiple env variables and i didn't find beatiful solving for export

.github/workflows/ci-build.yml
Smolevich: For example https://github.com/dunglas/vulcain/blob/master/.github/workflows/ci.yml 
https://github.com/api-platform/api-platform/blob/master/.github/workflows/ci-cd.yml

Smolevich: I add if condition for phpunit tests step
I don't understand why you attach link to bitbucket docs, but there is no using anchors in current travis configuration
Github actions doen't support anchors in yaml config yet. Discussing https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/td-p/30336

Smolevich: @simPod can see on PR again?

simPod: I thought those anchors might help to clear up things that might repeat. If it's not supported, no can do :)

simPod:
@Smolevich I'll try to implement what I mean this weekend

Smolevich:
> I have pushed some tweaks, you can cherry pick them from [simPod@469aa55](https://github.com/simPod/graphql-php/commit/469aa55c8e8d2da8121c2961af6450b43db92765) and build on top of it if you agree.
> 
> Missing support for yaml anchors is ok, eventually it will come later and we can reduce config repetition.
> 
> What worries me is that coverage job runs for 38m! 😄 there must be something wrong with it, dunno what tho. Travis does coverage in around 6m.

 I will see on your changes

Smolevich:
Long time of code coverage linked with https://xdebug.org/announcements/2019-12-02

Smolevich:
@simPod, I think how i can update xdebug version without error. See [step](https://github.com/Smolevich/graphql-php/commit/c866829561bacbf67ad6b19ede7feeab4ac05120/checks?check_suite_id=349596084)

Smolevich:
I created issue in https://github.com/shivammathur/setup-php/issues/109

Smolevich:
@simPod i fixed time executing

simPod:
@Smolevich 2:50, looks much better than what we have in travis now 👍

Smolevich:
@simPod, is it  enough changes or something else?

.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,156 @@
...
+    - name: Install Dependencies
+      run: composer install ${DEPENDENCIES}
+
+    - name: Update xdebug version
+      run: |
+        sudo pecl channel-update pecl.php.net
+        sudo pecl install xdebug-2.9.0
+      continue-on-error: true
+
+    - name: Show php version
```
simPod: -debug

Smolevich: @simPod do you want that I delete this step?

simPod: Yes please :)

vladar:
Yeah, we should definitely try it first and then if it works OK in the real-life - switch to Github Actions. Thanks a lot for this big work! @simPod feel free to set it up as needed.

.github/workflows/ci-build.yml
Smolevich: Ok, i delete it

Smolevich: @simPod fixed

PR #585, opened 2019-11-20
Unify implementation of defaultValueExists

spawnia:
It is now the same as `FieldArgument`.


vladar:
Code style check fails again %) They are a bit annoying but `composer fix` should do the trick.

vladar:
Thanks!

PR #586, opened 2019-11-20
Scrutinizer: Analyze only src

simPod:
I noticed [this failure](https://scrutinizer-ci.com/g/webonyx/graphql-php/inspections/68739c79-cb95-486f-8742-0ad3701250ec) in build from https://github.com/webonyx/graphql-php/pull/539

What about this change? Tests are here often more "messy", long, and have high complexity, I'd rather focus on src with these checks.


vladar:
I agree, thanks!

PR #587, opened 2019-11-21
Wrap field path in quotes

simPod:
Recently I encountered an error message like `Cannot return null for non-nullable field queryname.field.` First I thought there's a bug and 3rd field name is missing `queryname.field.namemissinghere`. Then I realized it's just a dot in a sentence 🤔


vladar:
Why is it a `BC break`? In case if someone is parsing those messages? Or am I missing something?

simPod:
Exactly. Documented.

PR #591, opened 2019-12-02
Fix ServerConfig and SchemaConfig

enumag:
According to setDebug it can be `int` as well.

It would be nice to document in the docblock of setDebug what `int` values are `true` and `false` equivalent to. Can you tell me?

----

Reasons for WIP:

There are more issues here... for example `fieldResolver` can be `null`. I need to check everything.

Then I want to check SchemaConfig too - query, mutation and subscription are all nullable for instance...

### DIFF
--- a/src/Utils/SchemaExtender.php
+++ b/src/Utils/SchemaExtender.php
@@ -609,11 +609,7 @@ class SchemaExtender
             }
         }
 
-        $schemaExtensionASTNodes = count($schemaExtensions) > 0
-            ? ($schema->extensionASTNodes
-                ? array_merge($schema->extensionASTNodes, $schemaExtensions)
-                : $schemaExtensions)
-            : $schema->extensionASTNodes;
+        $schemaExtensionASTNodes = array_merge($schema->extensionASTNodes, $schemaExtensions);
 
         $types = array_merge(
             // Iterate through all types, getting the type definition for each, ensuring
--- a/src/Server/ServerConfig.php
+++ b/src/Server/ServerConfig.php
@@ -54,7 +54,7 @@ class ServerConfig
         return $instance;
     }
 
-    /** @var Schema */
+    /** @var Schema|null */
     private $schema;
 
     /** @var mixed|callable */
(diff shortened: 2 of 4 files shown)


simPod:
_Just a note, I tried to do something a bit related here https://github.com/webonyx/graphql-php/pull/423/files#diff-c328ef62636d6b01012ec2265c28b058 Though it's BC and long run if even that._

enumag:
I don't really want to make any BC break here, just fix the wrong annotations and add some docs about the meaning of the bool values.

simPod:
This seems 100% valid to me, just a bug fix, no bc break. Waiting for no-wip :)

src/Utils/SchemaExtender.php
```diff
@@ -609,11 +609,7 @@ static function (string $typeName) use ($schema) {
             }
         }
 
-        $schemaExtensionASTNodes = count($schemaExtensions) > 0
-            ? ($schema->extensionASTNodes
-                ? array_merge($schema->extensionASTNodes, $schemaExtensions)
-                : $schemaExtensions)
-            : $schema->extensionASTNodes;
+        $schemaExtensionASTNodes = array_merge($schema->extensionASTNodes, $schemaExtensions);
```
enumag: both are always arrays so the conditions are useless, we can just merge them... empty arrays don't matter

enumag:
@simPod done!

src/Utils/SchemaExtender.php
spawnia: Actually, it looks like they are not. Check travis:

```
1) GraphQL\Tests\Type\ValidationTest::testRejectsASchemaExtendedWithInvalidRootTypes
array_merge(): Expected parameter 1 to be an array, null given
/home/travis/build/webonyx/graphql-php/src/Utils/SchemaExtender.php:612
/home/travis/build/webonyx/graphql-php/tests/Type/ValidationTest.php:530
```

enumag:
@spawnia yeah I saw it, just didn't have the time to look into it yet.

src/Utils/SchemaExtender.php
enumag: @spawnia yeah I saw it, just didn't have the time to look into it yet.

enumag: @spawnia Should be fixed now.

enumag:
> though I'd put most of these into native type hints where eligible

@simPod I agree ofc but adding type hints might be a BC break in some cases so I chose not to do it... but if @vladar gives me a green to do it, I will.

simPod:
@enumag it's not bc break where type hint is already declared in phpdoc. If not in phpdoc, it's bc break.

> if @vladar gives me a green to do it, I will

Sure, but it's usually done in PRs here so I guess he's fine with it

vladar:
I think it will go to a new major, so BC breaks like this shouldn't be a big issue.

vladar:
Thanks! And sorry for the delay.

Issue #285, opened 2018-06-11
Put NodeList first in the PHPDoc blocks for fields that return them


spawnia:
When working with the AST, i found that variables that contain references to multiple children are defined as typed arrays, for example in https://github.com/webonyx/graphql-php/blob/master/src/Language/AST/ObjectTypeDefinitionNode.php

However, they do actually return a `NodeList`, for example:

````php
>>> GraphQL\Language\Parser::parse('type Foo { bar: Int }')->definitions[0]->fields
=> GraphQL\Language\AST\NodeList {#951}

>>> GraphQL\Language\Parser::parse('type Foo { bar(baz: Int): Int }')->definitions[0]->fields[0]->arguments
=> GraphQL\Language\AST\NodeList {#1046}
````

I think the return type should match the definition to avoid surprises.

Additionally, the behaviour for how empty lists function may differ:

````php
>>> GraphQL\Language\Parser::parse('type Foo {}')->definitions[0]->fields
=> GraphQL\Language\AST\NodeList {#1012}

>>> GraphQL\Language\Parser::parse('type Foo { bar: Int }')->definitions[0]->fields[0]->arguments
=> []
````

This has caused me trouble when trying to use `merge()` on the arguments of a field.

**Comment by vladar** [comment]
Yeah, the code base starves for a cleanup and maybe some code style analysis applied. PRs welcome! %) 

This particular issue is caused by the fact that originally lists were just arrays, but docblocks were not updated.

**Comment by vladar** [comment]
Looks like I misunderstood the following note:
```php
GraphQL\Language\Parser::parse('type Foo { bar: Int }')->definitions[0]->fields[0]->arguments
=> []
```
If it's the case, then it is a bug. AST must contain `NodeList` here. But I can't reproduce it. For me it returns `NodeList`, not `[]`.

Which version do you use? Can you double check with the latest?

spawnia:
It does return `NodeList` in v0.12.4 indeed, so that seems like it was fixed.

However, the main issue actually is that the return types for lists of nodes are still not correct. While it would certainly be an easy fix to change them all to NodeList, i do not quite like that this would then hide the underyling type.

vladar:
Yeah, that was a primary reason why they are marked as `SomeNode[]`. This has more value over just `NodeList`.

PHP type annotations are too weak to express complex types. Maybe some IDEs do support things like `NodeList<SomeNode>` but obviously not all of them.

vladar:
[...]

PR #592, opened 2019-12-05
Upgrade to phpstan 0.12

spawnia:
I am particularly interested in the new generics feature that will allow us to resolve https://github.com/webonyx/graphql-php/issues/285


composer.json
```diff
@@ -15,10 +15,10 @@
...
   "require-dev": {
     "doctrine/coding-standard": "^6.0",
-    "phpbench/phpbench": "^0.14.0",
-    "phpstan/phpstan": "0.11.16",
-    "phpstan/phpstan-phpunit": "0.11.2",
-    "phpstan/phpstan-strict-rules": "0.11.1",
+    "phpbench/phpbench": "^0.14",
+    "phpstan/phpstan": "^0.12",
+    "phpstan/phpstan-phpunit": "^0.12",
+    "phpstan/phpstan-strict-rules": "^0.12",
```
simPod: We need to lock patch versions as well. New "patch" release might break build otherwise.

src/Language/AST/NodeList.php
```diff
@@ -14,13 +14,16 @@
 use function count;
 use function is_array;
 
+/**
+ * @template T
```
simPod: I'd use vendor prefix for these `@phpstan-template`

src/Language/AST/NodeList.php
```diff
@@ -14,13 +14,16 @@
...
+ */
 class NodeList implements ArrayAccess, IteratorAggregate, Countable
 {
-    /** @var Node[]|mixed[] */
+    /** @var T[] */
```
simPod: Same here, let's keep `@var` for IDE support and use `@phpstan-var` for generic type

src/Language/AST/NodeList.php
simPod: Shouldn't we also have

```php
@phpstan-implements ArrayAccess<T>
@phpstan-implements IteratorAggregate <T>
```

src/Language/AST/NodeList.php
```diff
@@ -50,7 +53,7 @@ public function offsetExists($offset)
     /**
      * @param mixed $offset
      *
-     * @return mixed
+     * @return T
```
simPod: If we add `implements` annotations, this might be possible to drop.

composer.json
simPod: `0.12.0`

src/Language/AST/NodeList.php
spawnia: So duplicate the types?

spawnia:
@simPod i am stuck with the remaining issues. I invited you as a collaborator on my fork, maybe you can give it a shot.

phpstan.neon.dist
```diff
@@ -25,6 +25,11 @@ parameters:
         - "~Access to an undefined property GraphQL\\\\Language\\\\AST\\\\.+::\\$didEnter~"
         - "~Access to an undefined property GraphQL\\\\Language\\\\AST\\\\.+::\\$didLeave~"
         - "~Access to an undefined property GraphQL\\\\Language\\\\AST\\\\Node::\\$value~"
+        # TODO fix those newly added errors and finish upgrading to PhpStan 0.12
```
spawnia: I think we can do those in separate PR's, since those throw over a hundred errors if you combine them.

simPod:
@spawnia will take a look tmrw

src/Error/Error.php
```diff
@@ -171,7 +171,7 @@ public static function createLocatedError($error, $nodes = null, $path = null)
             $message = (string) $error;
         }
 
-        return new static(
+        return new self(
```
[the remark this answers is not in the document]
spawnia: Right. I must say i don't really like the phpstan rule behind this for library code. Locking down classes in very rigid ways hurts reusability and flexibility.

spawnia:
@simPod great job at getting the `NodeList` working. You have a really good understanding of PhpStan!

I did clean up the phpstan config a bit, we can iterate on the smaller issues piece by piece from now on.

shmax:
It's nice to see a collaborative branch like this take shape. 👍 

Are either of you interested in teaming up on PHPStan level 4?

spawnia:
> It's nice to see a collaborative branch like this take shape. 
> 
> Are either of you interested in teaming up with me on PHPStan level 4?

Definitely interested in leveling up further. There are a whole bunch of `ignoreErrors` that i think we should try and fix before piling on even more from level 4.

shmax:
> There are a whole bunch of ignoreErrors that i think we should try and fix before piling on even more from level 4.

Sounds good.

src/Language/AST/NodeList.php
```diff
@@ -6,101 +6,127 @@
...
      */
-    public function offsetExists($offset)
+    public function offsetExists($offset) : bool
     {
         return isset($this->nodes[$offset]);
     }
 
     /**
-     * @param mixed $offset
+     * TODO this *should* always return a Node, but sometimes a string is in here?
```
spawnia: @vladar can you enlighten us on what is going on here?

During some tests, the `Printer` puts items such as `"input:SomeInput"` in a NodeList and mutates it somehow? That makes static validation quite hard/impossible, since we can no longer guarantee that the type `T` is always the same.

spawnia:
> phpstan v0.12.1 just came out few minutes ago, can bump ;)

Done, they even fixed a bug.

src/Language/AST/NodeList.php
vladar: Printer uses a Visitor which can transform any node to an arbitrary data structure (in case of printer it transforms a node to a printed version). See here:

https://github.com/webonyx/graphql-php/blob/34fe9615097be1f1d7f628ef563489973905e759/src/Language/Visitor.php#L31-L58

Technically visitor doesn't mutate but creates a new structure (clones where appropriate). I guess for strict typing we will have to adjust Visitor and replace NodeList with array (or less strict List structure when modifications are made).

spawnia: Switching to an array when transforming seems fine, since i think the lazyness aspect of `NodeList` does not matter at this point anymore.

How about we open a separate PR for that and merge this one as is to enable further work on phpstan:0.12 compatibility?

vladar: Makes sense, we have a TODO in place there so can revisit when Visitor part is unblocked

src/Error/Error.php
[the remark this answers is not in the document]
spawnia: @vladar i think the route we choose to go on these issues is an important design decision. There are basically two paths we can go down.

We can lock everything down to the max, making all classes final and methods preferrably private. This does prevent misuse and keeps the API surface small, but hurts flexibility and extensibility. I personally think this approach makes more sense in large project codebases where you want to keep core components clean and limit technical debt.

The other way would be to keep classes open, allowing extending them and overwriting or calling protected methods. The BC guarantees are only upheld for the parts of the library marked with `@api`. While this provides some footguns and may cause pain for developers who choose to mess with the insides of this library, i think that this is not really a drawback from a library perspective. It gives more power and responsibility to users of this library.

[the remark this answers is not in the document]
vladar: I agree with you here. I think it is quite an impractical rule. `@api` convention should be enough to communicate which APIs we consider `stable`. Then the risk of overriding something is on developers messing with internals.

I know this feeling when library authors didn't consider a specific use-case and locked-down everything so that you couldn't even "fix" it on the user-land without forking or some crazy reflection hacks.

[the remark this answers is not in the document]
spawnia: I reverted the breaking changes here for now, we can continue this discussion in https://github.com/webonyx/graphql-php/issues/593

src/Validator/Rules/KnownDirectives.php
```diff
@@ -186,6 +188,8 @@ private function getDirectiveLocationForASTPath(array $ancestors)
                     ? DirectiveLocation::INPUT_FIELD_DEFINITION
                     : DirectiveLocation::ARGUMENT_DEFINITION;
         }
+
+        throw new Exception('Unknown directive location: ' . get_class($appliedTo));
```
simPod: Maybe something like `InvariantViolation::shouldNotHappen()` can be used?

PR #562, opened 2019-10-23
Clean CS config, upgrade Doctrine CS to v8

simPod:
_Dropping PHP 7.1 is prerequisite for this https://github.com/webonyx/graphql-php/pull/470_

This includes:

- CS config cleanup
- upgrade to Doctrine CS v8
----------------
~~- upgrade to Doctrine CS v7~~

image

PR #594, opened 2019-12-09
Drop unnecessary parenthesis

simPod:
This is part of https://github.com/webonyx/graphql-php/pull/562 which is blocked.

This can be merged before waiting.


simPod:
Ok, this is false positive.

PR #596, opened 2019-12-09
Fix CoroutineExecutor

enumag:
Fixes this bug:

```
Error: Function name must be a string

<projectpath>/vendor/webonyx/graphql-php/src/Experimental/Executor/CoroutineExecutor.php:367
```

I wrote down how this happened here: https://github.com/webonyx/graphql-php/pull/551#issuecomment-563269245

### DIFF
--- a/src/Experimental/Executor/CoroutineExecutor.php
+++ b/src/Experimental/Executor/CoroutineExecutor.php
@@ -367,7 +367,7 @@ class CoroutineExecutor implements Runtime, ExecutorImplementation
 
         $this->run();
 
-        if ($this->pending > 0) {
+        if ($this->pending > 0 || $this->doResolve === null) {
             return;
         }
 
--- a/src/Experimental/Executor/CoroutineExecutor.php
+++ b/src/Experimental/Executor/CoroutineExecutor.php
@@ -91,7 +91,7 @@ class CoroutineExecutor implements Runtime, ExecutorImplementation
     /** @var int|null */
     private $pending;
 
-    /** @var callable */
+    /** @var callable|null */
     private $doResolve;
 
     public function __construct(
@@ -177,6 +177,7 @@ class CoroutineExecutor implements Runtime, ExecutorImplementation
         $this->queue      = new SplQueue();
         $this->schedule   = new SplQueue();
         $this->pending    = 0;
+        $this->doResolve  = null;
 
         $this->collector = new Collector($this->schema, $this);
         $this->collector->initialize($this->documentNode, $this->operationName);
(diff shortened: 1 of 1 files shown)


vladar:
@jakubkulhan Could you please review?

jakubkulhan:
@enumag The fact that the promise calls `$onFulfilled` synchronously in the same tick of the event loop seems to me like a broken promise implementation, see https://promisesaplus.com/#point-34 & https://promisesaplus.com/#point-67

However, the `CoroutineExecutor` suffers from the same mistake. The fix you submitted works around the bug, but doesn't fix it.

IMHO the fix should look something like this:

```php
class CoroutineExecutor implements Runtime, ExecutorImplementation
{
    /** @var bool */
    private $running = false;

    private function run()
    {
        $this->running = true;
        RUN:
        // ...
                        $this->promiseAdapter
                            ->then(
                                $value,
                                function ($value) use ($strand) {
                                    $strand->success = true;
                                    $strand->value   = $value;
                                    $this->queue->enqueue($strand);
                                    if (!$this->running) {
                                        $this->done();
                                    }
                                },
                                function (Throwable $throwable) use ($strand) {
                                    $strand->success = false;
                                    $strand->value   = $throwable;
                                    $this->queue->enqueue($strand);
                                    if (!$this->running) {
                                        $this->done();
                                    }
                                }
                            );
                        continue;
        // ...
        if ($this->pending > 0 || $this->schedule->isEmpty()) {
            $this->running = false;
            return;
        }
        // ...
    }
{
```

enumag:
> @enumag The fact that the promise calls $onFulfilled synchronously in the same tick of the event loop seems to me like a broken promise implementation, see promisesaplus.com/#point-34 & promisesaplus.com/#point-67

Possibly... but then again Amphp doesn't follow Promises A+ so perhaps this is intentional. (ping @trowski, @kelunik)

@jakubkulhan Also it seems that you were actually trying to make it work in such case: https://github.com/webonyx/graphql-php/blob/0f4434d773885b75ad089371e95347aa8ba43c1f/src/Experimental/Executor/CoroutineExecutor.php#L306

As for your suggested fix, it won't work as you wrote it because you still increase `$this->pending` but never decrease it. I'd move the `                                    if (!$this->running) {` condition into `$this->done()`.

jakubkulhan:
@enumag Yes, missed that, `$this->pending` also needs to be decremented. I'd rather move decremeting pending to `then()` callbacks and rename `done()` method to `resume()`

kelunik:
Neither ReactPHP nor Amp follows Promises A+. ReactPHP implements Promises A, but not Promises A+. In case of Amp this is fully intentional, because it reduces the latency quite a bit if not every single promise callback needs to await the next tick.

enumag:
Hmm somehow the problem disappeared in my project. I'm not sure why but I can no longer reproduce the bug. I'll reopen this in case it appears again.

PR #601, opened 2020-01-04
DO NOT MERGE: Drop Travis

simPod:
https://github.com/webonyx/graphql-php/issues/574#issuecomment-569678316

GA test:

- [x] test CS break
- [x] test PHPStan break
- [x] test unit test break

To compare with Travis run https://travis-ci.org/webonyx/graphql-php/builds/632608253

PR #603, opened 2020-01-04
Use Coveralls, drop Scrutinizer

simPod:
I was not very satisfied with Scrutinizer in several past weeks (a lot of false positives, bugs etc.).

I though might be good idea to switch to something else. Tried to replace it with Coveralls. WDYT?

You can see the report here: https://coveralls.io/github/simPod/graphql-php

Also dropped the coverage from Travis as we plan to switch to GA anyways in #601.

@vladar If we decide to go with it, we just need to register repository on Coveralls.io and add repo token as a new secret:

image


Smolevich:
@simPod, this [library](https://github.com/cedx/coveralls.php) supports GIthub Actions
I prepare PR here https://github.com/php-coveralls/php-coveralls/pull/275, but this in paused

simPod:
@Smolevich thanks for links. Is there anything particular you would like to point out?

simPod:
@vladar do not forget to [add the secret](https://github.com/webonyx/graphql-php/settings/secrets)

vladar:
Yeah, just added it

simPod:
I have to yet figure out why the coverage is not being linked to branches.

vladar:
Yeah, noticed it too

simPod:
@Smolevich ok, now I get what you meant. Cool!

PR #606, opened 2020-01-06
Speedup coverage

simPod:
- Replace XDebug by PCOV (faster for coverage)
- Upgrade PHPUnit to v8 (keep v7 for EOLed PHP 7.1)
- Run coverage on PHP 7.2
- Fix PHPUnit v8 deprecations
- Port assertArraySubset() from dms/phpunit-arraysubset-asserts for the method in PHPUnit it's deprecated. Also library does not support PHP 7.1, therefore must be ported


vladar:
Sorry for a long absence - had a crazy couple of months.

PR #613, opened 2020-01-20
Change type hints in Utils.php to iterable pseudo-type

spawnia:
This has come up in a discussion in https://github.com/webonyx/graphql-php/pull/539.

Added strong types where possible and changed the PHPDoc to use the `iterable` pseudo-type.

### DIFF
--- a/src/Utils/Utils.php
+++ b/src/Utils/Utils.php
@@ -103,18 +103,18 @@ class Utils
     }
 
     /**
-     * @param mixed|Traversable $traversable
+     * @param iterable<mixed> $iterable
      *
      * @return mixed|null
      */
-    public static function find($traversable, callable $predicate)
+    public static function find($iterable, callable $predicate)
     {
         self::invariant(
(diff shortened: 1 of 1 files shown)


vladar:
Looks good to me but there are conflicts in this branch. Can you please take a look?

src/Utils/Utils.php
```diff
@@ -199,21 +199,21 @@ public static function mapKeyValue($traversable, callable $fn)
     }
 
     /**
-     * @param mixed|Traversable $traversable
+     * @param iterable<mixed> $iterable
```
simPod: I'd also add a documentation for key everywhere (`iterable<keytype, valuetype>`)

spawnia: Feel free to add a PR.

spawnia:
@vladar fixed.

As @simPod noted,there can be more improvements - but i think we can add them later, as needed.

vladar:
Awesome, thanks! 🚢

PR #614, opened 2020-01-21
Fixed assumption that throwables from custom scalar are always client-safe

adorin:
Exceptions thrown in parseLiteral are currently caught and thrown as `Error` with client-safe true.

### DIFF
--- a/src/Validator/Rules/ValuesOfCorrectType.php
+++ b/src/Validator/Rules/ValuesOfCorrectType.php
@@ -224,7 +224,11 @@ class ValuesOfCorrectType extends ValidationRule
                         $context,
                         $fieldName
                     ),
-                    $node
+                    $node,
+                    null,
+                    null,
+                    null,
+                    $error
                 )
             );
         }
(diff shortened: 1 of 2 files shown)


vladar:
Thanks for the PR!

Looks correct to me but need to spend a bit time on this to check locally and ensure I do not miss anything here.

vladar:
Sorry for the delay but finally merged. Thank you! 👍

mfn:
Why this change? Why wouldn't what is essential type validation not deemed client safe suddenly?

The impact is that when I provide a wrong type for an input arg before I got (for example):
- `Field "someMutation" argument "activated" requires type Boolean, found "false".`
  but now I just get
- `Internal server error`

Even for the _consumer_ of the API this is unfriendly.

Any exception passed to `\GraphQL\Error\Error` in `$previous` marks it as `internal`. Previously it was not passed, thus it was of type `graphql` and thus `isClientSafe=true`.

I would like to better understand the arguments _for_ this change as I think this should be reverted.

spawnia:
@mfn definitely agree the built-in scalars are client-safe and should produce a friendly error. However, not every scalar might be.

For example (not saying this is good design), someone might be using a scalar `Password` to validate login.

mfn:
@spawnia sure, but then they just throw a custom error, no? 🤷‍♀️

PR #621, opened 2020-02-14
Add an option to Schema Printer to exclude descriptions

StymiedSloth:
This helps certain tools like GraphQL Inspector to function properly

GraphQL Inspector(https://github.com/kamilkisiela/graphql-inspector) has a real problem with the way multi line comments are output by Schema Printer.

Adding an option to exclude type and field descriptions resolves this.


simPod:
But why not fix the original problem then?

StymiedSloth:
Closing this. Might be a good option for people, but you're right. Gotta fix the tooling.

Closing the PR.

PR #626, opened 2020-02-28
Make sure resolveField function is preserved in SchemaExtender

SevInf:
For ObjectTypes, SchemaExtender did not copy over `resolveType` config
option. See test for an example.


phpstan-baseline.neon
```diff
@@ -2277,7 +2277,7 @@ parameters:
 
 		-
 			message: "#^Anonymous function should have native return typehint \"string\"\\.$#"
-			count: 5
+			count: 6
```
simPod: you should add native return typehint string to the function instead

SevInf: done

PR #631, opened 2020-03-12
print @deprecated directive when deprecationReason is empty string

Oliboy50:
## Why
- `if ($reason === '')` can't be `true` after `if (empty($reason))`
=> because `empty()` already test all falsy values (including `''`)

## How
- add tests for current `@deprecated` printing behavior
- ~remove the useless condition~ replace `empty()` with `=== null`

### DIFF
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -393,10 +393,10 @@ class SchemaPrinter
     private static function printDeprecated($fieldOrEnumVal) : string
     {
         $reason = $fieldOrEnumVal->deprecationReason;
-        if (empty($reason)) {
+        if ($reason === '' || $reason === null) {
             return '';
         }
-        if ($reason === '' || $reason === Directive::DEFAULT_DEPRECATION_REASON) {
+        if ($reason === Directive::DEFAULT_DEPRECATION_REASON) {
             return ' @deprecated';
         }
 
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -393,10 +393,10 @@ class SchemaPrinter
     private static function printDeprecated($fieldOrEnumVal) : string
     {
         $reason = $fieldOrEnumVal->deprecationReason;
-        if ($reason === '' || $reason === null) {
+        if ($reason === null) {
             return '';
         }
-        if ($reason === Directive::DEFAULT_DEPRECATION_REASON) {
+        if ($reason === '' || $reason === Directive::DEFAULT_DEPRECATION_REASON) {
             return ' @deprecated';
         }
 
--- a/tests/Utils/SchemaPrinterTest.php
+++ b/tests/Utils/SchemaPrinterTest.php
@@ -178,7 +178,7 @@ type Query {
         ];
         yield 'when deprecationReason is empty string' => [
             '',
-            '',
+            ' @deprecated',
         ];
         yield 'when deprecationReason is the default deprecation reason' => [
             Directive::DEFAULT_DEPRECATION_REASON,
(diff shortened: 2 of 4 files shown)


src/Utils/SchemaPrinter.php
```diff
@@ -396,7 +396,7 @@ private static function printDeprecated($fieldOrEnumVal) : string
         if (empty($reason)) {
```
simPod: This can be changed to 

```suggestion
        if ($reason === '' || $reason === null) {
```

tests/Utils/SchemaPrinterTest.php
```diff
@@ -148,6 +148,67 @@ public function testPrintNonNullArrayNonNullStringField() : void
...
+            '
+type Query {
+  singleField: Int'.$expectedDeprecationDirective.'
+}
+',
+            $output
+        );
+    }
+
+    public function deprecationReasonDataProvider(): \Generator
```
simPod: `deprecationReason` is typed as `string|null` so there's no need to test for other types like int or array.

Oliboy50: > /** @var string|null */

is "typed" 😄

simPod: Yes. We have static analysis and CI to guard that.

tests/Utils/SchemaPrinterTest.php
```diff
@@ -148,6 +149,47 @@ public function testPrintNonNullArrayNonNullStringField() : void
         );
     }
 
+    /**
+     * @see it('Prints Field With "@deprecated" Directive')
+     *
+     * @dataProvider deprecationReasonDataProvider
+     */
+    public function testPrintDeprecatedField($deprecationReason, string $expectedDeprecationDirective) : void
```
simPod:
```suggestion
    public function testPrintDeprecatedField(?string $deprecationReason, string $expectedDeprecationDirective) : void
```

tests/Utils/SchemaPrinterTest.php
```diff
@@ -148,6 +149,47 @@ public function testPrintNonNullArrayNonNullStringField() : void
...
+            '
+type Query {
+  singleField: Int' . $expectedDeprecationDirective . '
+}
+',
+            $output
+        );
+    }
+
+    public function deprecationReasonDataProvider() : Generator
```
simPod: Add return `Generator<string, array<mixed>>`

Oliboy50: when I add:

```php
    /**
     * @return Generator<string, array<mixed>>
     */
```

running `vendor/bin/phpcbf` removes it

simPod: Ayright, let it be for now then

src/Utils/SchemaPrinter.php
```diff
@@ -393,10 +393,10 @@ static function ($f, $i) use ($options) {
     private static function printDeprecated($fieldOrEnumVal) : string
     {
         $reason = $fieldOrEnumVal->deprecationReason;
-        if (empty($reason)) {
+        if ($reason === '' || $reason === null) {
```
vladar:
```suggestion
        if ($reason === null) {
```

src/Utils/SchemaPrinter.php
```diff
@@ -393,10 +393,10 @@ static function ($f, $i) use ($options) {
...
             return '';
         }
-        if ($reason === '' || $reason === Directive::DEFAULT_DEPRECATION_REASON) {
+        if ($reason === Directive::DEFAULT_DEPRECATION_REASON) {
```
vladar:
```suggestion
        if ($reason === '' || $reason === Directive::DEFAULT_DEPRECATION_REASON) {
```

Oliboy50:
Is there any else you want me to do, here? 🙏

vladar:
Great job! Let's ship it 🚢 And sorry for the delay.

PR #632, opened 2020-03-12
Fixed grouping implementors fields for abstract types in QueryPlan

yura3d:
This PR is for fix #513 by adding support of grouping implementors fields for abstract types in QueryPlan at any nesting level, not only at root.

### DIFF
--- a/src/Type/Definition/QueryPlan.php
+++ b/src/Type/Definition/QueryPlan.php
@@ -161,18 +161,20 @@ class QueryPlan
      *
      * @throws Error
      */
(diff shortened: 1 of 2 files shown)


vladar:
Makes sense, but before merging I need to understand if it break things for existing usages or not? If it does, we probably need a note in breaking changes.

vladar:
I've resolved conflicts and merged the PR manually. But the question still holds. If this PR breaks something for existing users, please add a note about this in Upgrade.md (in a new PR).

This change goes to a new major version, so breaking changes are expected but we still should at least warn users about those (if any).

And thank you for the PR! 👍

PR #637, opened 2020-03-20
Add support for __typename field in QueryPlan

simPod:


simPod:
ah, same as https://github.com/webonyx/graphql-php/pull/545

PR #642, opened 2020-04-13
Replace deprecated expectExceptionMessageRegExp with expectExceptionMessageMatches

spawnia:
This will work after we drop support for PHP `7.1`: https://github.com/webonyx/graphql-php/pull/470


spawnia:
We can add those changes in https://github.com/webonyx/graphql-php/pull/470.

Closing this to keep the repo clean.

PR #643, opened 2020-04-13
Support repeatable directives

### LINKED ISSUE #608: Support `repeatable` directives
The specification recently included a new keyword `repeatable` that allows specifying directives that may be defined multiple times.

https://github.com/graphql/graphql-spec/pull/472

Reading multiple directives already works fine, so that part should not be too hard. We will have to add support for the new keyword and adapt the validation steps to handle it.

Partial support was already added to the reference implementation: https://github.com/graphql/graphql-js/pull/1965

Full support is underway, too: https://github.com/graphql/graphql-js/pull/1541

spawnia:
Resolves #608 

See https://github.com/graphql/graphql-js/pull/1965

### DIFF
--- a/src/Validator/Rules/UniqueDirectivesPerLocation.php
+++ b/src/Validator/Rules/UniqueDirectivesPerLocation.php
@@ -5,13 +5,21 @@ declare(strict_types=1);
 namespace GraphQL\Validator\Rules;
 
 use GraphQL\Error\Error;
+use GraphQL\Language\AST\DirectiveDefinitionNode;
 use GraphQL\Language\AST\DirectiveNode;
 use GraphQL\Language\AST\Node;
+use GraphQL\Type\Definition\Directive;
 use GraphQL\Validator\ASTValidationContext;
 use GraphQL\Validator\SDLValidationContext;
 use GraphQL\Validator\ValidationContext;
 use function sprintf;
 
+/**
+ * Unique directive names per location
+ *
+ * A GraphQL document is only valid if all non-repeatable directives at
+ * a given location are uniquely named.
+ */
 class UniqueDirectivesPerLocation extends ValidationRule
 {
     public function getVisitor(ValidationContext $context)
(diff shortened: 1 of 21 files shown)


spawnia:
@simPod have you had a chance to review this PR yet?

simPod:
@spawnia I'll take a look but definitely don't feel competent to merge this after on my own, @vladar will have to do it :)

src/Validator/Rules/UniqueDirectivesPerLocation.php
```diff
@@ -26,16 +34,41 @@ public function getSDLVisitor(SDLValidationContext $context)
 
     public function getASTVisitor(ASTValidationContext $context)
     {
+        $uniqueDirectiveMap = [];
+
+        $schema            = $context->getSchema();
+        $definedDirectives = $schema !== null
```
simPod, line 40: Personally, I'd prefer not to use negation in conditions where eligible. Reduces cognitive load while reading it.

tests/Type/DirectiveTest.php
```diff
@@ -0,0 +1,15 @@
...
+declare(strict_types=1);
+
+namespace GraphQL\Tests\Type;
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @see describe('Type System: Directive', () => {
+ */
+class DirectiveTest extends TestCase
```
simPod, line 12:
```suggestion
final class DirectiveTest extends TestCase
```

src/Validator/Rules/UniqueDirectivesPerLocation.php
spawnia: Generally agree, although in this case `$schema` is usually non-null. It makes sense to have the happy path first and the fallback afterwards.

shmax: I will agree with whomever merges some of my PRs first, so I don't have to generate my basefiles again...

vladar:
@spawnia Is it ready for merging? (just to make sure)

spawnia:
> @spawnia Is it ready for merging? (just to make sure)

Yes. I just checked if there are breaking changes we might want to mention, haven't found any.

PR #644, opened 2020-04-15
Rule to validate subscriptions include one field

stayallive:
Subscriptions must only include one field.

A GraphQL subscription is valid only if it contains a single root field.

Spec: http://spec.graphql.org/June2018/#sec-Single-root-field
JS rule: https://github.com/graphql/graphql-js/blob/master/src/validation/rules/SingleFieldSubscriptionsRule.js

### DIFF
--- a/src/Validator/Rules/SingleFieldSubscription.php
+++ b/src/Validator/Rules/SingleFieldSubscription.php
@@ -17,7 +17,10 @@ use function sprintf;
 
 class SingleFieldSubscription extends ValidationRule
 {
-    public function getVisitor(ValidationContext $context)
+    /**
+     * @return array<string, callable>
+     */
+    public function getVisitor(ValidationContext $context) : array
     {
         return [
             NodeKind::OPERATION_DEFINITION => static function (OperationDefinitionNode $node) use ($context) : VisitorOperation {
@@ -43,7 +46,7 @@ class SingleFieldSubscription extends ValidationRule
         ];
     }
 
-    public static function multipleFieldsInOperation($operationName)
+    public static function multipleFieldsInOperation(?string $operationName) : string
     {
         if ($operationName === null) {
             return sprintf('Anonymous Subscription must select only one top level field.');
--- a/src/Validator/Rules/SingleFieldSubscription.php
+++ b/src/Validator/Rules/SingleFieldSubscription.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Validator\Rules;
+
+use GraphQL\Error\Error;
+use GraphQL\Language\AST\NodeKind;
+use GraphQL\Language\AST\NodeList;
+use GraphQL\Language\AST\OperationDefinitionNode;
+use GraphQL\Language\Visitor;
+use GraphQL\Language\VisitorOperation;
+use GraphQL\Validator\ValidationContext;
+use function array_splice;
+use function count;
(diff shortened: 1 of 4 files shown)


src/Validator/Rules/SingleFieldSubscription.php
```diff
@@ -0,0 +1,54 @@
...
+use GraphQL\Language\Visitor;
+use GraphQL\Language\VisitorOperation;
+use GraphQL\Validator\ValidationContext;
+use function array_splice;
+use function count;
+use function sprintf;
+
+class SingleFieldSubscription extends ValidationRule
+{
+    public function getVisitor(ValidationContext $context)
```
spawnia, line 20:
```suggestion
    /**
     * @return array<string, Closure>
     */
    public function getVisitor(ValidationContext $context): array
```

src/Validator/Rules/SingleFieldSubscription.php
```diff
@@ -0,0 +1,54 @@
...
+                        ));
+                    }
+                }
+
+                return Visitor::skipNode();
+            },
+        ];
+    }
+
+    public static function multipleFieldsInOperation($operationName)
```
spawnia, line 46:
```suggestion
    public static function multipleFieldsInOperation(?string $operationName): string
```

tests/Validator/SingleFieldSubscriptionsTest.php
```diff
@@ -0,0 +1,190 @@
...
+        third: catSubscribe {
+          meows
+        }
+      }
+        ',
+            [$this->multipleFieldsInOperation(null, [6, 9], [9, 9])]
+        );
+    }
+
+    private function multipleFieldsInOperation($operationName, ...$locations)
```
spawnia, line 179:
```suggestion
    /**
     * @param array<int, int> ...$locations A tuple of line and column
     */
    private function multipleFieldsInOperation(?string $operationName, array ...$locations): string
```

Not too sure if PHPStan will accept that.

src/Validator/Rules/SingleFieldSubscription.php
simPod: 👍 maybe even something like `* @return array<string, callable():VisitorOperation>`

spawnia: I did not know that PHPStan supports function types. Neat.

stayallive: phpcs does not like this one though...

```
FILE: src/Validator/Rules/SingleFieldSubscription.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 21 | ERROR | @return annotation of method
    |       | \GraphQL\Validator\Rules\SingleFieldSubscription::getVisitor()
    |       | does not specify type hint for items of its traversable
    |       | return value.
    |       | (SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingTraversableReturnTypeHintSpecification)
----------------------------------------------------------------------
```

Just this it does like (and same goes for phpstan):

```php
    /**
     * @return array<string, callable>
     */
```

spawnia: I think that's good enough.

simPod: ah right, that should be fixed in https://github.com/webonyx/graphql-php/pull/562 nvm

PR #646, opened 2020-04-18
Fix phpunit deprecation

shmax:
`expectExceptionMessageRegExp` was deprecated in PHPUnit 8.4, and will be removed in 9.0. Migration seems easy enough...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "psr/http-message": "^1.0",
     "react/promise": "2.*",
     "simpod/php-coveralls-mirror": "^3.0",
-    "squizlabs/php_codesniffer": "^3.5.2"
+    "squizlabs/php_codesniffer": "3.5.4"
   },
   "config": {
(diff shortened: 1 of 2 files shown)


shmax:
Ah, I see we support both phpunit 7.2 and 8.5. I guess we have to just sort of live with this for a while?

simPod:
TBH I'd rather wait with this for #470 and then solve it only for latest compatible phpunit version.

shmax:
> I'd rather wait with this for #470 and then solve it only for latest compatible phpunit version.

Sure, but there's no harm in merging, this, is there? Those warnings are a big distraction, and we can always touch this up if you ever sort out #470

simPod:
Seemed like a bit of pollution to me but ofc it depends on a roadmap. If @vladar does not plan to drop 7.1 support soon this can be merged I guess.

shmax:
Thanks much!

PR #647, opened 2020-04-19
Fix native return typehint errors

shmax:
This fixes about half of the "Anonymous function should have native return typehint" phpstan errors. If this PR goes through with no problem I'll tackle the rest in a second PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,8 +11,7 @@
   "require": {
     "php": "^7.1||^8.0",
     "ext-json": "*",
-    "ext-mbstring": "*",
-    "squizlabs/php_codesniffer": "3.5.4"
+    "ext-mbstring": "*"
   },
   "require-dev": {
     "amphp/amp": "^2.3",
@@ -25,7 +24,8 @@
     "phpunit/phpunit": "^7.2|^8.5",
     "psr/http-message": "^1.0",
     "react/promise": "2.*",
-    "simpod/php-coveralls-mirror": "^3.0"
+    "simpod/php-coveralls-mirror": "^3.0",
+    "squizlabs/php_codesniffer": "^3.5.4"
   },
   "config": {
     "preferred-install": "dist",
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,8 @@
   "require": {
     "php": "^7.1||^8.0",
     "ext-json": "*",
-    "ext-mbstring": "*"
+    "ext-mbstring": "*",
+    "squizlabs/php_codesniffer": "3.5.4"
   },
   "require-dev": {
     "amphp/amp": "^2.3",
@@ -24,8 +25,7 @@
     "phpunit/phpunit": "^7.2|^8.5",
     "psr/http-message": "^1.0",
     "react/promise": "2.*",
-    "simpod/php-coveralls-mirror": "^3.0",
-    "squizlabs/php_codesniffer": "^3.5.2"
+    "simpod/php-coveralls-mirror": "^3.0"
   },
   "config": {
     "preferred-install": "dist",
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "psr/http-message": "^1.0",
     "react/promise": "2.*",
     "simpod/php-coveralls-mirror": "^3.0",
-    "squizlabs/php_codesniffer": "^3.5.4"
+    "squizlabs/php_codesniffer": "3.5.4"
   },
   "config": {
     "preferred-install": "dist",
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "doctrine/coding-standard": "^6.0",
     "phpbench/phpbench": "^0.14",
     "phpstan/extension-installer": "^1.0",
-    "phpstan/phpstan": "0.12.11",
+    "phpstan/phpstan": "0.12.18",
     "phpstan/phpstan-phpunit": "0.12.6",
     "phpstan/phpstan-strict-rules": "0.12.2",
     "phpunit/phpunit": "^7.2|^8.5",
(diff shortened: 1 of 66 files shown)


src/Validator/Rules/ProvidedRequiredArguments.php
```diff
@@ -42,10 +43,12 @@ public function getVisitor(ValidationContext $context)
...
                     }
+
+                    return null;
                 },
             ],
             NodeKind::DIRECTIVE => [
-                'leave' => static function (DirectiveNode $directiveNode) use ($context) {
+                'leave' => static function (DirectiveNode $directiveNode) use ($context) : ?VisitorOperation {
                     $directiveDef = $context->getDirective();
                     if (! $directiveDef) {
```
simPod, line 53: I'd invert the condition

src/Validator/Rules/ProvidedRequiredArguments.php
```diff
@@ -19,7 +20,7 @@ public function getVisitor(ValidationContext $context)
     {
         return [
             NodeKind::FIELD     => [
-                'leave' => static function (FieldNode $fieldNode) use ($context) {
+                'leave' => static function (FieldNode $fieldNode) use ($context) : ?VisitorOperation {
                     $fieldDef = $context->getFieldDef();
 
                     if (! $fieldDef) {
```
simPod, line 26: I'd invert the condition

src/Validator/Rules/ProvidedRequiredArgumentsOnDirectives.php
```diff
@@ -85,7 +85,7 @@ static function (NamedTypeNode $argument) : string {
         }
 
         return [
-            NodeKind::DIRECTIVE => static function (DirectiveNode $directiveNode) use ($requiredArgsMap, $context) {
+            NodeKind::DIRECTIVE => static function (DirectiveNode $directiveNode) use ($requiredArgsMap, $context) : ?string {
                 $directiveName = $directiveNode->name->value;
                 $requiredArgs  = $requiredArgsMap[$directiveName] ?? null;
                 if (! $requiredArgs) {
```
simPod, line 91: I'd invert the condition

src/Validator/Rules/ValuesOfCorrectType.php
```diff
@@ -77,6 +78,8 @@ public function getVisitor(ValidationContext $context)
 
                     return Visitor::skipNode();
                 }
+
+                return null;
```
simPod, line 82: I'd invert the condition so the `skipNode()` is in the main path and `null` as early exit.

phpstan-baseline.neon
```diff
@@ -3,2230 +3,1690 @@ parameters:
 		-
 			message: "#^Only booleans are allowed in &&, GraphQL\\\\Language\\\\AST\\\\Node\\|null given on the left side\\.$#"
 			count: 1
-			path: src/Error/Error.php
+			path: src\Error\Error.php
```
simPod, line 6: Can you check why it regenerated backslashes? Does not seem correct 🤔Maybe bump phpstan if required.

src/Validator/Rules/ProvidedRequiredArguments.php
shmax: I don't disagree, but that sort of opens a can of worms and I think that kind of thing will be covered by other phpstan warnings which I'll tackle in a new PR; I'd rather focus on "return native typehint" errors in this PR if you don't mind.

src/Validator/Rules/ProvidedRequiredArgumentsOnDirectives.php
shmax: Will address in future PR.

src/Validator/Rules/ValuesOfCorrectType.php
shmax: Will address in future PR.

phpstan-baseline.neon
shmax: This was a flaw in phpstan innards, which was fixed in 0.12.18. I updated our lock to that version, made the necessary tweak to compose.json, and regenerated the file in 6bc5841. Thanks for catching that!

composer.json
```diff
@@ -54,7 +54,7 @@
     "lint" : "phpcs",
     "fix" : "phpcbf",
     "stan": "phpstan analyse --ansi --memory-limit 2048M",
-    "phpstan-baseline": "phpstan analyse --ansi --error-format baselineNeon > phpstan-baseline.neon",
+    "phpstan-baseline": "phpstan analyse --ansi --generate-baseline=phpstan-baseline.neon",
```
simPod, line 57:
```suggestion
    "phpstan-baseline": "phpstan analyse --ansi --generate-baseline",
```

src/Language/AST/NamedTypeNode.php
```diff
@@ -9,6 +9,6 @@ class NamedTypeNode extends Node implements TypeNode
     /** @var string */
     public $kind = NodeKind::NAMED_TYPE;
 
-    /** @var NameNode */
+    /** @var string */
```
simPod, line 12: Can you check if this is correct? There are some new errors in baseline I guess should not be there like

```
"#^Cannot access property \\$value on string\\.$#"
src/Validator/Rules/ProvidedRequiredArgumentsOnDirectives.php
```

composer.json
shmax: That works. d873515

src/Language/AST/NamedTypeNode.php
shmax: You're right (it can be either NameNode or string). Fixed in a702169

shmax:
Thanks much for reviewing 👍

PR #648, opened 2020-04-23
Fix construct empty phpstan violations

shmax:
Fix all `#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#` phpstan errors.


src/Executor/ExecutionResult.php
```diff
@@ -153,7 +154,7 @@ public function toArray($debug = false)
             $result['data'] = $this->data;
         }
 
-        if (! empty($this->extensions)) {
+        if (count($this->extensions ?? []) > 0) {
```
simPod, line 157: Sometimes `!== 0` is used and sometimes `> 0`. Is there a reason?

tests/StarWarsValidationTest.php
```diff
@@ -37,7 +38,7 @@ public function testValidatesAComplexButValidQuery() : void
         }
       ';
         $errors = $this->validationErrors($query);
-        self::assertEquals(true, empty($errors));
+        self::assertEquals(true, count($errors) === 0);
```
simPod, line 41: We could replace all these with proper assertions like assertCount

src/Executor/ExecutionResult.php
shmax: No, just changed my mind midway through. Fixed in 44d6d20

tests/StarWarsValidationTest.php
shmax: SGTM. 6b6e3ad

phpstan-baseline.neon
```diff
@@ -3,2200 +3,1990 @@ parameters:
 		-
 			message: "#^Only booleans are allowed in &&, GraphQL\\\\Language\\\\AST\\\\Node\\|null given on the left side\\.$#"
 			count: 1
-			path: src/Error/Error.php
+			path: src\Error\Error.php
```
simPod, line 6: This slipped in again :P

shmax: Whoopsie. Fixed.

shmax:
@vladar 
Merged master. Please review and merge when you have a chance.

vladar:
The general problem I have with changes like this is that it replaces operation (`empty`) with a function call (`count`). Do we really need to count in most of these cases? Nope.

This makes things a little bit slower. And while I understand that this is a micro-optimization but libraries like this should probably care about it.

Did you have a chance to run before and after benchmarks on this? Curious to see the numbers. If the difference is neglectable on those synthetic benchmarks then it should be good to go.

shmax:
Eh, you're right. It didn't occur to me that there might be a performance impact. I'll see if I can come up with some other way to appease 'ol Stan, but in the meantime, what would you consider to be a valid benchmark sample? Would running the unit tests be enough?

simPod:
`count()` is built-in-optimized as ZEND_COUNT opcode so should no be a big deal.

Though I benchmarked `count($arr) ===0` vs `$arr === []` few weeks ago and the second was a bit faster.

vladar:
@shmax We have benchmarks folder (it is probably too naive though). Probably the bench in `HugeSchemaBench` could be somewhat useful.

I usually run it like this:

```
vendor\bin\phpbench run --report=aggregate benchmarks\HugeSchemaBench.php
```

shmax:
AH, very helpful, thanks.

shmax:
> I usually run it like this:
> 
> ```
> vendor\bin\phpbench run --report=aggregate benchmarks\HugeSchemaBench.php
> ```

Has it by any chance been a while since anyone has run this? When I use the command you suggested, I get:

```sh
Notice: fgets(): read of 8192 bytes failed with errno=9 Bad file descriptor
```

And if I run the command from contributing.md:

```sh
[Semantical Error] The annotation "@include" in subject \GraphQL\Executor\ReferenceExecutor::shouldIncludeNode was never imported. Did you maybe forget to add a "use" statement for this annotation?
```

shmax:
I ran the benchmark suite through blackfire.
master:
https://blackfire.io/profiles/a4ce5b2e-f9bb-4155-a959-c7ffbcbd6212/graph

this branch:
https://blackfire.io/profiles/accaa005-92f6-45d3-983b-4fbdc199aae1/graph

I see no degradation.

vladar:
Cool. Thanks, @shmax for taking the time to do this. And yeah, @simPod is probably correct - `count` just returns a value and doesn't loop through the whole thing (as happens for pure `C` arrays/strings)

shmax:
Thanks very much for reviewing 👍

PR #650, opened 2020-04-24
Fix NodeList::getFieldSelection() return type

shmax:
Correcting the return type of `ResolveInfo.getFieldSelection`.

Not a big deal, but it was causing stanning errors in my app code.

Also fixed one of the new contravariant errors that was introduced in a recent version of PHPStan.

### DIFF
--- a/src/Language/AST/NodeList.php
+++ b/src/Language/AST/NodeList.php
@@ -83,8 +83,8 @@ class NodeList implements ArrayAccess, IteratorAggregate, Countable
     }
 
     /**
-     * @param int|string   $offset
-     * @param Node|mixed[] $value
+     * @param int|string|null $offset
+     * @param Node|mixed[]    $value
      *
      * @phpstan-param T|mixed[] $value
      */
--- a/src/Language/AST/NodeList.php
+++ b/src/Language/AST/NodeList.php
@@ -83,8 +83,8 @@ class NodeList implements ArrayAccess, IteratorAggregate, Countable
     }
 
     /**
-     * @param int|string|null $offset
-     * @param Node|mixed[]    $value
+     * @param int|string   $offset
+     * @param Node|mixed[] $value
      *
      * @phpstan-param T|mixed[] $value
      */
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "psr/http-message": "^1.0",
     "react/promise": "2.*",
     "simpod/php-coveralls-mirror": "^3.0",
-    "squizlabs/php_codesniffer": "^3.5.2"
+    "squizlabs/php_codesniffer": "3.5.4"
   },
   "config": {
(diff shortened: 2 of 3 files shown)


src/Language/AST/NodeList.php
```diff
@@ -83,8 +83,8 @@ public function offsetGet($offset)// : Node
     }
 
     /**
-     * @param int|string   $offset
-     * @param Node|mixed[] $value
+     * @param int|string|null $offset
```
simPod, line 86: I believe `$offset` cannot be null

simPod: Also the case for previous PRs

shmax: Are you sure? PHPStan seems to disagree... https://github.com/phpstan/phpstan-src/blob/ab851203d475b0b141817912fbb0955eaa7803d3/stubs/ArrayObject.stub#L23

simPod: Ah, that's `ArrayObject`... mkay. Therefore, I think we have a bug here https://github.com/webonyx/graphql-php/pull/650/files#diff-d22cf662ac36edb6150cfc3e80fdfbc0R107 as `null` can't be used for array index directly. 

Though I would solve it in separate PR, you have this change in all your recent PRs. Does phpstan complain about it?

shmax: > as null can't be used for array index directly.

It can.
According to the docs:
https://www.php.net/manual/en/language.types.array.php

> Null will be cast to the empty string, i.e. the key null will actually be stored under "".

It's weird, but I don't think we need to worry about it, now. 

> you have this change in all your recent PRs. Does phpstan complain about it?

Well, yes. If you prefer I could add the warning to the baseline file and leave the annotation as-is.

simPod: > Null will be cast to the empty string, i.e. the key null will actually be stored under "".

Umm, black magic...

Any idea why it did not complain here? https://github.com/webonyx/graphql-php/commit/a2780a0da01f6f106940d82e8ddff51314afd697 It's recent commit, the build passed though.

shmax: Not sure. I'll see if I can figure it out...

shmax: In my other PRs I lock phpstan to 0.12.18, so that's what I still have on local. In
a2780a0 you're still locked to 3.5.4. Tell you what, since it's bothering you, I'll go ahead and remove this bit from this PR so we can move forward, and we can deal with it some other time.

simPod: Ah cool. This should be done in PR related to phpstan upgrade. That's why I've locked phpstan to concrete version to prevent irrelevant build breakage.

shmax:
Thanks much!

PR #651, opened 2020-04-24
Use psr7 request body instead of parsedBody

mwijngaard:
According to the PSR-7 spec, the behavior of ServerRequestInterface::parsedBody is only really specified for multipart/form-data and application/x-www-form-urlencoded requests (in which case it should contain the contents of $_POST). With other request types, the parsedBody MAY return an unserialized version of the body, but since it's not really practical to build parsers for all known formats into a http library it should be perfectly acceptable to return nothing. This PR moves the parsing of the json body into parsePsrRequest.

### DIFF
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -535,13 +535,11 @@ class Helper
             if (stripos($contentType[0], 'application/graphql') !== false) {
                 $bodyParams = ['query' => $request->getBody()->getContents()];
             } elseif (stripos($contentType[0], 'application/json') !== false) {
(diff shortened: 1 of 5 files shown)


simPod:
@mwijngaard mind checking if https://github.com/webonyx/graphql-php/pull/598 does not replace this already?

mwijngaard:
@simPod Yes it would seem so :-). I noticed that after i submitted the PR. But I left mine open as your PR has been open for a long time and this fix seems more trivial so I was hoping it would be merged a little quicker.

vladar:
I am going to close this as #598 is merged now

PR #652, opened 2020-04-26
Fix return type for ResolveInfo.getFieldSelection

shmax:
Er, sorry--my previous fix for this was not correct. `ResolveInfo.getFieldSelection` can actually be a deeply nested structure. We apparently don't yet have the ability to express [recursive types](https://github.com/phpstan/phpdoc-parser/issues/9), yet, so `array<string, mixed>` may be the best we can do.


shmax:
Thanks much! 👍

PR #653, opened 2020-04-26
Make type hint for OperationDefinitionNode#name nullable

spawnia:
See https://github.com/webonyx/graphql-php/blob/master/src/Language/Parser.php#L505


spawnia:
Those PHPStan baseline errors are giving me a headache. What's the recommended workflow for fixing them?

simPod:
New errors should be fixed and after that baseline needs to be regenerated. 

Maybe I'd fix those errors so they can be removed from baseline completely?

PR #656, opened 2020-04-29
Fix native return typehint pt ii

shmax:
As promised, this is the second half of the "Anonymous function should have native return typehint..." phpstan fixes.


simPod:
I wonder whether we can also drop this now https://github.com/webonyx/graphql-php/blob/master/phpcs.xml.dist#L19?

shmax:
>  wonder whether we can also drop this now /phpcs.xml.dist@master#L19 ?

I don't know what that does--even with it enabled I was getting the errors. But I removed it, and everything still seems to check out. 🤷

shmax:
@simPod Thanks for reviewing 👍

PR #658, opened 2020-05-09
Allow `DocumentNode` in `OperationParams#query`

esamattis:
In WPGraphQL we want to parse the GraphQL queries to `DocumentNode` by ourselves before passing it to the graphql-php Server instance via `OperationParams`. This would allow us to add an filter on which userland plugins can hook into for AST inspection, transformation and [caching](https://github.com/valu-digital/wp-graphql-cache).

But unfortunately the current `validateOperationParams` in the Helper class does not allow it to be a `DocumentNode`.

But later in the file the it can skip parsing if it is already a `DocumentNode`:

https://github.com/webonyx/graphql-php/blob/1f0ec2408010e01879b1f980afac65de2344c631/src/Server/Helper.php#L279-L285

Here's the WPGraphQL PR for the filters https://github.com/wp-graphql/wp-graphql/pull/1302

I realize this implementation might not be ideal since `OperationParams` is mean to represent "HTTP parameters" where AST does not necessarily make sense but I could not figure any other way to do this without larger refactoring. Other ideas would be very welcome.

Thanks!

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -307,11 +307,16 @@ parameters:
 
 		-
 			message: "#^Only booleans are allowed in a negated boolean, string given\\.$#"
+			count: 1
+			path: src/Server/Helper.php
+
+		-
+			message: "#^Only booleans are allowed in a negated boolean, GraphQL\\\\Language\\\\AST\\\\DocumentNode|string given\\.$#"
 			count: 2
 			path: src/Server/Helper.php
 
 		-
-			message: "#^Only booleans are allowed in &&, string given on the left side\\.$#"
+			message: "#^Only booleans are allowed in &&, GraphQL\\\\Language\\\\AST\\\\DocumentNode|string given on the left side\\.$#"
 			count: 1
 			path: src/Server/Helper.php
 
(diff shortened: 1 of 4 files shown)


esamattis:
To be honest I don't understand why the PHPStan check started failing with

> Only booleans are allowed in a negated boolean, GraphQL\Language\AST\DocumentNode|string given

It wasn't a boolean to begin with just made it into a union 🤔

simPod:
@esamattis it's because the error message changed for already ignored errors.

esamattis:
Oh right, makes sense! Thanks @simPod! Updated the ignore file.

vladar:
Sorry for the late reply.

On this PR. This doesn't sound right to me. We already have persisted query support for cases like this via [`persistentQueryLoader` option](https://webonyx.github.io/graphql-php/executing-queries/#using-server) of the server. So what you can do instead:

```php
$server = new StandardServer([
    'persistentQueryLoader' => function($queryId, OperationParams $params) {
        return $whateverDocumentNodeYouWant;
    }
]);
```

It should work for you without any changes. Thoughts?

esamattis:
Ah! It probably can be used for this.

> A function which is called to fetch actual query **when server encounters queryId** in request vs query.

I guess we'll have add some kind of dummy queryId id to actually trigger it but it's totally doable. I'll experiment with it soon.

vladar:
@esamattis Did it work for you? If so, can we close this PR?

esamattis:
On a vacation ATM. I but I should be able to check this during next week when I return.

spawnia:
Closing for now, feel free to reopen if your issue persists.

PR #659, opened 2020-05-17
Fix "only booleans are allowed" errors

shmax:
Fixing more PHPStan errors, specifically ones related to "truthy" checks. There are still 130 of these to go, so I figured this time we could break it up into several PRs.


shmax:
@simPod @vladar Now we're cookin'. Thanks for reviewing/merging.

PR #670, opened 2020-06-04
Drop Error::$message

simPod:
It just overwrites parent's `Exception::$message` and is not initialised in constructor. I believe it can be dropped.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,7 @@
 ## Master
 
+- Dropped `GraphQL\Error\Error::$message`
+
 ### Breaking (major): dropped deprecations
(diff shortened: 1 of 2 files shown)


spawnia:
I agree that we should make this change, but it is actually a breaking change. `Exception::$message` is `protected`, it is overwritten to be public.

simPod:
I'll edit changelog 👍🏿

vladar:
Yup, this is a breaking change

UPGRADE.md
```diff
@@ -1,5 +1,7 @@
 ## Master
 
+- Dropped `GraphQL\Error\Error::$message`
```
spawnia, line 3:
```suggestion
- Dropped `GraphQL\Error\Error::$message`, use `getMessage()` instead.
```

PR #675, opened 2020-06-16
Switch to null-safe lazy initialization

spawnia:
Let's continue the discussion we are having here: https://github.com/webonyx/graphql-php/pull/557#discussion_r437184406

@simPod @shmax


src/Type/Definition/InputObjectType.php
```diff
@@ -45,16 +48,12 @@ public function __construct(array $config)
...
-     *
-     * @throws Exception
+     * @throws InvariantViolation
      */
-    public function getField($name)
+    public function getField(string $name) : InputObjectField
     {
-        if ($this->fields === null) {
-            $this->getFields();
+        if (! isset($this->fields)) {
```
spawnia, line 55: Those checks are pretty repetitive, we might move them into `initializeFields()`. Not sure if the cost of always calling a method is significant and in a hot path.

src/Error/Error.php
```diff
@@ -203,10 +207,7 @@ public function getCategory()
         return $this->category;
     }
 
-    /**
-     * @return Source|null
-     */
-    public function getSource()
+    public function getSource() : ?Source
```
shmax, line 210: 👍

src/Type/Definition/EnumType.php
```diff
@@ -88,9 +94,9 @@ private function getNameLookup() : ArrayObject
     /**
      * @return EnumValueDefinition[]
      */
-    public function getValues()
+    public function getValues() : array
     {
-        if ($this->values === null) {
+        if (! isset($this->values)) {
```
shmax, line 99: Could `values` be initialized to an empty array at declaration time? Or can it legitimately be an empty array even after lazy load?

spawnia: If it were to become an empty array after initialization, and we check for `count($this->values) === 0`, we would re-initialize unnecessarily. I think such an Enum type would be invalid, but i see nothing that stops a `$config` with empty `values` being passed.

The way it is written now clearly expresses that we run initialize `$this->values` only if it is not set, no matter what it is set to afterwards.

shmax: > If it were to become an empty array after initialization

That's exactly what I'm asking. If that is a possibility, then what you have now is great. But if empty array is impossible post-lazy load, then you could try the other way. But it sounds like we're not exactly sure, so I suppose it's safest to go with what you have.

vladar:
LGTM 👍 Mind resolving conflicts? Will merge it then.

src/Type/Definition/InputObjectType.php
spawnia: @vladar what is your opinion here?

vladar: I would keep it as is.

spawnia:
@vladar ready to go.

PR #676, opened 2020-06-16
Do not use call_user_func or call_user_func_array

spawnia:
Fewer concepts results in simpler code.

Performance is also better, Symfony made this change too. The only reason for keeping `call_user_func_*` is PHP 5.5 compatibility, which we fortunately don't need.


spawnia:
@vladar ready and working. I don't quite get why, but [https://github.com/webonyx/graphql-php/pull/676/commits/5056e4ef84232b725db27a1a61ef1b1811f54670 ](https://github.com/webonyx/graphql-php/pull/676/commits/5056e4ef84232b725db27a1a61ef1b1811f54670) was the culprit. Seems like the callable can return `null` for some reason?

PR #679, opened 2020-06-23
Argument defaults with integer/float values crashes introspection query

m4rw3r:
Using integer or float values as the default values of a field argument causes the introspection query to crash with an error along the lines of `"Return value of GraphQL\Language\Printer::GraphQL\Language\{closure}() must be of the type string, int returned"`.

This seems to be caused by `GraphQL\Language\Printer::printAST` since it is supposed to return a string at all times, which is not the case when it encounters an `IntValueNode` or `FloatValueNode` from a default value.

I have provided test-cases and an attempted fix.

### DIFF
--- a/src/Language/AST/FloatValueNode.php
+++ b/src/Language/AST/FloatValueNode.php
@@ -9,6 +9,6 @@ class FloatValueNode extends Node implements ValueNode
     /** @var string */
     public $kind = NodeKind::FLOAT;
 
-    /** @var string */
+    /** @var float */
     public $value;
 }
--- a/src/Language/AST/FloatValueNode.php
+++ b/src/Language/AST/FloatValueNode.php
@@ -9,6 +9,6 @@ class FloatValueNode extends Node implements ValueNode
     /** @var string */
     public $kind = NodeKind::FLOAT;
 
-    /** @var float */
+    /** @var string */
     public $value;
 }
--- a/src/Language/Printer.php
+++ b/src/Language/Printer.php
@@ -181,12 +181,12 @@ class Printer
                             . $node->selectionSet;
                     },
 
-                    NodeKind::INT => static function (IntValueNode $node) : string {
-                        return (string) $node->value;
+                    NodeKind::INT => static function (IntValueNode $node) {
+                        return $node->value;
                     },
 
                     NodeKind::FLOAT => static function (FloatValueNode $node) : string {
-                        return (string) $node->value;
+                        return $node->value;
                     },
 
                     NodeKind::STRING => function (StringValueNode $node, $key) {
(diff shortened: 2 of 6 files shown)


src/Language/AST/FloatValueNode.php
```diff
@@ -9,6 +9,6 @@ class FloatValueNode extends Node implements ValueNode
     /** @var string */
     public $kind = NodeKind::FLOAT;
 
-    /** @var string */
+    /** @var float */
```
vladar, line 12: Please revert. This should still be a `string`.

src/Language/Printer.php
```diff
@@ -181,12 +181,12 @@ public function printAST($ast)
                             . $node->selectionSet;
                     },
 
-                    NodeKind::INT => static function (IntValueNode $node) {
-                        return $node->value;
+                    NodeKind::INT => static function (IntValueNode $node) : string {
+                        return (string) $node->value;
```
vladar, line 185: `$node->value` must always be a string. So no need to cast in this case - the error is actually here:

https://github.com/webonyx/graphql-php/blob/2dcc81e4d7f01470e66c72a2a1db8050c8b0b70a/src/Utils/AST.php#L254-L268

Instead, we must cast `$serialized` to string when creating all of those nodes.

src/Language/Printer.php
```diff
@@ -181,12 +181,12 @@ public function printAST($ast)
...
                     },
 
                     NodeKind::FLOAT => static function (FloatValueNode $node) : string {
-                        return $node->value;
+                        return (string) $node->value;
```
vladar, line 189: Also, please revert, see the comment above.

tests/Utils/AstFromValueTest.php
```diff
@@ -88,7 +88,7 @@ public function testConvertsFloatValuesToIntOrFloatASTs() : void
         self::assertEquals(new IntValueNode(['value' => '123']), AST::astFromValue(123.0, Type::float()));
         self::assertEquals(new FloatValueNode(['value' => '123.5']), AST::astFromValue(123.5, Type::float()));
         self::assertEquals(new IntValueNode(['value' => '10000']), AST::astFromValue(1e4, Type::float()));
-        self::assertEquals(new FloatValueNode(['value' => '1e+40']), AST::astFromValue(1e40, Type::float()));
+        self::assertEquals(new FloatValueNode(['value' => '1.0E+40']), AST::astFromValue(1e40, Type::float()));
```
vladar, line 91: What happens with the original format? I mean why do we have this change?

m4rw3r: `GraphQL\Type\Definition\FloatType::serialize` casts to `float` which destroys any original format, and then it is re-cast back to a string using PHPs default float format which is `1.0E+40`.

PR #682, opened 2020-06-26
update phpbench

shmax:
Bumping the minimum required version of phpbench to get a [fix](https://github.com/phpbench/phpbench/commit/cc354c33093581b22b8a5a05d5e4397d4a01b179#diff-384c8ba7168bc7b228a1b08c7de98d39) for an issue that was causing the benchmark script to fail in my PHP 7.4 environment.

Also, for some reason the syntax listed on `contributing.md` doesn't work for me, so I've updated it to what does work. If I'm just not understanding something simple please let me know.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,5 +51,5 @@ Based on [PHPStan](https://github.com/phpstan/phpstan).
 ## Running benchmarks
 Benchmarks are run via [PHPBench](https://github.com/phpbench/phpbench).
 ```sh
-./vendor/bin/phpbench run .
+./vendor/bin/phpbench run benchmarks
 ```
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -51,5 +51,5 @@ Based on [PHPStan](https://github.com/phpstan/phpstan).
 ## Running benchmarks
 Benchmarks are run via [PHPBench](https://github.com/phpbench/phpbench).
 ```sh
-./vendor/bin/phpbench run .
+./vendor/bin/phpbench run benchmarks
```
spawnia, line 54: Have you tried `vendor/bin/phpbench run`?

vladar: I just tried and it fails for me for some reason indeed. While `./vendor/bin/phpbench run benchmarks` works.

shmax: Ah, that works, too. Should I open another PR?

shmax:
Thanks for the quick review!

Issue #685, opened 2020-06-27
Invalid AST Node: false


shmax:
@vladar 
While working on optimizations, I updated my local code to master, and am now seeing this error whenever I use an `include` directive:

```
{
  "error": {
    "message": "Invalid AST Node: false"
  }
}
```

It's easy to reproduce. You can run the 02-blog sample, and run this query:
```
{
  viewer {
    id @include (if: true)
  }
}
```

The culprit seems to be this commit: cab159942136de30f105f8ea64f3124f600c90ed

Somehow, the `false` value returned [here](https://github.com/webonyx/graphql-php/commit/cab159942136de30f105f8ea64f3124f600c90ed#diff-b44685a5d7938a2da6b80af78c3780beR109) becomes the value of `node` and trips an exception [here](https://github.com/webonyx/graphql-php/blob/master/src/Language/Visitor.php#L275).

If I remove the boolean returns then everything starts working again, but then some of the tests fail. I thought I'd throw an issue up here to see if you have any thoughts. Thanks much.

PR #686, opened 2020-06-28
check for false node  in visit

shmax:
Fix for #685 

A recent [commit](cab1599) introduced a flaw in directive parsing. It seems to have been drawn from the js library, but there's a check for [false](https://github.com/graphql/graphql-js/blob/master/src/language/visitor.js#L308) that apparently didn't get brought over. 

Note: I tried to re-port the entire block in the surrounding code, but the tests blew up. Some day we should re-port the entire visit function, as it has diverged quite a bit.

### DIFF
--- a/src/Language/Visitor.php
+++ b/src/Language/Visitor.php
@@ -277,11 +277,16 @@ class Visitor
 
                 $visitFn = self::getVisitFn($visitor, $node->kind, $isLeaving);
 
-                if ($visitFn) {
+                if ($visitFn !== null) {
                     $result    = $visitFn($node, $key, $parent, $path, $ancestors);
                     $editValue = null;
 
-                    if ($result !== null) {
+                    if ($result === false) {
+                        if (! $isLeaving) {
+                            array_pop($path);
+                            continue;
+                        }
+                    } elseif ($result !== null) {
                         if ($result instanceof VisitorOperation) {
                             if ($result->doBreak) {
                                 break;
--- a/src/Language/Visitor.php
+++ b/src/Language/Visitor.php
@@ -281,12 +281,7 @@ class Visitor
                     $result    = $visitFn($node, $key, $parent, $path, $ancestors);
                     $editValue = null;
 
-                    if ($result === false) {
-                        if (! $isLeaving) {
-                            array_pop($path);
-                            continue;
-                        }
-                    } elseif ($result !== null) {
+                    if ($result !== null) {
                         if ($result instanceof VisitorOperation) {
                             if ($result->doBreak) {
                                 break;
--- a/src/Language/Visitor.php
+++ b/src/Language/Visitor.php
@@ -417,23 +417,18 @@ class Visitor
 
                     $result = $fn(...func_get_args());
 
-                    if (! ($result instanceof VisitorOperation)) {
-                        if ($result !== null) {
+                    if ($result instanceof VisitorOperation) {
+                        if ($result->doContinue) {
+                            $skipping[$i] = $node;
+                        } elseif ($result->doBreak) {
+                            $skipping[$i] = $result;
+                        } elseif ($result->removeNode) {
                             return $result;
                         }
-                        continue;
-                    }
-
-                    if ($result->doContinue) {
-                        $skipping[$i] = $node;
-                    } elseif ($result->doBreak) {
-                        $skipping[$i] = $result;
-                    } elseif ($result->removeNode) {
+                    } elseif ($result !== null) {
                         return $result;
                     }
                 }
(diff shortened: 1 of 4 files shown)


shmax:
> Can you check if it works for you?

b110268

It does indeed work, thank you, but I'm a bit puzzled by a few things: 
1. I don't see a KnownArgumentNamesOnDirectives rule in the reference library code
2. All execution paths in `KnownArgumentNamesOnDirectives.getASTVisitor` result in a skip. Is that expected?
3. The reference library _does_ have that check for false. Is it important?

shmax:
> If you can dig up an accompanying test case

There are a few test cases that look relevant, but for some reason it doesn't trigger the bug: https://github.com/webonyx/graphql-php/blob/master/tests/Executor/DirectivesTest.php#L69

I'll see if I can figure it out...

vladar:
> I don't see a KnownArgumentNamesOnDirectives rule in the reference library code

It's inside `KnownArgumentNames` there

shmax:
> It's inside `KnownArgumentNames` there

Ah, so it is. Okay. I don't see any "skip" logic in there, but if you're satisfied, I'm satisfied. Please merge at your convenience.

shmax:
Ack, coverage decreased...

vladar:
> Ah, so it is. Okay. I don't see any "skip" logic in there, but if you're satisfied, I'm satisfied. Please merge at your convenience.

They use `return false` for skip logic here:

https://github.com/graphql/graphql-js/blob/1e1d75e9b7914abb66d91bfa0ae1de8612674628/src/validation/rules/KnownArgumentNamesRule.js#L95

In php library we introduced explicit `Visitor::skipNode()` (and several other static methods) so we deviate a bit from the reference implementation here (and also that's the reason for this error - I just forgot about this, although surprised that tests didn't catch it)

shmax:
> Ack, coverage decreased...

The skip change seems to have resulted in some dead code, which I have removed. 4722ce7

shmax:
> They use return false for skip logic here

Gotcha. That works for me, thanks for the background.

src/Language/Visitor.php
```diff
@@ -417,18 +417,20 @@ public static function visitInParallel($visitors)
...
+                    if (! ($result instanceof VisitorOperation)) {
+                        continue;
+                    }
+
+                    if ($result->doContinue) {
+                        $skipping[$i] = $node;
+                    } elseif ($result->doBreak) {
+                        $skipping[$i] = $result;
+                    } elseif ($result->removeNode) {
                         return $result;
```
vladar, line 429: Looks like we miss this branch now? When the result is not an instance of `VisitorOperation` we must return it

shmax: Sorry, what is your suggestion?

shmax: I think I understand what you mean. With your skip change, that bit of code was being flagged as dead by the coverage tests, so I removed it.

vladar: GitHub shows this comment against new code but I've actually left it against the old code. So I mean this part:

```php
elseif ($result !== null) {
    return $result;
}
```

is missing now. We must restore it as it is basically how you "edit" an AST node with a parallel visitor.

shmax: Restored in 11b59cd. Are you sure this is necessary? I only ask because the tests passed without it.

vladar: Absolutely. Probably missing tests for this (we don't use a parallel visitor to modify nodes in this library, yet this feature is indispensable for AST manipulations)

vladar: Actually, can we revert this change completely (except for `$visitFn !== null` part)? Doesn't look more readable to me after the change.

shmax: Wait, what? Revert what change? The one I just made?

vladar: I mean all changes in this file except for `if ($visitFn !== null) {` line. Those changes don't really change anything now that we restored `return $result`. Or am I missing something?

shmax: Sure, you got it 7e680fd

shmax:
It seems coverage has decreased a bit:

https://coveralls.io/builds/31804139/source?filename=src%2FLanguage%2FVisitor.php#L429

@vladar  Can I ask your assistance in doing whatever it takes to get this PR merged? Days are rolling by, and this problem is blocking other developers.

shmax:
Thanks @vladar and @spawnia for reviewing!

vladar:
Published in `v14.0.1`

PR #687, opened 2020-06-29
Revert "move execution of lazy type to typeloader"

shmax:
This reverts commit 484d2dd188fa0ac9267f3a59a62f9b6c60a487d4, the change we discussed that moves responsibility for loading types from client typeloader back to schema.

I'm still sort of not sure, one way or another. I had hoped I would develop a strong opinion after doing the performance work, but still nothing. 🤷 

Do you guys have any insights?

### DIFF
--- a/src/Type/Schema.php
+++ b/src/Type/Schema.php
@@ -341,6 +341,10 @@ class Schema
 
         $type = $typeLoader($typeName);
 
+        if (is_callable($type)) {
+            $type = $type();
+        }
+
         if (! $type instanceof Type) {
             throw new InvariantViolation(
                 sprintf(
--- a/tests/Type/LazyTypeLoaderTest.php
+++ b/tests/Type/LazyTypeLoaderTest.php
@@ -6,7 +6,6 @@ namespace GraphQL\Tests\Type;
 
 use Exception;
 use GraphQL\Error\InvariantViolation;
-use GraphQL\Exception\InvalidArgument;
 use GraphQL\Tests\PHPUnit\ArraySubsetAsserts;
 use GraphQL\Type\Definition\InputObjectType;
 use GraphQL\Type\Definition\InterfaceType;
@@ -43,7 +42,7 @@ final class LazyTypeLoaderTest extends TestCase
     /** @var callable */
     private $postStoryMutationInput;
 
-    /** @var callable(string $name):Type */
+    /** @var callable */
     private $typeLoader;
 
     /** @var string[] */
@@ -174,24 +173,24 @@ final class LazyTypeLoaderTest extends TestCase
             },
         ]);
 
-        $this->typeLoader = function (string $name) : Type {
+        $this->typeLoader = function (string $name) : ?callable {
             $this->calls[] = $name;
             $prop          = lcfirst($name);
 
             switch ($prop) {
                 case 'node':
-                    return ($this->node)();
+                    return $this->node;
                 case 'blogStory':
-                    return ($this->blogStory)();
+                    return $this->blogStory;
                 case 'content':
-                    return ($this->content)();
+                    return $this->content;
                 case 'postStoryMutation':
-                    return ($this->postStoryMutation)();
+                    return $this->postStoryMutation;
                 case 'postStoryMutationInput':
-                    return ($this->postStoryMutationInput)();
+                    return $this->postStoryMutationInput;
             }
 
-            throw new InvalidArgument('Unknown type');
+            return null;
         };
     }
 
--- a/src/Type/SchemaConfig.php
+++ b/src/Type/SchemaConfig.php
@@ -42,7 +42,7 @@ class SchemaConfig
     /** @var Directive[]|null */
     public $directives;
 
-    /** @var callable(string $name):Type|null */
+    /** @var callable|null */
     public $typeLoader;
 
     /** @var SchemaDefinitionNode|null */
@@ -253,7 +253,7 @@ class SchemaConfig
     }
 
     /**
-     * @return callable(string $name):Type|null
+     * @return callable|null
      *
      * @api
      */


tests/Type/LazyTypeLoaderTest.php
```diff
@@ -43,7 +42,7 @@ final class LazyTypeLoaderTest extends TestCase
     /** @var callable */
     private $postStoryMutationInput;
 
-    /** @var callable(string $name):Type */
+    /** @var callable */
```
simPod, line 45:
```suggestion
    /** @var callable(string):?callable */
```

shmax: SGTM, thanks

spawnia:
So, theoretically the performance cost would be an extra `is_callable()` check once for every lazy-loaded type?

shmax:
master:
https://blackfire.io/profiles/796a7d37-5f99-42f1-b998-f350da60fea0/graph

this branch:
https://blackfire.io/profiles/03622834-6e0a-4985-af11-202c2fdbb066/graph

But that's not particularly telling, as the benchmark suite doesn't seem to exercise the typeloader. Here's what happens if I run the unit tests:

master:
https://blackfire.io/profiles/72d37075-8502-4527-af66-a63f1d53a8df/graph

this branch:
https://blackfire.io/profiles/18e3b65c-0b84-47f5-9f81-c1456d2d76fb/graph

Execution times seem the same, so I think it just comes down to preference. Semantically, I'm kind of starting to lean towards not merging this, since one could argue that the typeloader's job is to, well, load types, so it make sense for it to have the responsibility. On the other hand, if we leave the responsibility to the library, we keep the details encapsulated and leave the door open for future optimizations, so I'm torn.

spawnia:
I can not view those blackfire profiles.

shmax:
> I can not view those blackfire profiles, after login it shows me a 404.

Sorry about that. Made 'em public.

shmax:
I can do this...

```php
if ($type !== null && ! $type instanceof Type) {
  $type = $type();
}
```

...if we throw out the tests that exercise returning `stdClass` and whatnot from the type loader.

shmax:
I can do this...

```php
if ($type !== null && ! $type instanceof Type) {
  $type = $type();
}
```

...if we throw out the tests that exercise returning `stdClass` and whatnot from the type loader.

src/Type/Schema.php
```diff
@@ -341,6 +341,10 @@ private function loadType(string $typeName) : ?Type
 
         $type = $typeLoader($typeName);
 
+        if (is_callable($type)) {
```
vladar, line 344: I think this should be a reasonable compromise (and also won't break for people having `__invoke` in their types - I've seen this weirdness somewhere):

```suggestion
        if (! $type instanceof Type && is_callable($type)) {
```

spawnia: Agree that we should not break `__invoke`-lovers.

If `! $type instanceof Type && ! is_callable($type)` would be invalid, right? Let's make sure we do the minimal amount of work required, as this part is somewhat performance-critical.

shmax:
@vladar @spawnia Thanks for the review comments--I'll get to them once I get my PC back from the shop--but what do you think at a high-level? Should the userland typeloader resolve lazy loaded types, or not? I'm leaning toward should, but I would love some additional insight...

src/Type/Schema.php
vladar: I think `instanceof` check will be a happy path for 90+% of situations and `is_callable` won't even run for those cases. And `instanceof` is much cheaper than `is_callable`.

Also, those wishing to squeeze the last bit of performance can optimize their `typeLoaders` to always return instances of Type to hit the happy path.

Or am I missing something?

shmax: > Also, people wishing to squeeze the last bit of performance can optimize their typeLoaders to always return instances of Type to hit the happy path.

Oh, that's an interesting point I hadn't considered--if we play all this right then it shouldn't matter what the client code returns from his type loader, as long as its a `Type` or a `callable`...

shmax: I went with @vladar's suggestion. Looks pretty clean to me. Dunno why I didn't think of it myself, but I'm happy with it if you guys are. Please review again when you have a chance.

src/Type/Schema.php
```diff
@@ -341,6 +341,10 @@ private function loadType(string $typeName) : ?Type
 
         $type = $typeLoader($typeName);
 
+        if (! $type instanceof Type && is_callable($type)) {
+            $type = $type();
+        }
```
spawnia, line 346:
```suggestion
        if (! $type instanceof Type) {
            if (is_callable($type) {
                $type = $type();
                if (! $type instanceof Type) {
                   $this->throwNotAType($typeName, $type);
                }
            } else {
                $this->throwNotAType($typeName, $type);
            }
        }
        
        // Remove the extra $type instanceof Type check below
```

This avoid one redundant call to `$type instanceof Type` in the happy path where the typeloader just returns a type.

shmax: SGTM. 09109bc

src/Type/Schema.php
```diff
@@ -335,20 +335,22 @@ private function loadType(string $typeName) : ?Type
...
 
         if (! $type instanceof Type) {
-            throw new InvariantViolation(
-                sprintf(
-                    'Type loader is expected to return a callable or valid type "%s", but it returned %s',
-                    $typeName,
-                    Utils::printSafe($type)
-                )
-            );
+            if (is_callable($type)) {
```
vladar, line 345: Mind adding a comment that we do this for the sake of performance? Otherwise, we can erase this in some refactoring in the future (did this in the past).

shmax: Sure, no prob. Will touch it up after work.

shmax: Here there be dragons: a16c80b

shmax:
@vladar I hate to be a pest, but is there any chance we can wrap this up?

PR #688, opened 2020-06-29
Fix double Error wrapping

enumag:
Since `$error` already is an instance of the Error class this results in wrapping an Error by another Error which is useless and complicates error handling. In my opinion this should be fixed.

### DIFF
--- a/src/Executor/Values.php
+++ b/src/Executor/Values.php
@@ -124,7 +124,7 @@ class Values
                                     $error->getSource(),
                                     $error->getPositions(),
                                     $error->getPath(),
-                                    $error,
+                                    $error->getPrevious(),
                                     $error->getExtensions()
                                 );
                             }


enumag:
Actually, since `Error::$message` is public maybe we can just prepend `$messagePrelude` and not create new Error instance at all. Thoughts?

spawnia:
> Actually, since `Error::$message` is public maybe we can just prepend `$messagePrelude` and not create new Error instance at all. Thoughts?

Not anymore, actually. Make sure your fork is up to date with the latest changes.

enumag:
Ah, right, I'm locally on 0.13 still. In that case we can keep this PR as is.

enumag:
Just realized I can't update to 0.14 yet due to overblog dependency. Is the v0.13 branch still maintained? Should I rebase this PR to target it instead?

spawnia:
You can open another PR that targets https://github.com/webonyx/graphql-php/tree/0.13.x

PR #689, opened 2020-06-30
Fix double Error wrapping

enumag:
Backporting #688 to `0.13.x`


vladar:
Just got a thought: are you sure we always have a `previous` exception here?

enumag:
Not really but it doesn't matter in my opinion.

vladar:
I guess I'll have to double check this, as it can potentially be a breaking change.

PR #691, opened 2020-06-30
cast numbers to string

shmax:
I broke this one in one of my stanning throwdowns from a while ago: https://github.com/webonyx/graphql-php/commit/e1bce3275b3b3229d98e4db28183001990c3c64f#diff-6f9b96c905acf9bade29996c9a3639dbR567

The resolve method is supposed to return a nullable string, but can in fact return an int. The issue seems to be the relevant `Printer` lambdas, which were not always returning strings.

### DIFF
--- a/src/Language/Printer.php
+++ b/src/Language/Printer.php
@@ -183,11 +183,11 @@ class Printer
                     },
 
                     NodeKind::INT => static function (IntValueNode $node) : string {
-                        return strval($node->value);
+                        return (string)$node->value;
                     },
 
                     NodeKind::FLOAT => static function (FloatValueNode $node) : string {
-                        return strval($node->value);
+                        return (string)$node->value;
                     },
 
                     NodeKind::STRING => function (StringValueNode $node, $key) {
--- a/src/Language/Printer.php
+++ b/src/Language/Printer.php
@@ -183,11 +183,11 @@ class Printer
                     },
 
                     NodeKind::INT => static function (IntValueNode $node) : string {
-                        return (string)$node->value;
+                        return (string) $node->value;
                     },
 
                     NodeKind::FLOAT => static function (FloatValueNode $node) : string {
-                        return (string)$node->value;
+                        return $node->value;
                     },
 
                     NodeKind::STRING => function (StringValueNode $node, $key) {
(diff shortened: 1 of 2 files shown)


src/Language/Printer.php
```diff
@@ -181,12 +182,12 @@ public function printAST($ast)
                             . $node->selectionSet;
                     },
 
-                    NodeKind::INT => static function (IntValueNode $node) {
-                        return $node->value;
+                    NodeKind::INT => static function (IntValueNode $node) : string {
+                        return strval($node->value);
```
spawnia, line 186:
```suggestion
                        return (string) $node->value;
```

Faster, since it is not a function call. As far as i know, it does the same. I think we should replace it everywhere.

vladar, line 186: In theory `IntValueNode->value` should be always a string. This change won't harm but it may simply mean that the error is somewhere else (specifically when constructing an instance of `IntValueNode`

shmax: Fair enough. 466f26b519f5b48ae2d524c5653a98a6cbb0f420

shmax: Yes, true. I'll investigate, but it would be great if you could merge this in the meantime, as it is blocking myself and another developer.

shmax:
Actually, I can't explain it, but after noodling around in my application code for a while, I can no longer reproduce the string/int problem, even without this change set. Since I don't know exactly what is going on, let me revert the casts, and limit my changes for now to typehinting and annotations, and I can investigate in more depth if the original problem happens again.

vladar:
Are you on the `master` branch? There was a similar PR recently which could have affected you: #679

shmax:
> Are you on the master branch? There was a similar PR recently which could have affected you

Ah, that must be it, thanks. Any idea why the CI is crapping out?

simPod:
ad CI, github has outages

shmax:
> Looks good. Maybe the other functions in printer can be type-hinted, too.

Feel free to add suggestions or commits. There's no rush on this one, as the fatal flaw has already been corrected.

shmax:
@vladar @simPod @spawnia Thanks very much for reviewing 👍

PR #692, opened 2020-07-01
Replace function calls with type casts

spawnia:
The effect is the same, efficiency is better.


spawnia:
Those coverage gains 🥇

spawnia:
> Is there even any valid use case for using these cast functions? Except things like `array_map('floatval', $arr)` where it makes sense.

`intval()` takes a base as its second parameter, but that's about it.

PR #693, opened 2020-07-01
Add partial parse functions for const variants

spawnia:
Fixes the previously broken partial parse functions where the underlying `parseX` function expects the argument `bool $isConst`.

I went with a prefix of `constThing` for the const variants in order to keep the signature
between the partial parse signature consistent. Also, this matches the existing variants `argument` and `constArgument`.

Resolves https://github.com/webonyx/graphql-php/issues/698


spawnia:
@vladar to be honest, the magic method approach is rearing its ugly head here. If I had endless time, it might be worth it to write out the actual code.

Pragmatically, this functionality is probably used by very few (I might be the only one), and there are bigger fish to catch. At least it's working now.

src/Language/Parser.php
```diff
@@ -248,7 +258,39 @@ public static function __callStatic(string $name, array $arguments) : Node
...
+                'array',
+                'object',
+                'objectField',
+                'directives',
+                'directive',
+            ],
+            true
+        )) {
+            $type = $parser->{'parse' . $name}(false);
+        } elseif ($name === 'constArguments') {
```
shmax, line 276: Is this critical code? I wonder if it might be more performant to use a map...

spawnia: > Is this critical code?

Not at all, it's backs a bunch of obscure convenience functions for some users that manipulate schemas on the SDL level.

> I wonder if it might be more performant to use a map...

Changed to use a switch, fits better and should be optimal in terms of performance.

shmax: > should be optimal in terms of performance

How do you figure? I know some languages will optimize by converting a switch statement to a lookup table, but I don't know if PHP is that smart. If it isn't, then you still have to 14 string operations to get down to the default case. But if the code isn't critical, then I suppose there's no point worry about it.

spawnia: I guess converting to actual methods will be a much bigger performance boost than anything we could do in a magic function.

shmax: Well, I was wondering if a map of keys to callbacks might be more efficient. Something like:

```php

$lookup = [
  'constArguments' => 'parseArguments`,
  'constValueLiteral` => 'parseValueLiteral`,
  ...
];

$type = $parser->{$lookup[$name] ?? 'parse' . $name}(true);

```

Untested. It's only an idea.

spawnia: I prefer not to spend any more time pondering possible microoptimizations to this marginally and non-performance critical bit of functionality. Feel free to add such in a future PR.

spawnia:
@vladar i depend on the bug fixes in this PR and would appreciate a timely review and merge.

src/Language/Parser.php
shmax: You're the boss 👍

PR #694, opened 2020-07-01
Simplify getVisitFn

shmax:
I noticed in the profiler that `Visitor::getVisitFn`seems to be consuming a lot of resources, so I attempted to simplify it a bit. 

Here are the results of running our unit test suite 5 times.

This branch:
https://blackfire.io/profiles/de85207f-a7bc-4d69-bbc1-e950c839eb24/graph

Master:
https://blackfire.io/profiles/a1b07e4f-cba2-4327-84c5-c42e7559d8b2/graph

This is kind of an extreme use case, but you can see we improve performance by about 5%.


shmax:
Maybe you can help me with the coverage reduction. It's flagging this line as not covered...

https://coveralls.io/builds/31813089/source?filename=src/Language/Visitor.php#L234

...but it was not covered on master, too (I verified in the debugger). How do I handle this?

spawnia:
@shmax since you simplified the function, one uncovered line now is worse in terms of percentage.

I don't think we use arrays in place of a `NodeList` anymore - maybe we never have? There are a few courses of action we could take:

1. Write an explicit test to cover it (might involve suppressing PHPStan errors in the setup code)
2. Ignore the uncovered line (we might be able to exclude it explicitly?)
3. Remove the functionality (we might do that in a next major, though, and go with 1. or 2. for now)

shmax:
> I don't think we use arrays in place of a `NodeList` anymore - maybe we never have? There are a few courses of action we could take:

Ahhh, I see... does this have anything to do with that one `fields` member we saw on one of the nodes that was an array, but should probably be a `NodeList`? How about if I add a fix for that to this change set, and we just remove the dead code in a major version, like you say?

shmax:
I went ahead and changed that `fields` member to a `NodeList` and removed that bit of dead code. I don't know if this will have any impact on the public API, but the tests and stan seem happy. Thank you very much for the guidance. 👍

shmax:
@vladar What do ya say?

shmax:
@spawnia @vladar Thanks very much for reviewing, guys. 👍

PR #695, opened 2020-07-02
NamedTypeNode::$name is generally a NameNode

spawnia:
It is easier to deal with it if the value is known to always be the same type.

Edit: Unless doing funky stuff within the visitor. In that case, the type system can not help you.


src/Language/Printer.php
```diff
@@ -226,6 +228,7 @@ public function printAST($ast)
                     },
 
                     NodeKind::NAMED_TYPE => static function (NamedTypeNode $node) : string {
+                        // @phpstan-ignore-next-line the printer works bottom up, so this is already a string here
```
spawnia, line 231: Why does PHPStan not respect that ignore?

shmax: Not available until `0.12.18`, and we're on ` 0.12.18`..?

spawnia: I see, guess i will attempt an upgrade then.

simPod, line 231: assert(is_string()) might be more suitable then?

spawnia: There is nothing external that might mess this up, we are not calling user code in the process of schema printing. It is just that the AST is manipulated in ways that are unusual and cause it to deviate from its usually well-typed form.

That modification is contained solely within this class and is also completely deterministic, so I think an `assert()` is out of place.

simPod: It would be to satisfy **only that one particular** phpstan issue. It's more specific than ignoring **all** errors on the next line.

src/Language/AST/NamedTypeNode.php
```diff
@@ -9,6 +9,6 @@ class NamedTypeNode extends Node implements TypeNode
     /** @var string */
     public $kind = NodeKind::NAMED_TYPE;
 
-    /** @var NameNode|string */
+    /** @var NameNode */
```
shmax, line 12: I'm not sure I understand what you're uo to, here. For the moment, `NamedTypeNode::$name` is legitimately both a `NameNode` and a `string` at different times. That may or may not be what is supposed to happen, but if it's not we should figure it out and fix it for real, and if it is, I'm not sure manipulating phpstan to pretend it's not happening is making things any more clear.

spawnia: The only time where it is a `string` is within `Printer`. As a user of this library, it is always a `NameNode`.

I think we do not need to complicate usage because we misuse this class in a single part of the library.

shmax: > That single usage is out of order

As I'm learning in my own branch, all nodes are actually being subjected to the same misuse, not just `NamedTypeNode`. Let me continue my study today, and I'll be in a better position to advise once I understand the problem better.

src/Language/Printer.php
```diff
@@ -226,6 +230,8 @@ public function printAST($ast)
                     },
 
                     NodeKind::NAMED_TYPE => static function (NamedTypeNode $node) : string {
+                        assert(is_string($node->name));
```
vladar, line 233: I don't think we need this assertion (otherwise we may need similar assertion in many other methods as well)

spawnia: It satisfies Phpstan and is preferable to ignoring an error: If we change things around, it will trigger.

There is no runtime cost.

spawnia: > otherwise we may need similar assertion in many other methods as well

I know what you mean. This conversion seems to be the only one affected, since there is no string concatination (and thus automatic type coercion) involved.

It seems imperfect to me, but pragmatically it looks like our best option for now. https://github.com/webonyx/graphql-php/pull/697 would be a clean fix, until then this PR improves things a bit at least.

src/Language/AST/NamedTypeNode.php
spawnia: You are correct that other places are also affected. This is the only one where no string coercion happens, so PHPStan rightfully complains.

src/Language/Printer.php
spawnia: Agreed, i added `assert()` instead.

spawnia: Upgrade done.

src/Language/AST/NamedTypeNode.php
vladar: The problem with typing AST is that it is an arbitrary data structure that could be transformed into anything using a Visitor. 

In other words, this `string` typing (and actually ANY other type for ANY node) is only possible during this call within an intermediate data structure:

```php
$literallyAnything = visit($ast, $visitor);
```

Nowhere during this call we modify the original `$ast`. There is a transient copy on each step that is transformed. 

We shouldn't add all possible types to AST nodes because they can be literally anything. The only reason why we have `string` here is that we have only one utility (Printer) which transforms AST to string.

But `Visitor` supports ANY kind of transformation. It can also be `int` or `array`, so if we want "pure" types we must literally type it as `mixed` or remove the Visitor.

Which is a point where typings start bringing more harm than good? I like them as a convenience tool to help developers but not as the final truth dictating how to structure the program. Otherwise, I would have picked Java or another strictly typed language.

vladar: I guess we could remove argument type hints from the printer Visitor which would indicate that this is not really an AST node instance but an intermediate structure.

src/Language/Printer.php
vladar: What happens if we change:

```php
NodeKind::NAMED_TYPE => static function (NamedTypeNode $node) : string
```
to:
```php
NodeKind::NAMED_TYPE => static function ($modifiedNode) : string
```

As this is a more correct typing.

spawnia:
@vladar you make a good point. We should not try to shoehorn strong types in places where they are not accurate, sometimes PHP can be quite loose - especially when porting directly from `graphql-js`.

That said, i do prefer code that can be statically validated to be type safe, if the implementation allows for it. Not saying we must do it at all costs (e.g. performance matters more).

Should I remove all type hints from the `Printer` closures in this PR?

src/Language/Printer.php
spawnia: That works fine, PHPStan no longer complains since it has nothing to work with.

vladar: I think it is the most reasonable thing to do here as we deal with temporary objects, not full AST nodes. So please, go ahead and remove those type hints.

src/Language/AST/NamedTypeNode.php
spawnia: Removed, added a PHPDocs that explains why for future reference.

shmax: > The problem with typing AST is that it is an arbitrary data structure that could be transformed into anything using a Visitor.

Yes, you can do anything you want in the visitor pattern, that is the point of the visitor pattern. But in our particular implementation you are calling `clone` on each node, and in PHP, that clone has the same type as the original. You then violate the types on the clone. I understand that this is a port of the javascript algorithm, but the javascript algorithm operates on javascript objects, not class instances.

> Nowhere during this call we modify the original $ast. There is a transient copy on each step that is transformed.

Again, you modify a clone, which like it or not, in PHP has the same type as the original. True, the original is not touched, but no one is accusing you of that; the problem is that the clone is being misused, and phpstan is detecting that. 

 > We shouldn't add all possible types to AST nodes because they can be literally anything. The only reason why we have string here is that we have only one utility (Printer) which transforms AST to string.

I'm a bit late in learning about the Visitor pattern, so I'm not really the best one to lecture about this, but every single visitor pattern I've looked at (I spent last night reading blogs and sample repos, watching videos, and studying up on the wiki page) does indeed switch on type. That is exactly how the visitor pattern works; it's a method for adding functionality to a family of classes without changing the classes, rather then subclassing or adding functionality to them, and to do this it has to do a lot of type checking, 

Now, one thing to keep in mind is that we have two pieces to our printer visitor puzzle: we have the `Visitor` class, and we have our `Printer` class. The `Printer` is actually the true implementation of the visitor pattern; it considers the type of each node, and adds implements its specialized functionality (producing a string). The `Visitor` class, is just a helper that iterates over the tree structure without recursion (which is apparently not a requirement of the visitor pattern). Making a copy of the node or transforming it in any way is not a key facet of the visitor pattern, and if it happens should probably happen in the actual visitor pattern implementation (meaning the `Printer`) and not in the helper.

> Which is a point where typings start bringing more harm than good? I like them as a convenience tool to help developers but not as the final truth dictating how to structure the program. Otherwise, I would have picked Java or another strictly typed language.

This has nothing to do with whether types are good or bad. The visitor pattern switches on type. That's how it works. We're okay with that, so far. Our only problem is that our visitor implementation--in this case a `Printer`--is doing its work in a graceless way. Okay, so it wants to take a tree of AST nodes and convert it to a string. Fine, this is something that happens all the time in compilers and language processors. The only problem here is that rather than creating a true intermediary format and storing it in visitor space (meaning the printer), there is a clone happening in the `Visitor` class (the helper) itself, and then the visitor implementation (the printer) is using that clone as raw storage to stuff strings into, without any kind of graceful handling. 

You're sort of getting away with it for now because those members are not currently typed at the native language level, but imagine what's going to happen to this pattern once we get around to supporting PHP 7.4 and start trying to type the node members. The architectural messiness of the current printer system will break down, and you will be forced to solve the problem properly.

My vote is to solve the problem properly, and it's what I'm currently attempting to do in my branch.

shmax: > I guess we could remove argument type hints from the printer Visitor closures which would indicate that this is not really an AST node instance but an intermediate structure.

No, please stop. You are making things worse. The visitor pattern does indeed switch on type. That's what it does. Our problem is not that we are using types. Our problem is that our intermediary format is not an intermediary format--it's literally a clone of the original instance, and so it has the same types as the original. If you want to use an intermediary format you can, but _make a new format_... don't just use the original and scramble its guts.

shmax:
Guys, please don't merge this. I know I'm a late "visitor" to the party, but I'm learning fast, and I believe this problem can be solved in a way that strengthens our typing, rather than weakening or disdaining it. Please give me a few days to progress my own branch, and I am sure I can make everybody happy.

src/Language/AST/NamedTypeNode.php
vladar: > but imagine what's going to happen to this pattern once we get around to supporting PHP 7.4 and start trying to type the node members

We'll live in the Java world then? I feel we are already half-way there :) Then why don't we use Java instead? Type hints in PHP are this: hints.

I've seen this several times - how adding strict types make things that used to be simple much harder and taking hours and days just to satisfy the type checker/compiler instead of writing something useful. I am not subscribing to 100% compatibility with PHPStan. We could have used this time to increase code coverage but instead, we fight with the type checker.

I still think we should use it as a tool where it makes sense but don't be fanatics with static analysis - the balance is important.

spawnia:
@shmax Your approach in #697 is very good and I want to pursue it further.

> If you want to use an intermediary format you can, but make a new format... don't just use the original and scramble its guts.

That would be a good solution too, but might not be worth the effort.

I do not intend or hope for the solution in this PR to be what we finally end up with, but I think it is a pragmatic solution that will work for now and improves upon what we have now on the following points:
- Clarifies that there is duck-typing going on within the Printer
- Satisfies PHPStan
- Improves the accuracy of the actual AST, this directly benefits end users

src/Language/AST/NamedTypeNode.php
shmax: > We'll live in the Java world then? I feel we are already half-way there :) Then why don't we use Java instead? Type hints in PHP are this: hints.

False equivalency. The problem we are dealing with has nothing to do with whether typing is good or bad, or being taken too far. We have the types already, like it or not. The fact that our current version of PHP allows us to trash the types does not mean that we have to, nor does it mean that our current algorithm is doing things correctly. It obviously has problems, because here we are.

> I've seen this several times - how adding strict types make things that used to be simple much harder and taking hours and days just to satisfy the type checker/compiler instead of writing something useful. I am not subscribing to 100% compatibility with PHPStan. We could have used this time to increase code coverage but instead, we fight with the type checker.

Please don't turn this into a holy war. We don't need this to devolve into some kind of weird meta discussion about whether types or good or bad. We have an architectural problem with our implementation of the visitor problem, and we can fix it. Just a reminder, this exact problem has been solved countless times in typed languages, so there's no reason we can't solve it here, too. Looseness of typing is obviously not a requirement to get this to work (I know you don't like my recursive solution in its current state, but it is an implementation of the visitor pattern with no mess).

> I still think we should use it as a tool where it makes sense but don't be fanatics with static analysis - the balance is important.

We don't have to make that choice, here. We can solve this problem in a cleaner way.

shmax:
> I do not intend or hope for the solution in this PR to be what we finally end up with, but I think it is a pragmatic solution that will work for now and improves upon what we have now on the following points:
> 
> * Clarifies that there is duck-typing going on within the Printer
> * Satisfies PHPStan
> * Improves the accuracy of types within the AST, this directly benefits end users

@spawnia and @vladar  I want to thank you for your patience and class in this matter. I dig your style, and I think deep down we agree on most things. You and @simPod have built this amazing library, which is frankly something I know I would not have been able to do. You guys are way, way better at this stuff than I am. But I know there's something deeply fishy about this visitor pattern business. We're just not doing it quite right. PHPStan is doing its job and trying to warn us, and rather than listen, you guys are looking for ways to put a sticker over the engine light. 

What concerns me is that you are leaping right from "php is not java" to "we have to throw out all the types and use duck typing". That logical flow does not compute. There are many, many ways to clean all this up that don't require us to weaken any types, and by the same token we should not be too slavish in trying to mimic the javascript architecture (if you were serious about doing it their way, you would not use `clone` in `Visitor`, but would instead  instantiate a new `stdClass` or associative array).

So, as of now we have three competing styles. Let's tally them up:

| branch                                    | Uses Visitor pattern?                         | Recursive? | Trashes types? | PHPStan coverage                                                                                                                                                      | Ready for PHP 7.4?                                                                |
|-------------------------------------------|-----------------------------------------------|------------|----------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------|
| master                                    | Yes, but not correctly (objects are modified) | No         | Yes            | Currently complaining about the Variable property access on the cloned nodes, but otherwise only complains about NamedNodeType.                                       | No. If we typehint our members, the code will break, with or without annotations. |
| spawnia:monomorphize-named-type-node-name | Yes, but not correctly (objects are modified) | No         | Yes            | Currently complaining about the Variable property access. Complaint about NamedNodeType removed, but only because we have weakened type annotations in Printer.       | No. Code will break, with or without annotations.                                 |
| shmax:fix-namenode-type                   | Yes (but does not currently use Visitor.php)  | Yes        | No             | Still complaining about the Variable property access in Visitor, but that code is no longer being exercised. NameNodeType violation resolved. Printer types retained. | Yes. No types are violated. Can typehint members.                                 |

Naturally I'm biased, but I feel like my branch is the closes to being on the right track. The only two arguments I've seen against it are both from from Vladar: 1) It's not easy to port changes from the javascript library and 2) it's recursive.

I think I can accommodate the first concern by arranging my switch into a shape that matches the javascript object. As for the recursion, well, none of the other libraries I've looked at seem to view it as a problem (and the visitor pattern itself doesn't seem to have an opinion about it), but if it's going to be a blocker then I am prepared to spend more time trying to merge it with our current `Visitor` class.

So I'm just asking that we not rush into anything. I'm  a single guy with few other interests, and I will work on this until I get it figured it out to everybody's satisfaction. Merging this branch won't really improve our situation any, and will create a conflict nightmare for me if we decide to go with my approach. 

I've got a few other fires to put out first with my optimize branch, but once that is sorted I'll focus on this. Thanks for hanging in there.

spawnia:
@shmax Thanks for the kind words. I admire your dedication and persistence on this matter and think it is great that you are willing to go the extra mile to really polish parts of the library. As I said before, I think your reworked approach to the `Printer` goes in the right direction. I will happily follow your progress in the other PR.

@vladar I reverted the changes to the point of your original review, replaced the `assert(is_string())` with a `@phpstan-ignore-next-line` and added a PHPDoc to the printer. It's a small change, but it will save myself and others a bit of headache.

vladar:
Hey, @shmax first, I want to say that I absolutely appreciate all the work you put into this library and happy we have you as a contributor!

I'll try to explain my point on type-safety.

> Please don't turn this into a holy war. We don't need this to devolve into some kind of weird meta discussion about whether types or good or bad. 

No offense but you've spent days to change typing `NameNode|string` to `NameNode` in a "type-approved", "correct" way. This work doesn't add new functionality, increases test coverage, or improves things significantly apart from making it type-safe. And this is a constantly re-occurring pattern when introducing type-safety (it is not the first time when I see this).

Not to mention disputes that always accompany this topic and that drain time and energy from people without real benefits to anyone.

Type-safety is not a self-goal here. Value for library users and the efficient use of our time is.

> We have an architectural problem with our implementation of the visitor problem, and we can fix it.

I can agree with this but what real-world issue do you have with this architectural problem? Did it manifest itself in some project? Do we have bug reports (apart from PHPStan complaining)? Will it open new possibilities for us in the foreseeable future (apart from type-safety again)?

I am happy to tackle any real-world problem that occurs within this library but for me, **substantial** changes like re-writing Printer must be well justified. The "pain" must be real for library users. I don't see it yet.

I do get that you consider "type-safety" as a good-enough reason for major changes but I disagree here. I am completely open and ready to return to this discussion when we face some real-world problems caused by the current implementation.

We can keep your PR open until then. An alternative approach would be to actually use an intermediate `stdClass` structure in the visitor vs using `clone` as you've mentioned but that will be a breaking change and require deeper investigation which, again, hard to justify at the moment.

shmax:
> No offense but you've spent days to change typing NameNode|string to NameNode in a "type-approved", "correct" way. This work doesn't add new functionality, increases test coverage, or improves things significantly apart from making it type-safe. And this is a constantly re-occurring pattern when introducing type-safety (it is not the first time when I see this).

No, I've spent days trying to preserve typing for _all_ nodes in the tree.

> This work doesn't add new functionality, increases test coverage, or improves things significantly apart from making it type-safe. And this is a constantly re-occurring pattern when introducing type-safety (it is not the first time when I see this).

> This work doesn't add new functionality, increases test coverage
That is all vague reasoning that could be used to argue against literally any change. It is not a rationale for or against any of the approaches being discussed.

> And this is a constantly re-occurring pattern when introducing type-safety (it is not the first time when I see this).

This issue is not about whether types are good or bad. The issue is that our visitor pattern is not implemented correctly, and the symptoms are occasionally apparent as we try to work with our static checker. In this case the types are very much helping us, as they have set off warning bells that something is amiss.

> Not to mention disputes that always accompany this topic and that drain time and energy from people without real benefits to anyone.

We don't have a dispute about types. We have a real technical problem: our visitor pattern is not implemented correctly.

> without real benefits to anyone

The benefit of solving this properly is that it will allow us to move forward in improving the quality of the code. I don't talk about this much, but I've spent the last 2 years of my life rewriting the back end in main application to use this library. My work is feature complete, but I haven't been able to launch because this library was not production ready. Basically, it was slow. Since I've started working on this library, I've improved performance for my queries by about what, 30%? I have a PR open that improves performance by another 5%, and my printer branch improves introspection query performance quite significantly (I don't have dedicated numbers). I'm able to do this through code analysis, and code analysis is aided hugely when the code is clean, clearly typed, and self-documented. When the types are strong (and not mis-annotated), all kinds of wonderful things become possible. Your IDE can assist with auto-completion and intellisense. It will flag problems it can see. Static checker like PHPStan can unearth unhealthy implementations, like we see with visitor.

> Type-safety is not a self-goal here. Value for library users and the efficient use of our time is.

Yes, and code quality helps we, the developers, use our time more efficiently. I don't _like_ having to spend so much time on this issue--I've got a 2-year long refactor I'm trying to launch--but if we solve it correctly instead of sweeping it under the carpet we will enable years of future improvements.

> I can agree with this but what real-world issue do you have with this architectural problem? Did it manifest itself in some project? Do we have bug reports (apart from PHPStan complaining)? Will it open new possibilities for us in the foreseeable future (apart from type-safety again)?

No, this is not a massive problem. But this kind of meta-argument about whether it's useful or not is just not productive. I could use your reasoning to diminish any number of work that I think you would agree is important. Watch this:

Code sniffer:
What-real world issue does it solve? Did it manifest itself in some project? Did it open new possibilities?

Code coverage:
What-real world issue does it solve? Did it manifest itself in some project? Did it open new possibilities?

PHPStan:
What-real world issue does it solve? Did it manifest itself in some project? Did it open new possibilities?

I think you'll agree these are all important (personally, I could live without the code sniffer, mainly because it seems to have a lot of bugs, and the maintainers are slow to fix them and adapt to new developments, such as array shapes), yet they all play a part in helping us maintain this library. 

> I am happy to tackle any real-world problem that occurs within this library but for me, substantial changes like re-writing Printer must be well justified. The "pain" must be real for library users. I don't see it yet.

The pain is on the DX end, which is just as important as public consumption. As a developer working hard to improve this library, the problem pops up often enough for me to take notice. All three of us (@spawnia @simPod and myself) have noticed the weird type issues independently. All three of us want to fix them. 

> We can keep your PR open until then. An alternative approach would be to actually use an intermediate stdClass structure in the visitor vs using clone as you've mentioned but that will be a breaking change and require deeper investigation which, again, hard to justify at the moment.

If you don't view type issues to be of importance, then by the same reasoning you should not be overly opposed to _improvements_ to them any more than you are opposed to weakening them. Give me some time to work on my PR some more. I'm still learning, and I've already said I think it's likely I can make everyone happy. What's not to like?

vladar:
> We don't have a dispute about types. We have a real technical problem: our visitor pattern is not implemented correctly.

Can you write a test case demonstrating this incorrectness? If no, then we ARE having a dispute about types.

> I could use your reasoning to diminish any number of major features in this project that I think you would agree are important. 

My argument is not even about types in general - they are useful up to a point of course. And surely improve DX a lot. But there is a point after which costs and benefits of increasing type-soundness do not quite match. I am pretty sure you understand what I am trying to say. Not sure why you start playing those games of words.

We are talking about "Before" and "After" as well as about the cost of this transition.

> Code sniffer:

Before: no enforced code conventions
After: enforced and auto-fixable code conventions, no endless disputes and friction about code style
Efforts: medium, incremental (especially if done early)

Worth the effort? 100% (as for me). I don't like some aspects of our code style but changing those doesn't worth the efforts.

The answer could have been quite different if there was no auto-fixing for example.

> Code coverage:

Before: any change can break things and go unnoticed to releases
After: most of the errors are caught early, code behaves as expected, way less time spent on debugging and user support, way less user confusion.

Worth the effort? 100%

But also up to a point - tests require maintenance, and at some point increased coverage won't worth the efforts - same as type soundness.

> PHPStan:

Before: some trivial errors sneak through tests
After: trivial errors are captured early

Worth the effort? Maybe - I am actually still not convinced because you can achieve much more with tests. PHPStan kinda creates a false sense of safety which makes people write fewer tests. And often forces you to "fix" things that are not bugs. In reality, it catches only few trivial bugs and lots of "potential" bugs most of which actually never occur in real projects.

But PHPStan didn't require major efforts to introduce and so the benefit and effort tradeoff was OK (until we pushed type-safety too far).

---

Now let's get back to our case with `Printer`:

Before: a working implementation that's been used in production for years now (with one potential type-safety issue)
After: another working implementation (with possibly unknown issues which are yet to be discovered)

Efforts: hours and days of implementation and discussions.
Worth the effort? I don't think so.

The difference is that in your opinion this `Before` step is a completely broken implementation and `After` step is a much better one. But that's only true in terms of type-soundness. Again - can you write a test case for the `Printer` that breaks with current implementation? If not, then what's incorrect?

> Give me some time to work on my PR some more. I'm still learning, and I've already said I think it's likely I can make everyone happy. What's not to like?

Sure, it is highly likely that we'll merge your PR eventually - at least it performs better.

My problem with this whole situation is that it is a re-occurring pattern when type-safety (or another abstract ideal) is pushed too far and its costs start to overweight its benefits.

shmax:
> Can you write a test case demonstrating this incorrectness? If no, then we ARE having a dispute about types.

No, not at the moment. I need to study up a bit more, but this is from the first paragraph from the wikipedia page:

"In essence, the visitor allows adding new virtual functions to a family of classes, without modifying the classes."

You could argue that since you are creating clones that the original objects are not altered, so that's fine, but if we follow your instinct to drop type knowledge, then we no longer have the "family of classes" part. If we keep the clones, then we are in violation of the "without modifying the classes" part (though I think that may be more about not having to add code--rewriting the fields as different types at runtime is not ordinarily possible in typed languages). I suppose a good lawyer could keep you out of jail, but we're just not obeying the spirit of the thing. I'm not an expert yet, so I'll back off on this until I can look at some more formal use cases, but in my other branch the problem has been rendered moot, as all cloning and need for wiggly semantics have been removed. I will power-up this weekend.


> But there is a point after which costs and benefits of increasing type-soundness do not quite match. 

Let me know the exact moment we get there. I don't think this issue is it, especially when a decent solution is ready to go in another branch. The work has been done. The tests pass. I'm happy. @spawnia is happy. PHPStan is happy. What's it going to take to make you happy?


> I am pretty sure you understand what I am trying to say. Not sure why you start playing those games of words.

Because I know when a discussion is drifting off point. We have a technical problem to solve here, and instead of focusing on that you're using a vague meta-argument about whether a solution will directly enable new features for users to enjoy. A lot of the work we do does not directly enable new features for the users. You could remove the code sniffer right now, and all the tests would still pass, and no one would ever know. Is it a good idea? Of course not. I just use the example to illustrate that your argument about end users or new features is not really relevant to this problem.

> After: another working implementation (with possibly unknown issues which are yet to be discovered)

That's fair. If you're a bit hesitant about it, maybe we could do something like a "PrinterExperimental". Since the new printer doesn't depend on any other changes, it can drop in and we can slowly trot it out for certain apps willing to be a test pilot.

> Efforts: hours and days of implementation and discussions. Worth the effort? I don't think so.

That's a matter of opinion that doesn't need to be debated, as the wasted hours are on me. I don't view them as wasted because the code quality is objectively better, and I enjoy this kind of work.

> The difference is that in your opinion this Before step is a completely broken implementation and After step is a much better one. But that's only true in terms of type-soundness. Again - can you write a test case for the Printer that breaks with current implementation? If not, then what's incorrect?

The old code does indeed work; it's just not a valid implementation of the design pattern it's meant to use. The evidence is that our types are being trashed. I know you view PHP's ability to change types on classes at runtime as a strength, but I don't think you will find many people that share that view, least of all the developers of PHP themselves (see PHP 7.4).

> Sure, it is highly likely that we'll merge your PR eventually - at least it performs better.

Great. If we get that far, then we can forget this whole nasty business until it crops up again.

> My problem with this whole situation is that it is a re-occurring pattern when type-safety (or another abstract ideal) is pushed too far and its costs start to overweight its benefits.

You're gonna be a tough nut to crack once we raise minimum support to 7.4 :)

shmax:
By the way, consider this: by avoiding the cloning and other churn happening in `Visitor` we were able to speed up the printer by a factor of 3. What would happen to regular query performance if we remove that clone? Just wondering...

vladar:
> By the way, consider this: by avoiding the cloning and other churn happening in Visitor we were able to speed up the printer by a factor of 3. What would happen to regular query performance if we remove that clone? Just wondering...

`clone` is only called when editing a node (in the printer), it is not called for simple traversal, i.e. validation.

vladar:
>> Can you write a test case demonstrating this incorrectness? If no, then we ARE having a dispute about types.

> No, not at the moment. 

> The old code does indeed work; it's just not a valid implementation of the design pattern it's meant to use

Works, just not valid. Right.

shmax:
> clone is only called when editing a node (in the printer), it is not called for simple traversal, i.e. validation.

Ah, I see, thank you. Well, if the one thing that was doing edits (I think I remember seeing something else set `false` on one of the nodes) can be removed, then if nothing else perhaps the complexity of `Visitor` can be reduced in the future.

shmax:
> Works, just not valid. Right. Hard to argue with that.

Sounds like we're making progress. I'm sure we can find common ground to meet on. And for the record I rather enjoy these debates; I think you've very good at it, and if you had a youtube channel I would definitely subscribe.

spawnia:
So, guys, the real question here is: When are we finally rewriting this library in Haskell?

Issue #655, opened 2020-04-28
NamedTypeNode#name should be of type NameNode (?)


simPod:
Merged in https://github.com/webonyx/graphql-php/pull/647 we introduced type for `NamedTypeNode#name` as `NameNode|string` (was `string` before).

Another context is in this discussion with @shmax https://github.com/webonyx/graphql-php/pull/645#discussion_r411432617

I'd vote for constraining it to the `NameNode` only, dropping the `string` type.

shmax:
Did you read my analysis of the situation? The type **is** legitimately both a `string` and a `NamneNode` at different times. It's weird and probably some kind of unintended flaw that never got noticed, but it is consistent with the reference library (which also incorrectly annotates it as a `NameNode`. If you change it back to a `NameNode`, you will (rightly) get complaints from PHPStan.

simPod:
Yes, but I was rather wondering if we could simplify the code around it so we can safely use NameNode everywhere. I'm not sure yet if we need to follow reference implementation here or not.

shmax:
You're talking about sorting out the runtime mechanics that write a string to `NamedTypeNode.Name`? Might be worth looking into, but that's a much more complex task than simply changing the annotation back to `NameNode`.

simPod:
I might have not described it clearly but just changing the type hint is not a solution. I shall take a look into it.

vladar:
Yeah, NamedTypeNode#name should be definitely of type NameNode.

We create it here:

https://github.com/webonyx/graphql-php/blob/ed8fb628ad20325aec2c9f99c451cc0eca8b7f53/src/Language/Parser.php#L1006-L1014

And `parseName` always returns `NameNode`. Annotation with `NameNode|string` is probably wrong. But then I am curious what usages rely on this being `string`. We should fix those.

shmax:
It gets written with a string as part of [Visitor::visit](https://github.com/webonyx/graphql-php/blob/master/src/Language/Visitor.php#L283). The [same general sort of thing](https://github.com/graphql/graphql-js/blob/4150d1f51360a7981181cfec42a135394c7340f1/src/language/visitor.js#L274) happens in the reference library. So, if there's a bug, then we inherited it from them.

spawnia:
Resolved as of https://github.com/webonyx/graphql-php/pull/697

PR #697, opened 2020-07-03
Fix namenode type / rewrite Printer

shmax:
An alternative fix for #655,  the infamous `NamedTypeNode` `NameNode` / `string` debacle. 

Took me a while to understand, but all the trouble was being caused by the unusual algorithm used to print ASTs. It tied into `Visitor::visit`, and starting at the leaf nodes, would stomp over properties with strings so that it could concatenate everything into a big string. We noticed this happening on the `NamedTypeNode` in particular, but actually this was happening with all of the nodes.

Once I understood what was happening, I was able to rearrange the algorithm into a more conventional top-down recursive one that does not "edit" any nodes, nor do any illegal writes. While I was at it, I corrected a number of type annotations. PHPStan is now happy. Note that this change is not as dramatic as it might look; all of the callbacks from the old `leave` section in `Printer` were simply ported into a big switch statement. I didn't change any of the core logic.

Interestingly, performance seems to improve with these changes, though I'm not sure why.

master:
https://blackfire.io/profiles/a1b07e4f-cba2-4327-84c5-c42e7559d8b2/graph

this branch:
https://blackfire.io/profiles/755fec5e-3bf3-491a-b8d0-8a6d78a30fd6/graph


src/Language/Visitor.php
```diff
@@ -211,36 +212,28 @@ public static function visit($root, $visitor, $keyMap = null)
...
                     }
                     $editOffset = 0;
-                    for ($ii = 0; $ii < count($edits); $ii++) {
-                        $editKey   = $edits[$ii][0];
-                        $editValue = $edits[$ii][1];
-
-                        if ($inArray) {
+                    if ($inArray) {
+                        foreach ($edits as $editKey => $editValue) {
                             $editKey -= $editOffset;
```
shmax, line 221: I watched this carefully in the debugger, and it was never anything but `0`. Is it important?

src/Language/AST/NamedTypeNode.php
```diff
@@ -9,6 +9,14 @@ class NamedTypeNode extends Node implements TypeNode
...
-    /** @var NameNode|string */
+    /** @var NameNode */
     public $name;
+
+    /**
+     * @param mixed[] $edits
+     */
+    public function setEdits(array $edits)
+    {
+        $this->name->value = $edits['name'] ?? $this->name->value;
```
shmax, line 20: The other thing that bothers me is that the incoming name value was never different from the value in `$this->name->value`. In other words, I have no idea what it means to edit this (or any other) node.

shmax:
@spawnia @simPod @vladar PTAL

shmax:
Heh, I decided to see if I could extend this fix to the other members that were being stomped by strings, and things sort of ballooned out of control. I still don't really understand where those strings are coming from, why they are being written, or what it means to "edit" a node, but I'm learning, and I have a feeling I'll be able to make this all simple again once I understand what the hell is going on. Let's call it a work in progress. More soon...

spawnia:
@shmax The happenings in the `Printer` are a bit confusing. Basically, it does a bottom-up traversal of the AST where it converts all Nodes to `string`. One of the first things to get converted are `NameNode`'s, since they are leaves of the AST.

shmax:
> @shmax The happenings in the `Printer` are a bit confusing. Basically, it does a bottom-up traversal of the AST where it converts all Nodes to `string`. One of the first things to get converted are `NameNode`'s, since they are leaves of the AST.

But why does the printer need to do any writes?

shmax:
> But why does the printer need to do any writes?

I've been studying away. Let me see if I can answer my own question. The printer prints an AST by concatenating children. Because the desired ultimate output is a string, each child must ultimately evaluate to a string. It does this by "visiting" each child, and converting fields on the children to strings as it "leaves" the child during traversal. This is the "editing" happening in the codebase (there's one other that seems to happen once per query, but I'll come back to that).

Now that I understand it a bit better, I'll see if I can make things a bit less sloppy...

phpstan-baseline.neon
```diff
@@ -20,6 +20,21 @@ parameters:
 			count: 2
 			path: src/Executor/Executor.php
 
+		-
+			message: "#^Switch condition type \\(true\\) does not match case condition \\$selection instanceof \\\\GraphQL\\\\Language\\\\AST\\\\FieldNode \\(false\\)\\.$#"
```
shmax, line 24: I have no idea where the hell these errors came from; I didn't touch ReferenceExecutor...

src/Language/AST/NamedTypeNode.php
```diff
@@ -9,6 +9,6 @@ class NamedTypeNode extends Node implements TypeNode
     /** @var string */
     public $kind = NodeKind::NAMED_TYPE;
 
-    /** @var NameNode|string */
+    /** @var NameNode */
```
shmax, line 12: It took 12 hours, but I finally managed to properly fix this guy 🎈

src/Language/AST/NamedTypeNode.php
shmax: I understand, now. The only "editing" being done was the printer class's kludgy print method.

tests/Language/VisitorTest.php
```diff
@@ -187,9 +187,7 @@ public function testAllowsEditingNodeOnEnterAndOnLeave() : void
                         $selectionSet = $node->selectionSet;
 
                         $newNode               = clone $node;
-                        $newNode->selectionSet = new SelectionSetNode([
-                            'selections' => [],
```
shmax, line 191: This empty array assignment was a type violation. PHPStan didn't catch it because we haven't annotated our config objects.

tests/Language/VisitorTest.php
```diff
@@ -1583,10 +1581,10 @@ public function testMaintainsTypeInfoDuringEdit() : void
                                 'directives'   => $node->directives,
                                 'selectionSet' => new SelectionSetNode([
                                     'kind'       => 'SelectionSet',
-                                    'selections' => [new FieldNode([
+                                    'selections' => NodeList::create([new FieldNode([
```
shmax, line 1584: Another violation: `selections` must be a `NodeList`.

composer.json
```diff
@@ -55,7 +55,7 @@
     "lint" : "phpcs",
     "fix" : "phpcbf",
     "stan": "phpstan analyse --ansi --memory-limit 2048M",
-    "phpstan-baseline": "phpstan analyse --ansi --generate-baseline",
+    "baseline": "phpstan analyse --ansi --generate-baseline",
```
spawnia, line 58: :pray: Shortcuts should be short. Totally agree with the change

spawnia:
> 1. It will make maintenance much harder (can't port changes from graphql-js as easy as we do now)

The mechanics of how the printer works have been drastically changed, sure. But the overall problem shape as well as the details for individual nodes still match quite well.

> 2. Printer (and visitor in general) is implemented this way exactly to avoid recursion

There is a natural limit to the recursion depth in GraphQL schemas, so that part should be fine. It might be interesting to figure out at what point nested queries would become problematic, but I don't see us hitting a limit there realistically.

I think it is worth it for us to take advantage of PHP's type system as much as possible. While it may not be the most convenient from a maintenance perspective, considering the drift from `graphql-js`, it certainly makes this library better for users. I rely heavily on static analysis in my projects. It is crucial that the libraries i depend on have accurate type definitions in order for it to work well. @shmax, @simPod and myself are certainly dedicated to it and willing to put in the work.

src/Language/AST/UnionTypeExtensionNode.php
```diff
@@ -12,9 +12,13 @@ class UnionTypeExtensionNode extends Node implements TypeExtensionNode
     /** @var NameNode */
     public $name;
 
-    /** @var DirectiveNode[]|null */
+    /** @var NodeList<DirectiveNode>|null */
     public $directives;
 
-    /** @var NamedTypeNode[]|null */
+    /**
+     * TODO: is this actually a NodeList<NamedTypeNode>?
```
spawnia, line 19: It seems like it should be. Have you considered `UnionTypeDefinitionNode`?

composer.json
simPod: 😢

src/Language/Printer.php
```diff
@@ -86,379 +92,356 @@ public static function doPrint($ast)
...
-                            . ($noIndent
-                                ? $this->wrap('(', $this->join($def->arguments, ', '), ')')
-                                : $this->wrap("(\n", $this->indent($this->join($def->arguments, "\n")), "\n"))
-                            . ($def->repeatable ? ' repeatable' : '')
-                            . ' on ' . $this->join($def->locations, ' | ');
-                    }),
-                ],
-            ]
-        );
+    public function p(?Node $node, bool $isDescription = false) : string
```
spawnia, line 141:
```suggestion
    public function print(?Node $node, bool $isDescription = false) : string
```

spawnia: Or `printNode`

src/Language/Printer.php
```diff
@@ -86,379 +92,356 @@ public static function doPrint($ast)
...
+        if ($node === null) {
+            return '';
+        }
+        switch (true) {
+            case $node instanceof ArgumentNode:
+                return $this->p($node->name) . ': ' . $this->p($node->value);
+            case $node instanceof BooleanValueNode:
+                return $node->value ? 'true' : 'false';
+            case $node instanceof DirectiveDefinitionNode:
+                $argStrings = array_map(function ($item) : string {
```
spawnia, line 153: What is the type of `$item`?

src/Language/Printer.php
```diff
@@ -86,379 +92,356 @@ public static function doPrint($ast)
...
+                        $this->printList($node->directives, ' '),
+                        $this->p($node->selectionSet),
+                    ],
+                    ' '
+                );
+            case $node instanceof FloatValueNode:
+                return $node->value;
+            case $node instanceof FragmentDefinitionNode:
+                // Note: fragment variable definitions are experimental and may be changed or removed in the future.
+                return sprintf('fragment %s', $this->p($node->name))
```
spawnia, line 232:
```suggestion
                return 'fragment ' . $this->p($node->name)
```

I think that would be much more readable. (Full disclosure: I despise `sprintf` :shrug:)

tests/Language/VisitorTest.php
spawnia, line 1584:
```suggestion
                                    'selections' => new NodeList([new FieldNode([
```

shmax:
> 1. It will make maintenance much harder (can't port changes from graphql-js as easy as we do now)

I don't agree with this statement at all. As I said in my PR description, the new architecture maps directly to the old one; I simply rearranged the old monstrous config object into a switch statement. The per-node logic is unchanged. If somebody changes the print logic for a node, you can make the corresponding change in the switch statement without any major cognitive friction.

I also want to say that the new architecture is MUCH easier to debug and understand. You want to talk about "easy"? It took me a few days just to understand the previous kooky algorithm before I could begin to improve it.

src/Language/Printer.php
shmax: Good eye. :) 885686e

src/Language/Printer.php
shmax: Agreed, thank you. e8340ae

tests/Language/VisitorTest.php
shmax: 👍  bb30738

src/Language/AST/UnionTypeExtensionNode.php
shmax: It seems that the parser writes an array for this one: https://github.com/webonyx/graphql-php/blob/master/src/Language/Parser.php#L1350

Maybe I shouldn't mess with it in this PR, as its fate seems a bit dubious, but it would be great if someone could fix this, or confirm why it is odd duck out.

shmax:
> It is crucial that the libraries i depend on have accurate type definitions in order for it to work well. @shmax, @simPod and myself are certainly dedicated to it and willing to put in the work.

Amen, brother!

src/Language/Printer.php
shmax: No problem if you really want it, but I chose `p` because I now use it inline all over the place, and it helps keep the code a bit more concise. Think it over a bit more and let me know what you think.

shmax:
I'm doing some research. I started by having another look at the graphql-js library. It sort of dances around the trashing of node members issue by using javascript objects for everything (they are too cool to use classes). The only typing they have is in the form of flow interfaces. I suppose one could argue that they don't violate the flow types, because when they "edit" a node, they first make an empty, typeless object, and copy the fields over one by one.

We don't do it that way. We use `clone`, and create a new instance that has the same class type as the original, and that's where we run into trouble--our types _are_ straight-up violated.

I also decided to see if I could learn something by having a look at some implementations of the graphql spec in other languages to see how they implement their printers: 

### [c#](https://github.com/graphql-dotnet/graphql-dotnet)

Printer implemented with recursion. At least, I think so; I put a breakpoint in their printer and saw the stack pile up, but their code is quite sophisticated, and they use a mixture of callbacks and reflected expression bodies to do some black magic that I don't quite understand, yet (though I do C# in my day job). I'll spend another evening analyzing it; there may be something in there we can adapt. At any rate, it does not use anything similar to the "Visitor.visit" functionality to do its work (it does have a class called `AstPrintVisitor`, but this is a helper defined in the printer code, and does not implement the visitor interface).

### [java](https://github.com/graphql-java/graphql-java)
Uses recursion. No visitor. Very straightforward.

Anyway, that's what I have so far. I'll continue my investigation after work.

composer.json
shmax: @simPod I take it you have an objection..?

simPod: Yup, was discussed previously though. Don't fancy those shortcuts as they're not descriptive, are ambiguous and save nothing anyway as every decent shell fills it for you anyway.

shmax: Ah, I see. I use gitbash, and have never noticed any auto-completion. Am I using a bad shell?

Oh lord, I've stumbled into a holy war. Speaking as someone who is a heavy user of the baseline functionality, I grew tired of having to check composer.json to see if it was "baseline" or "phpstan-baseline" or "stan-baseline" or "phpstanbaseline", etc, and this seemed to lift that block. As for the former being more descriptive, I figure neither "baseline" nor "phpstan-baseline" are going to make sense to anyone who is not familiar with the concept, and if you are, then both are equally informative.

If you're okay with this for now, I'll leave it as-is...

simPod: that's why I only 😢 and passed by. 

It's phpstan baseline tho, not some baseline. If we started to use baseline from another tool, the conflict is already there. Same is with lint, fix, stan etc. In my opinion it's broken already so doesn't really matter.

yes I guess :P

shmax: >  If we started to use baseline from another tool, the conflict is already there

Fair enough, but the way I look at it phpstan baseline is our "default" baseline. If another one gets added in the future, we can add it with a more specific key. That's the way I'm generally used to doing it, but in any case we can certainly revisit all this if new tools do appear and compete for the namespace.

simPod: I fancy open closed principle also in cases like this. That's why I'm against changes like this. BC breaks happen then and DX suffer.

shmax: > and DX suffe

Speaking as a D, the X was already suffering. But your point is taken and understood.

shmax:
I've been reading up on the [Visitor](https://en.wikipedia.org/wiki/Visitor_pattern) pattern, and actually it does seem like a very good fit for exactly this kind of work (blogs often mention pretty-printing ASTs). A few things puzzle me, though; none of the academic descriptions of the pattern that I've seen have really mention anything about the presence or absence of recursion--the pattern is more about adding functionality to a collection of classes (that may or may not be related, but are often hierarchical) without modifying the classes, and without, say, creating subclasses of each node type. So I think one of our problems--and I think @simPod will agree with me here, as he's a fan of open/closed--is that we are violating the core principle of the pattern; we are modifying the classes. We are changing the types of their members  and converting them to raw storage. I'm not totally sure, but so far, from what I understand, my recursive Print algorithm is in fact an implementation of the Visitor pattern; it's just not using our Visitor class. So I'm going to keep studying here, and see if I can come up with a way to merge the best of both worlds together and come up with something that makes everybody happy. Please bear with me for a few days (or more)...

vladar:
I did a quick benchmark and the Printer from this PR is ~2-3 times faster than the existing Printer with approximately equal memory usage (with shallow nesting - up to 1000 depth).

It approaches the existing Printer and has increased memory consumption as nesting level increases (1000+ depth) but for any real-world usage, it is faster indeed.

I tested it with PHP7.4. So you are correct that recursion is not a big problem anymore as it used to be in PHP5+.

`Printer` itself is not a bottleneck but I thought I should share this as this is an interesting observation.

Benchmark source code bellow (`Printer2` is the one from this PR, `Printer` is the standard one):

```php
<?php
namespace GraphQL\Benchmarks;

use GraphQL\Benchmarks\Utils\QueryGenerator;
use GraphQL\Benchmarks\Utils\SchemaGenerator;
use GraphQL\Language\AST\FieldNode;
use GraphQL\Language\AST\NodeKind;
use GraphQL\Language\AST\NodeList;
use GraphQL\Language\AST\SelectionSetNode;
use GraphQL\Language\Parser;
use GraphQL\Language\Printer;
use GraphQL\Language\Printer2;
use GraphQL\Language\Visitor;

/**
 * @BeforeMethods({"setUp"})
 * @OutputTimeUnit("milliseconds", precision=3)
 * @Warmup(1)
 * @Revs(2)
 */
class PrinterBench
{
    const MAX_DEPTH = 100;

    private $query;

    public function setUp()
    {
        $depth = 0;
        $ast = Parser::parse("{ field }");

        $this->query = Visitor::visit($ast, [
            NodeKind::FIELD => function (FieldNode $node) use (&$depth) {
                if ($depth++ > self::MAX_DEPTH) {
                    return;
                }
                $editedNode = clone $node;
                $editedNode->selectionSet = new SelectionSetNode(['selections' => new NodeList([clone $node])]);
                return $editedNode;
            }
        ]);
    }

    /**
     * @Iterations(2)
     */
    public function benchPrinter()
    {
        Printer::doPrint($this->query);
    }

    /**
     * @Iterations(2)
     */
    public function benchNewPrinter()
    {
        Printer2::doPrint($this->query);
    }
}
```

shmax:
> I did a quick benchmark and the Printer from this PR is ~2-3 times faster than the existing Printer with approximately equal memory usage (with shallow nesting - up to 1000 depth).

Well, that is certainly encouraging! Thank you very much for your analysis. Would you still like me to continue down the path of looking into integrating `Visitor.php`?

src/Language/AST/UnionTypeExtensionNode.php
shmax: Fixed here: https://github.com/webonyx/graphql-php/pull/694

composer.json
vladar: Pragmatism vs purity @shmax ? 🤔 Sorry, couldn't resist temptation 😄

vladar:
> Would you still like me to continue down the path of looking into integrating Visitor.php?

Not sure what do you mean by this? We must avoid breaking changes in Visitor. But maybe we can get back to this for the next major - along with PHP version bump.

shmax:
> Not sure what do you mean by this? We must avoid breaking changes in Visitor. But maybe we can get back to this for the next major - along with PHP version bump.

One of your original complaints about this architecture was that it uses recursion. Since `Visitor` does not use recursion (I don't think, anyway...), I thought I might try to investigate using it to correct the problem in a more "correct" fashion. I have no idea if it makes sense without doing major surgery on Visitor, but I'm willing to try.

composer.json
shmax: Shots fired! I love you guys, but it seems like in a lot of these discussions we fall into the false dichotomy trap. We are not limited to choosing between "baseline" and "phpstan-baseline". I'm not trying to impress anybody, here, but at work I'm used to seeing script sections that are far, far more complex than this (with 20 or more commands). They usually wind up taking on a recursive form, where broadly-named scripts invoke the more specific ones. You could have your `phpstan-baseline` script, and have the "default" `baseline` script invoke it. Any other baseline tools that appear likewise get specific names, and they may or may not be invoked by the default one. 

I rather like the simplicity of what we have, but we don't have to sweat blood over this (I could always just alias it in my shell, anyway). If ya'll would rather sort this out in some other PR I don't object.

spawnia:
@shmax mind changing the title to something more descriptive of what this PR actually does (rewrite the printer).

shmax:
Mergest master, but now seeing this error from phpstan:

```
 ------ ----------------------------------------------------------------------- 
  Line   src/Executor/ReferenceExecutor.php                                     
 ------ ----------------------------------------------------------------------- 
  351    Switch condition type (true) does not match case condition $selection  
         instanceof \GraphQL\Language\AST\FieldNode (false).                    
  351    Switch condition type (true) does not match case condition $selection  
         instanceof \GraphQL\Language\AST\FragmentSpreadNode (false).           
  351    Switch condition type (true) does not match case condition $selection  
         instanceof \GraphQL\Language\AST\InlineFragmentNode (false).           
 ------ ----------------------------------------------------------------------- 
```

I use the exact same sort of construct in `Printer.php` without issue, and I didn't touch ReferenceExecutor.php. How did I break this?

spawnia:
> I use the exact same sort of construct in `Printer.php` without issue, and I didn't touch ReferenceExecutor.php. How did I break this?

Maybe the other error is covered by the baseline?

shmax:
> Maybe the other error is covered by the baseline?

Didn't see anything like that in there--I think it has something to do with me changing the type of `SelectionSetNode.selections` from `SelectionNode[]` to `NodeList<SelectionSetNode>` (that should probably be `NodeList<SelectionNode>`)...

spawnia:
> Didn't see anything like that in there--I think it has something to do with me changing the type of `SelectionSetNode.selections` from `SelectionNode[]` to `NodeList<SelectionSetNode>` (that should probably be `NodeList<SelectionNode>`)...

PHPStan, saving the day once more. This also showed that uncontrolled usage of the baseline *is evil* and should be avoided. I hope we can clear it one day.

shmax:
> PHPStan, saving the day once more. This also showed that uncontrolled usage of the baseline is evil and should be avoided. I hope we can clear it one day.

Absolutely, and I'm working away at it, but I think this one was my fault. In any case, I fixed the annotated type and the error went away (and I was able to remove another wad of lines from the baseline).

shmax:
Can I ask what the blocker is on this one (other than the conflicts)? I don't want to get another long fight started, I just want to know what the remaining objections are, if any, so I can address them.

vladar:
@shmax Sorry for the delay on this. I feel that a shallow code review is not enough to get this merged. I need to pull this version and play with various use-cases to make sure we don't miss something important here.

Probably will do this the next time I spend a dedicated weekend or so for the library. Sadly, had too little time recently. 

But if this PR blocks someone's work or is critical for some project, please describe your use-case here, I may reconsider priorities.

shmax:
No blockers, other than general progress towards 0 phpstan errors. I appreciate that you want to do more of a deep dive, but the tests do pass, and we've both confirmed a significant performance increase, so I hope this doesn't sit for too much longer. If you can give me some idea of what kind of further tests you want to do, perhaps I can help. Put me to work!

vladar:
Basically my plan for this issue is:

1. Check for updates on `graphql-js` side of things to ensure we will be able to follow them with this approach

2. Take a couple of complex real-world projects and run the printer before/after to see the difference in both output and performance (my benchmarks above are pretty synthetic and naive)

shmax:
> 1. Check for updates on `graphql-js` side of things to ensure we will be able to follow them with this approach

Well, if you like I can reorganize my switch statement so that the cases are in the same order as the keys in the `graphql-js` config object. Would that make you happier?

spawnia:
Hey @shmax. I have had a burst of energy and a bunch of free time today, so I went ahead and fixed the conflicts, tightened the types and got this merge request working again.

As always, pretty nice to see how much defensive programming we can get rid of by using strict, static types.

spawnia (approved):
As part of a new major version, this will provide a nice boost to performance and maintainability. Since we will go through a period of beta testing anyways with a new version, the risk of merging this is very small.

tests/Language/VisitorTest.php
spawnia: Tightening the types in `Printer` bubbled up a bunch more of those violations.

shmax:
> Hey @shmax. I have had a burst of energy and a bunch of free time today, so I went ahead and fixed the conflicts, tightened the types and got this merge request working again.

@spawnia You are an absolute legend, thank you. I was dreading revisiting this, and it means a lot to me that you took up the reins. Let me know if there's anything I can do to help, and thanks again!

spawnia:
I just tried comparing both printers in our large project. The old one crashed, this one produced a pretty good result 😃

shmax:
Nice! The baseline file continues to dwindle...

PR #684, opened 2020-06-27
Optimize

shmax:
Hi again. I've been working with some profilers (phpbench and blackfire), and I'm afraid that while my lazy load feature does speed up my application, the performance of normal non-lazy loaded types has suffered a bit. Apologies for that, I really feel awful about it. One of the culprits is that new `Schema.resolveType` method; in my app it was getting called over 25k times (!). I came up with a few simple one-liners in this PR to lessen the pain (and drop the number of calls by an order of magnitude). Here's some profiler data to look at:

Before lazy-load-type branch:

```
\GraphQL\Benchmarks\HugeSchemaBench

    benchSchema.............................I0 [μ Mo]/r: 144.375 144.375 (ms) [μSD μRSD]/r: 0.000ms 0.00%
    benchSchemaLazy.........................I0 [μ Mo]/r: 0.027 0.027 (ms) [μSD μRSD]/r: 0.000ms 0.00%
    benchSmallQuery.........................I0 [μ Mo]/r: 190.369 190.369 (ms) [μSD μRSD]/r: 0.000ms 0.00%
    benchSmallQueryLazy.....................I0 [μ Mo]/r: 174.710 174.710 (ms) [μSD μRSD]/r: 0.000ms 0.00%

\GraphQL\Benchmarks\LexerBench

    benchIntrospectionQuery.................I4 [μ Mo]/r: 6.979 7.021 (ms) [μSD μRSD]/r: 0.069ms 0.98%

\GraphQL\Benchmarks\StarWarsBench

    benchSchema.............................I1 [μ Mo]/r: 0.522 0.522 (ms) [μSD μRSD]/r: 0.002ms 0.47%
    benchHeroQuery..........................I1 [μ Mo]/r: 2.923 2.923 (ms) [μSD μRSD]/r: 0.043ms 1.47%
    benchNestedQuery........................I1 [μ Mo]/r: 6.756 6.755 (ms) [μSD μRSD]/r: 0.026ms 0.38%
    benchQueryWithFragment..................I1 [μ Mo]/r: 7.702 7.702 (ms) [μSD μRSD]/r: 0.068ms 0.88%
    benchStarWarsIntrospectionQuery.........I1 [μ Mo]/r: 61.422 61.422 (ms) [μSD μRSD]/r: 0.205ms 0.33%

10 subjects, 19 iterations, 170 revs, 0 rejects, 0 failures, 0 warnings
(best [mean mode] worst) = 0.027 [59.578 59.583] 0.027 (ms)
[...]

PR #699, opened 2020-07-08
Revert "Optimize lazy types (#684)"

shmax:
This reverts commit 6096b0174c2b40f3bff9889f3a91dcf071bae933.

@mfn Reported that his graphql-laravel tests are breaking since we merged my optimizations. I think we're going to have to hide some previously public members, and try this again in a major version release.

### DIFF
--- a/src/Type/Definition/FieldArgument.php
+++ b/src/Type/Definition/FieldArgument.php
@@ -30,14 +30,19 @@ class FieldArgument
     /** @var mixed[] */
     public $config;
 
-    /** @var Type&InputType */
+    /** @var InputType&Type */
     private $type;
 
-    /** @param mixed[] $def */
+    /**
+     * @param mixed[] $def
+     */
     public function __construct(array $def)
     {
         foreach ($def as $key => $value) {
             switch ($key) {
+                case 'type':
+                    $this->type = $value;
+                    break;
                 case 'name':
                     $this->name = $value;
                     break;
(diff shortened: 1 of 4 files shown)


kandy:
Can you add link on broken test/build of graphql-laravel?

kandy:
https://github.com/rebing/graphql-laravel/pull/645/checks?check_run_id=848330436

mfn:
There was some errors in the test suite (shame on me 😅 ) but one real break was that, for lazy types, you're not supposed to access the _public_ property `->type` directly, instead you **have** to use the `->getType()` method.

----
I'm now putting on my practical hat: it doesn't matter for graphql-laravel, because there was release yet with `^14` of graphql-php. Insofar, the revert is not strictly necessary.

----
Taking off my practical hat: in strict semver, this is obviously a break in a minor version. But depending on how many already did you use, graphql-php might just get away with it and document that `->getType()` is generally required for `^14`

----
🤷‍♀️

shmax:
@vladar What do you think? Is this revert necessary? Please advise.

vladar:
Can we just add `__get`, `__isset` and `__set` fallback for `type` property strictly for backwards compatibility (with a deprecation notice)?

shmax:
> Can we just add __get, __isset and __set fallback for type property strictly for backwards compatibility (with a deprecation notice)?

You mean as an interim fix until the next major version? Well, maybe, but I opened this PR 12 days ago as an emergency fix for people in the same situation as @mfn. No doubt you were busy with other things, but no one else has come forward since then, so do you think this PR is really necessary?

vladar:
Agreed. Let's move on with deprecation in #702

PR #702, opened 2020-07-09
Fix public types

shmax:
# Background

Recently, I added an optimization that lazy loads all types (whether they are callable or not) on fields. Because the types are no longer written at construction time, it is no longer appropriate or safe to access the `type` fields publicly.

# Changes
* made type private on `FieldDefinition` and `InputObjectField` (it was already private on `FieldArgument` )
* modified a few places in the code to use the getters

This is indeed an API change, so I imagine some kind of version bump will be in order.

### DIFF
--- a/src/Type/Definition/FieldDefinition.php
+++ b/src/Type/Definition/FieldDefinition.php
@@ -6,6 +6,7 @@ namespace GraphQL\Type\Definition;
 
 use GraphQL\Error\Error;
 use GraphQL\Error\InvariantViolation;
+use GraphQL\Error\Warning;
 use GraphQL\Language\AST\FieldDefinitionNode;
 use GraphQL\Type\Schema;
 use GraphQL\Utils\Utils;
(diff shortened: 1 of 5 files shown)


shmax:
> I suggest changing the variable to private $resolvedType and add a fallback __get, __isset and __set methods that catch $type variable usage with a deprecation notice.

If it's just a temporary thing until the next major version I guess I can go along with that, but I hope we can agree that in general we should be trying to move away from magic methods, for a variety of reasons:

* no native typehinting
* it's not completely compatible with the old `public` form (for example, things like `property_exists` don't work)
* static checking can't do much with it (though we can write extensions to handle it)
* no intellisense

In any case, my development computer is in the shop at the moment, but it should be back with me by this weekend. I'll give your approach a try then.

In the meantime, I would be grateful if you could work on reviewing (and hopefully merging) some of my other PRs...

vladar:
> If it's just a temporary thing until the next major version

Yes, it's just a way to gracefully deprecate things. We'll remove them in the next major.

shmax:
> Yes, it's just a way to gracefully deprecate things. We'll remove them in the next major.

Okay, then. I'll give it a try when I get my other machine back.

vladar:
I suggest using https://github.com/webonyx/graphql-php/blob/master/src/Error/Warning.php (maybe even `warnOnce`) for the deprecation message.

shmax:
@vladar I added the magic deprecations. PTAL.

shmax:
> Does it work like this?

Sure seems to--I stepped through in the debugger. But I'll add a test...

shmax:
Added deprecation tests for `FieldDefinition::type` and `InputObjectField::type`

shmax:
@vladar Please review.

shmax:
Thanks much for reviewing and merging 👍

mfn:
FTR, this change again broke https://github.com/rebing/graphql-laravel 😅 => https://github.com/rebing/graphql-laravel/runs/984425603?check_suite_focus=true#step:11:21

I just discovered it, no conclusion yet.

mfn:
The problem is `\GraphQL\Type\Definition\InputObjectField::__set`

graphql-laravel works by setting arbitrary fields (=properties), e.g. a field called `alias` => https://github.com/rebing/graphql-laravel/blob/5e232370c577d3f3ac34b78ca8e812404484eb1f/tests/Unit/AliasAguments/Stubs/ExampleValidationInputObject.php#L24

Since non-existing properties are now funneled through __set but it only handles `type`, setting these custom fields doesn't no anything and thus breaks graphql-laravel.

I.e. the following _would_ fix this:
```diff
diff --git a/vendor/webonyx/graphql-php/src/Type/Definition/InputObjectField.php b/vendor/webonyx/graphql-php/src/Type/Definition/InputObjectField.php
index 31ebdb5..bab73e2 100644
--- a/vendor/webonyx/graphql-php/src/Type/Definition/InputObjectField.php
+++ b/vendor/webonyx/graphql-php/src/Type/Definition/InputObjectField.php
@@ -97,6 +97,8 @@ class InputObjectField
                     Warning::WARNING_CONFIG_DEPRECATION
                 );
                 $this->type = $value;
+            default:
+                $this->$name = $value;
         }
     }
```
__isset/__get/__set are added for multiple classes and since they don't have a default behaviour except **only** listening for `type`, this breaks any such use.

----
That said, when I look at `\GraphQL\Type\Definition\InputObjectField` _I guess_ the proper fix from my side would be to use `\GraphQL\Type\Definition\InputObjectField::$config` instead.

Tried via https://github.com/rebing/graphql-laravel/pull/665 and seems to look good.

shmax:
Hi, thanks for reporting. I'll open yet another PR, but the review/merge cycle is very slow on this repo, so your best route forward is going to be to do a workaround on your end. That said, doing arbitrary writes like this sounds like a very dangerous practice to me, and if you can avoid it you probably should.

PR #706, opened 2020-07-15
Differentiate between client-safe and non-client-safe errors in scalar validation

mfn:
## Summary
This reverts commit ce834e28

I'm creating this PR in light of discussion happened in https://github.com/webonyx/graphql-php/pull/614 after the release to lobby for reverting it.

See my arguments from https://github.com/webonyx/graphql-php/pull/614#issuecomment-656896591
> Why this change? Why wouldn't what is essential type validation not deemed client safe suddenly?
>
> The impact is that when I provide a wrong type for an input arg before I got (for example):
>
> - `Field "someMutation" argument "activated" requires type Boolean, found "false".`
>   but now I just get
> - `Internal server error`
> 
> Even for the _consumer_ of the API this is unfriendly.
> 
> Any exception passed to `\GraphQL\Error\Error` in `$previous` marks it as `internal`. Previously it was not passed, thus it was of type `graphql` and thus `isClientSafe=true`.
> 
> I would like to better understand the arguments _for_ this change as I think this should be reverted.


spawnia:
Simply reverting is taking it back to far.

The optimal solution is to allow both client-safe and non-client-safe exceptions to propagate properly from within scalar coercion.

The validation for built-in scalars is client-safe and should be changed to throw appropriate exceptions. Also, we should add a test for a scalar that throws an unsafe exception. After all, there was a reason why the original PR was committed.

mfn:
Maybe, but IMHO it's unfair to but the burden on the one realizing the (for me, my app) breaking change and having to defend the change back compared to the original change which was got kind of a free pass without any questioning.

If someone wants to improve this: be my guest, but please don't put this on me.

Maybe I got it all wrong, waiting for @vladar for his views, thanks 😄

spawnia:
I can do the work, do you mind if i hijack your PR?

mfn:
Not at all!

src/Error/TypeError.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace GraphQL\Error;
+
+use Exception;
+
+/**
+ * Caused if a user passes a wrong type.
+ */
+class TypeError extends Exception implements ClientAware
```
spawnia, line 12: Not sure if we should use `UserError` instead or something else entirely. I can not use `Error`, since it requires passing a message, which is explictly *not* done.

vladar: Good point. We should also preserve `graphql` category for those kinds of errors as they are described in the GraphQL spec: https://spec.graphql.org/draft/#sec-Values-of-Correct-Type (and also for backwards compatibility)

So maybe just make `message` in the `Error` optional and use it instead?

P.S. This whole piece of catching and wrapping is confusing though. Maybe worth a refactor someday.

PR #709, opened 2020-07-21
Base CHANGELOG.md on https://keepachangelog.com/en/1.0.0/

spawnia:
Makes it easier to recognize what kind of changes are in a given release.

@vladar Started from v14 onwards, I think the older entries are fine as is.


CHANGELOG.md
```diff
@@ -1,33 +1,42 @@
...
+
+### Optimized
 - Visitor: simplify getVisitFn (#694)
 - Replace function calls with type casts (#692)
-- Fix "only booleans are allowed" errors (#659)
-
 
 ## v14.0.2
 
+### Optimized
```
spawnia, line 29: Not a part of the standard, but I think it is worthwhile to distinguish between functional changes and non-functional optimizations/refactorings.

vladar:
Oh, this is good! I like it. Thanks for the suggestion!

CHANGELOG.md
vladar: Yeah, I see they have a discussion about it: https://github.com/olivierlacan/keep-a-changelog/issues/247

PR #712, opened 2020-08-04
Use `foreach` over slower functions `array_map()` and `Utils::map()`

spawnia:
This increases the performance measured in the benchmarks by about 1-3% on average.

While there are some spots where the map functions may look a bit nicer, other parts were complicated quite a bit by its use (e.g. double `array_map()` into `array_combine()`).


spawnia:
> I just don't want to think about the type of iterable, which you kinda always have to do with `array_map`

I totally agree with that sentiment. However, in PHP this kind of flexibility comes with a runtime cost. Manual monomorphization can be tedious, but PHPStan and the test suite should have us covered here.

I will try and produce some benchmark results.

spawnia:
Added a benchmark that does a lot of schema parsing, triggering code within `ASTDefinitionBuilder`. The `foreach` variant consistently performed better, around 1-3%.

> I just don't want to think about the type of iterable, which you kinda always have to do with array_map

We could just convert everything to `foreach`, which is always the fastest.

I think that particularly in library code, such micro-optimizations *do* matter. Personal gusto or ease of writing becomes less important when you consider the (as of writing this) 6652 users that are currently wasting CPU cycles: https://github.com/webonyx/graphql-php/network/dependents?package_id=UGFja2FnZS01NDI5NjAyMjc%3D

generate-class-reference.php
```diff
@@ -130,9 +133,12 @@ function unwrapDocblock($docBlock, $stripAnnotations = true) {
 
 function unpadDocblock($docBlock) {
     $lines = explode("\n", $docBlock);
-    $lines = \GraphQL\Utils\Utils::map($lines, function($line) {
-        return ' ' . trim($line);
-    });
+    $lines = array_map(
+        static function (string $line): string {
```
simPod, line 137: we can use short version here

src/Executor/Promise/Adapter/ReactPromiseAdapter.php
```diff
@@ -76,15 +75,17 @@ public function createRejected($reason)
     /**
      * @inheritdoc
      */
-    public function all(array $promisesOrValues)
+    public function all(array $promisesOrValues): Promise
     {
         // TODO: rework with generators when PHP minimum required version is changed to 5.5+
```
simPod, line 80: maybe OT but what about the todo?

src/Type/Definition/InterfaceType.php
```diff
@@ -104,12 +103,14 @@ public function getInterfaces(): array
                 );
             }
 
-            /** @var array<int, InterfaceType> $interfaces */
-            $interfaces = $interfaces === null
-                ? []
-                : array_map([Schema::class, 'resolveType'], $interfaces);
-
-            $this->interfaces = $interfaces;
+            $this->interfaces = [];
```
simPod, line 106: I guess it's not a problem now but shouldn't we do atomic swap instead?

src/Utils/Utils.php
```diff
@@ -155,28 +154,6 @@ public static function filter($iterable, callable $predicate): array
...
     }
 
-    /**
-     * @param iterable<mixed> $iterable
-     *
-     * @return array<mixed>
-     *
-     * @throws Exception
-     */
-    public static function map($iterable, callable $fn): array
```
simPod, line 165: I'm very ok with dropping this. But `array_map()` should be almost of the same speed as foreach in PHP 8 and sometimes `array_map()` is more readable IMO

https://stackoverflow.com/a/64460563/519333

src/Executor/Promise/Adapter/ReactPromiseAdapter.php
spawnia: Not quite sure how the code should be changed and if there is any benefit to be had.

src/Type/Definition/InterfaceType.php
spawnia: I only see that leading to more code would even guess that the intermediary variable costs performance.

src/Utils/Utils.php
spawnia: As a foundational library, I think it is a good choice for us to forgo readability in favor of performance when in question. It is nice to see that the speed of abstractions is catching up. If there was a generic `map` which performed better for all `iterable`, I would be all for it.

src/Type/Definition/InterfaceType.php
simPod: The intermediary variable should be removed by php optimalizations.

i'm used to doing that when working with multiple threads. It is not a thing in php yet but might be in the future (maybe even with Fibers in 8.1?)

src/Utils/Utils.php
simPod: Yup, dropping this utility is just fine but e.g. this https://github.com/webonyx/graphql-php/pull/712/files#diff-17372694d53aa9749dbda064d15720cf6fb7df4f071c97c26d8d692a66d39a9bL205-L208 I'd keep because the performance should be the same on 8.0 already and is more readable.

src/Type/Definition/ObjectType.php
```diff
@@ -148,10 +147,14 @@ public function getInterfaces(): array
...
-
-            $this->interfaces = $interfaces;
+            $this->interfaces = [];
+            if (is_array($interfaces)) {
+                foreach ($interfaces as $interface) {
+                    /** @var InterfaceType $interfaceType Ensured to be true during schema validation */
+                    $interfaceType      = Schema::resolveType($interface);
+                    $this->interfaces[] = $interfaceType;
+                }
+            }
```
simPod, line 157:
```suggestion
            foreach ($interfaces ?? [] as $interface) {
                /** @var InterfaceType $interfaceType Ensured to be true during schema validation */
                $interfaceType      = Schema::resolveType($interface);
                $this->interfaces[] = $interfaceType;
            }
```

also seems more readable without any cost

src/Utils/AST.php
```diff
@@ -536,28 +533,20 @@ public static function valueFromASTUntyped($valueNode, ?array $variables = null)
...
-                        },
-                        iterator_to_array($valueNode->fields)
-                    )
-                );
+                $values = [];
+                foreach ($valueNode->fields as $field) {
+                    $values[$field->name->value] = self::valueFromASTUntyped($field->value, $variables);
+                }
+
+                return $values;
```
simPod, line 549: This case is IMO more readable with foreach 👍🏾

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -407,13 +407,13 @@ private function makeUnionDef(UnionTypeDefinitionNode $def): UnionType
...
             // values below, that would throw immediately while type system
             // validation with validateSchema() will produce more actionable results.
-            'types'       => function () use ($def): array {
-                return Utils::map(
-                    $def->types,
-                    function ($typeNode): Type {
-                        return $this->buildType($typeNode);
-                    }
-                );
+            'types'       =>  function () use ($def): array {
```
simPod, line 410: ok, CS should have caught this additional space :/

src/Utils/SchemaExtender.php
```diff
@@ -262,9 +261,10 @@ protected static function extendValueMap(EnumType $type): array
      */
     protected static function extendPossibleTypes(UnionType $type): array
     {
-        $possibleTypes = array_map(static function ($type) {
-            return static::extendNamedType($type);
-        }, $type->getTypes());
+        $possibleTypes = [];
+        foreach ($type->getTypes() as $possibleType) {
+            $possibleTypes[] = static::extendNamedType($possibleType);
+        }
```
simPod, line 267:
```suggestion
        $possibleTypes = array_map(
           static fn ($type) => static::extendNamedType($type),
           $type->getTypes())
        );
```
in many cases here, this would be my way to go

src/Utils/Utils.php
spawnia: Restored some of the short `array_map`.

src/Type/Definition/InterfaceType.php
spawnia: It is done.

src/Utils/ASTDefinitionBuilder.php
spawnia: Agree

src/Utils/SchemaExtender.php
spawnia: It is now so ;)

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -81,17 +80,17 @@ public function __construct(
 
     public function buildDirective(DirectiveDefinitionNode $directiveNode): Directive
     {
+        $locations = [];
```
simPod, line 83: maybe we can inline it using array_map to 'locations' key

src/Utils/ASTDefinitionBuilder.php
simPod:
```suggestion
            'types'       => function () use ($def): array {
```

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -407,13 +407,13 @@ private function makeUnionDef(UnionTypeDefinitionNode $def): UnionType
...
             // validation with validateSchema() will produce more actionable results.
-            'types'       => function () use ($def): array {
-                return Utils::map(
-                    $def->types,
-                    function ($typeNode): Type {
-                        return $this->buildType($typeNode);
-                    }
-                );
+            'types'       =>  function () use ($def): array {
+                $types = [];
```
simPod, line 411: nitpicking. array map also suits here

simPod (commented):
should I target some places where I think we can use map in extra branch so you can cherrypick?

src/Error/Error.php
```diff
@@ -39,59 +36,55 @@ class Error extends Exception implements JsonSerializable, ClientAware, Provides
...
-    private $positions;
+    /** @var array<int, int>|null */
+    private ?array $positions;
 
     private bool $isClientSafe;
 
-    /** @var array<string, mixed> */
-    protected array $extensions;
+    /** @var array<string, mixed>|null */
+    protected ?array $extensions;
```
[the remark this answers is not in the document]
spawnia: Consistency across the other props here that can maybe hold an `array`. While an empty array is an acceptable default for most cases, I think the optionality of this is best expressed by `null`. We don't go around and add empty strings as defaults for optional strings everywhere either.

examples/01-blog/Blog/Data/DataSource.php
```diff
@@ -147,12 +147,17 @@ public static function findLastStoryFor(int $authorId): ?Story
     public static function findLikes(int $storyId, int $limit): array
     {
         $likes = self::$storyLikes[$storyId] ?? [];
-        $users = array_map(
-            static fn (int $userId) => self::$users[$userId],
-            $likes
-        );
 
-        return array_slice($users, 0, $limit);
+        $result = [];
```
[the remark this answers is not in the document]
spawnia: reverted

spawnia:
> should I target some places where I think we can use map in extra branch so you can cherrypick?

Sounds good. I would like to get this one merged if it is alright in order to limit the amount of WIP merge requests. I do plan to follow up with removal of other, expensive utils.

simPod (approved):
Aight, let's go 🚢 . I'll follow with something on top of this soon.

Issue #711, opened 2020-08-03
Validation messages translation


Erindi:
Hi, I want to add translations to your validation rules messages, but I can't find a right way to do it more or less "acceptable". Can you help me a bit with this problem?

As far as i can see almost all your validation rules has static methods with messages which return dynamic text. And almost all this static methods used by their own class objects in non static methods using self:: .
For example: 
```
class ValuesOfCorrectType extends ValidationRule
{
    public function getVisitor(ValidationContext $context)
    {
            // ...

                $context->reportError(
                    new Error(
                        self::unknownFieldMessage($parentType->name, $node->name->value, $didYouMean),
                        $node
                    )
                );

            // ...
    }

    // ...

    public static function unknownFieldMessage($typeName, $fieldName, $message = null)
    {
        return sprintf('Field "%s" is not defined by type %s', $fieldName, $typeName) .
            ($message ? sprintf('; %s', $message) : '.');
    }

    // ...
}
```
I can extend any ```ValidationRule``` class give it proper ```$name``` and rewrite it in ```DocumentValidator::$defaultRules``` array before validation using ```DocumentValidator::addRule()``` method. But i can't only redefine ValidationRules static methods with error mesages and do something like
```
class ValuesOfCorrectType extends \GraphQL\Validator\Rules\ValuesOfCorrectType
{
    private static TranslatorInterface $translator;

    protected $name = 'GraphQL\Validator\Rules\ValuesOfCorrectType';

    // ...

    public static function unknownFieldMessage($typeName, $fieldName, $message = null)
    {
        return sprintf(static::$translator->trans('Field "%s" is not defined by type %s'), $fieldName, $typeName) .
            ($message ? sprintf('; %s', $message) : '.');
    }
}
```
because this methods called via ```self::```  and not via ```static::``` (but even if ```static::``` was used, this is also not the best way to add translations to messages) and static methods is not comfortable to translate if you don't  have a static translator (i use symfony/translation 5.0). 

All i can do in this situation for now, as i can see it, is to make a full copy of every validation class that i want to translate to my project, make all necessary changes and rewrite it via ```DocumentValidator::addRule()```. But that is not a proper way to do it, because of updates and etc.

Maybe there is another way to solve this problem? Or maybe you working on some changes for validation messages (like some messages storage class where i can redefine any validation messages)?

Thank you in advance!

vladar:
[...]

PR #713, opened 2020-08-05
Validation messages text in class public static properties (#711)

Erindi:


src/Validator/Rules/ValuesOfCorrectType.php
```diff
@@ -253,24 +271,24 @@ static function (EnumValueDefinition $value) : string {
                 )
             );
 
-            return $suggestions ? 'Did you mean the enum value ' . Utils::orList($suggestions) . '?' : null;
+            return $suggestions ? static::$enumTypeSuggestionMessage . Utils::orList($suggestions) . '?' : null;
```
spawnia, line 274: I think this one should be replaced with `sprintf`, grammar may not be the same in all languages.

Erindi: Replaced.

shmax:
Is this something to do with localization of the error messages? If so, I'm not sure I can really recommend this approach. You're adding a huge number of new public properties, expanding the maintenance surface, and spreading the feature all across the code base. I would suggest an approach that keeps the details of each class insular, centralizes the error strings for easier reference, and exposes a single point where localization can happen.

I'm sure there are a million ways to do this, but I was involved in this exact same work on another repo I used to contribute to years ago, and here's what I came up with then:

https://github.com/justinrainbow/json-schema/pull/364/files

In short:
* used an enum class to define the error codes
* all of the default error message strings are in one place
* when logging an error, instead of adding the error message on the spot, we add the error code and the vars (to later pass to sprintf)

Just some ideas to get you thinking.

Erindi:
> Is this something to do with localization of the error messages? If so, I'm not sure I can really recommend this approach. You're adding a huge number of new public properties, expanding the maintenance surface, and spreading the feature all across the code base. I would suggest an approach that keeps the details of each class insular, centralizes the error strings for easier reference, and exposes a single point where localization can happen.
> 
> I'm sure there are a million ways to do this, but I was involved in this exact same work on another repo I used to contribute to years ago, and here's what I came up with then:
> 
> https://github.com/justinrainbow/json-schema/pull/364/files
> 
> In short:
> 
>     * used an enum class to define the error codes
> 
>     * all of the default error message strings are in one place
> 
>     * when logging an error, instead of adding the error message on the spot, we add the error code and the vars (to later pass to sprintf)
> 
> 
> Just some ideas to get you thinking.

Hi, I've created an [issue](https://github.com/webonyx/graphql-php/issues/711). Got an answer that translation for this lib is not an important case and @vladar [don't want to complicate things](https://github.com/webonyx/graphql-php/issues/711#issuecomment-668505205) because of translation and proposed a minor changes so that we had at least some opportunity to translate the text of errors. @spawnia proposed another option with also minor changes [that prevents silly errors](https://github.com/webonyx/graphql-php/issues/711#issuecomment-668653299), i agreed to this terms and did this PR to get a solution for my problem in a way acceptable to you.
I know that there are a million ways to do this better but how I thought it belonged to the category of "complicate things", for not so worthy cause. And this PR here is only because this is some fast and "better than none" solution for my project that was some kind of acceptable for you in my issue.

spawnia:
> this is some fast and "better than none" solution

@Erindi I agree with that assessment and think we can merge this as-is.

@shmax We do not have to guarantee stability of those messages forever. If someone decides to put in the time to develop a better translation system, we can change it in the next major release.

shmax:
@spawnia @Erindi This is not a "minor change". You touched 32 files. You added 50+ new public properties to the API surface that we will have to maintain, and eventually break once one of us takes the time to support this properly. 

But the point is that this is simply not how you localize a code base. With this approach a consumer of the library would have to figure out how to get access to each of these classes so he can set those strings, and it would do absolutely nothing for client side projects (eg. Javascript).

If @Erindi is in that much of a hurry he can just fork the repo and merge his branch there. Let's not make a mess of this one.

All that said, I'm happy to support this feature properly, but let's take the time to understand it and come up with a solution that is mature, robust, and a little more accessible to other consumers of this library. I'm free all weekend.

shmax:
I see that we already have an `extensions` property on our `Error` class, but it doesn't currently seem to be used for anything (other than some tests that write "foo" to it).

shmax:
Here's the 1-hour version: https://github.com/webonyx/graphql-php/pull/718

spawnia:
I noticed that `sprintf` based internationalization is fundamentally flawed, sentence structure can differ between languages.

PR #715, opened 2020-08-13
Read getParsedBody() instead of getBody() when Request is ServerRequest

simPod:
Follows up https://github.com/webonyx/graphql-php/pull/598


simPod:
Sure, if this is alright, I'll cover it.

vladar:
Looks good to me. @PowerKiKi will it work for you?

PowerKiKi:
Thank you for reaching out. I meant to come back to this earlier, but it slipped off my mind.

I'll take some time to test this over the weekend and will get back to you.

PowerKiKi:
This work well for me. Thank you for taking the time fix this regression.

However I still think that `parsePsrRequest()` signature should explicitly accept a `ServerRequestInterface`, because it is semantically correct.

According to [the spec](https://www.php-fig.org/psr/psr-7/#32-psrhttpmessagerequestinterface), `RequestInterface` is:

> Representation of an **outgoing**, client-side request.

Which is incorrect for our use-case. This lib does not send outgoing requests to other targets. But instead it does accept incoming requests and create a response for them. And that is exactly what `ServerRequestInterface` [is for](https://www.php-fig.org/psr/psr-7/#321-psrhttpmessageserverrequestinterface):

> Representation of an **incoming**, server-side HTTP request.

Also hardcoding "polymorphism" via `instanceof` looks like code smell to me. Especially in this case where it was not needed before and then complexity was introduced.

This also introduce the new responsibility to this lib to be able to parse incoming request body. I would rather let that responsibility to dedicated libraries. (Yes, if this PR is accepted, it will allow to parse via another lib, but the point is to avoid setting expectation for this lib and instead be clear that parsing should be done by another lib).

Finally, I fail to see why the [original issue](https://github.com/webonyx/graphql-php/pull/598#issuecomment-569672667) was not solved outside of this library by something as simple as:

```php
function processOutgoingRequest(\Psr\Http\Message\RequestInterface $outgoingRequest)
{
    $incomingRequest = new \Laminas\Diactoros\ServerRequest();
    $incomingRequest = $incomingRequest
        ->withBody($outgoingRequest->getBody());

    $server = new StandardServer([/* ... */]);
    foreach ($outgoingRequest->getHeaders() as $key => $value) {
        $incomingRequest = $incomingRequest->withHeader($key, $value);
    }

    return $server->executePsrRequest($incomingRequest);
}
```

tl;dr: i'm fine with merging this because it un-breaks things, but it is a sub-optimal solution

PR #716, opened 2020-08-17
Fix default get/set behavior on InputObjectField and FieldDefinition

shmax:
As reported by @mfn, my recent addition of `__get`/`__set` magic methods to `FieldDefinition` and `InputObjectFIeld` as caused the portions of his code that rely on the getting and setting of undefined properties to fail. This PR restores the old default functionality.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -86,8 +86,13 @@ parameters:
 			path: src/Type/Definition/EnumType.php
 
 		-
+			message: "#^Variable property access on \\$this\\(GraphQL\\\\Type\\\\Definition\\\\FieldDefinition\\)\\.$#"
+			count: 2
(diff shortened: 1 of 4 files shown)


mfn:
@shmax I think you missed to adapt `__isset()`, it exhibits the same problem

shmax:
>  I think you missed to adapt __isset(), it exhibits the same problem

You're the boss. e2137cf

mfn:
LGTM 👍 

Like the last time, after figuring out the problem I already prepared a fix in the library via https://github.com/rebing/graphql-laravel/pull/665 and won't need this anymore, it's for "all the others".

I still haven't made a release for `^14` 😅 but I'm in fact glad because that would also cause pain for the users of the lib which I could avoid so far by holding out.

shmax:
@vladar Please review

PR #720, opened 2020-08-24
Enhance type definitions for Executor

spawnia:
I converted Executor and ReferenceExecutor to use native type hints where possible and cleaned up a few mismatched types. Turns out a few of them, if translated 1-1, caused type errors.

Now we can be sure the types are correct and will have an easier time when trying to understand what is happening in the code and when refactoring.

### DIFF
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -20,7 +20,7 @@ use function is_object;
  */
 class Executor
 {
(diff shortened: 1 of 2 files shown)


shmax:
Pretty keen! 👍 

No changes to the baseline file..?

vladar (approved):
Great stuff, thanks! 👍

PR #722, opened 2020-08-28
fix $debug type in StandardServer::send500Error()

janzimmermannnetrex:
`$debug` in `StandardServer::send500Error()` should be `int` in the same way as in definition of the `FormattedError::createFromException()`

### DIFF
--- a/src/Server/StandardServer.php
+++ b/src/Server/StandardServer.php
@@ -4,9 +4,9 @@ declare(strict_types=1);
 
 namespace GraphQL\Server;
 
+use GraphQL\Error\DebugFlag;
 use GraphQL\Error\FormattedError;
 use GraphQL\Error\InvariantViolation;
-use GraphQL\Error\DebugFlag;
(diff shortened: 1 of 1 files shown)


src/Server/StandardServer.php
```diff
@@ -50,12 +51,12 @@ class StandardServer
...
      *
      * @param Throwable $error
-     * @param bool      $debug
+     * @param int       $debug
      * @param bool      $exitWhenDone
      *
      * @api
      */
-    public static function send500Error($error, $debug = false, $exitWhenDone = false)
+    public static function send500Error($error, $debug = DebugFlag::NONE, $exitWhenDone = false)
```
spawnia, line 59:
```suggestion
    public static function send500Error(Throwable $error, int $debug = DebugFlag::NONE, bool $exitWhenDone = false): void
```

spawnia (approved):
The change makes sense, I would go ahead and convert to proper type hints.

simPod (approved):

Issue #279, opened 2018-05-04
Accept requests where `query` is a parsed AST representation


terion-name:
Hello.

When using `apollo` or `weaver` graphql tools (as proxies and gateways), they are parsing input graphql and sending already parsed request. This works ok with node-js or scala servers, but with php server (this one) it breaks with `GraphQL query body is expected to be string, but got array at /app/vendor/webonyx/graphql-php/src/Utils/Utils.php:231`.
Why this queries are not supported (while reference implementation supports it) and is there a way to make it work? In other words, is there a way to map this structure to DocumentNode?

Example request body:
```
{
  "query": {
    "kind": "Document",
    "definitions": [
      {
        "kind": "OperationDefinition",
        "operation": "query",
        "variableDefinitions": [],
        "selectionSet": {
          "kind": "SelectionSet",
          "selections": [
            {
              "kind": "Field",
              "name": {
                "kind": "Name",
                "value": "municipalityByDomain",
                "loc": {
                  "start": 28,
                  "end": 48
                }
              },
              "arguments": [
                {
                  "kind": "Argument",
                  "name": {
                    "kind": "Name",
                    "value": "domain",
                    "loc": {
                      "start": 49,
                      "end": 55
                    }
                  },
                  "value": {
                    "kind": "StringValue",
                    "value": "foo",
                    "block": false,
                    "loc": {
                      "start": 57,
                      "end": 62
                    }
                  },
                  "loc": {
                    "start": 49,
                    "end": 62
                  }
                }
              ],
              "directives": [],
              "selectionSet": {
                "kind": "SelectionSet",
                "selections": [
                  {
                    "kind": "Field",
                    "name": {
                      "kind": "Name",
                      "value": "id",
                      "loc": {
                        "start": 72,
                        "end": 74
                      }
                    },
                    "arguments": [],
                    "directives": [],
                    "loc": {
                      "start": 72,
                      "end": 74
                    }
                  }
                ]
              },
              "loc": {
                "start": 28,
                "end": 80
              }
            }
          ]
        }
      }
    ]
  },
  "variables": {},
  "context": {
    "graphqlContext": {}
  }
}
```

terion-name:
For now I've handled it like this, but it more seems like a hack:

```php
protected function convertGraphQL($input)
    {
        if (is_array($input)) {
            foreach ($input as $k => $v) {
                $input[$k] = $this->convertGraphQL($v);
            }
            if (isset($input['kind'])) {
[...]

PR #725, opened 2020-08-31
Support DocumentNode, AST in Helper

shmax:
Fix for #279 

Recently I've started experimenting with [graphql-tag](https://github.com/apollographql/graphql-tag). It's a node package popular with users of [apollographql](https://github.com/apollographql?type=source) that does the parsing of graphql queries in the js layer (mainly so that Apollo can do merging and other operations on the AST, which is difficult on a string query).

As noted in issue #279, this library does not directly support an AST structure. I'm aware that you can doctor the incoming request using `AST::fromArray` before passing it to `\GraphQL\GraphQL::executeQuery`, but my original client code was using `StandardServer` and this does not expose any hook where this kind of preprocessing can happen.

So, in this PR I add knowledge of both `DocumentNode` and array representations of AST to `Helper`.

I'll wait for the green light before adding tests.

~~edit: note that I didn't actually change anything in the `GraphQL::executeQuery` entry point; wasn't sure how much push back I'm going to get on this, but I'm happy to modify it as well if anyone wants it.~~ Added the functionality.

### DIFF
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -15,6 +15,7 @@ use GraphQL\Executor\Promise\Promise;
 use GraphQL\Executor\Promise\PromiseAdapter;
 use GraphQL\GraphQL;
 use GraphQL\Language\AST\DocumentNode;
+use GraphQL\Language\AST\NodeKind;
 use GraphQL\Language\Parser;
 use GraphQL\Utils\AST;
 use GraphQL\Utils\Utils;
@@ -166,9 +167,9 @@ class Helper
             $errors[] = new RequestError('GraphQL Request parameters "query" and "queryId" are mutually exclusive');
         }
 
-        if ($params->query !== null && ! is_string($params->query)) {
+        if ($params->query !== null && ! (is_string($params->query) || $params->query instanceof DocumentNode || AST::isAst($params->query))) {
             $errors[] = new RequestError(
-                'GraphQL Request parameter "query" must be string, but got ' .
+                'GraphQL Request parameter "query" must be string, a DocumentNode, or an array representation of an AST, but got ' .
                 Utils::printSafeJson($params->query)
             );
         }
@@ -283,7 +284,7 @@ class Helper
 
             $doc = $op->queryId
                 ? $this->loadPersistedQuery($config, $op)
-                : $op->query;
+                : (AST::isAst($op->query) ? AST::fromArray($op->query) : $op->query);
 
             if (! $doc instanceof DocumentNode) {
                 $doc = Parser::parse($doc);
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -166,7 +166,13 @@ class Helper
             $errors[] = new RequestError('GraphQL Request parameters "query" and "queryId" are mutually exclusive');
         }
 
-        if ($params->query !== null && ! (is_string($params->query) || $params->query instanceof DocumentNode || AST::isAst($params->query))) {
+        if ($params->query !== null
+            && ! (
+                is_string($params->query)
+                || $params->query instanceof DocumentNode
+                || AST::isAst($params->query)
+            )
+        ) {
             $errors[] = new RequestError(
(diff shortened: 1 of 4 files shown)


shmax:
@vladar @simPod @spawnia Hey folks, please review.

src/Utils/AST.php
```diff
@@ -118,6 +118,14 @@ public static function fromArray(array $node) : Node
         return $instance;
     }
 
+    /**
+     * @param mixed[] $node
```
spawnia, line 122: Since this method is used as a type guard to check if whatever is passed in is an AST, I think that any type should be fine.

```suggestion
     * @param mixed $maybeNode
```

src/Server/Helper.php
```diff
@@ -283,7 +283,7 @@ static function (RequestError $err) : Error {
 
             $doc = $op->queryId
                 ? $this->loadPersistedQuery($config, $op)
-                : $op->query;
+                : (AST::isAst($op->query) ? AST::fromArray($op->query) : $op->query);
```
spawnia, line 286:
```suggestion
                : (AST::isAst($op->query)
                    ? AST::fromArray($op->query)
                    : $op->query);
```

src/Server/Helper.php
```diff
@@ -166,9 +166,9 @@ public function validateOperationParams(OperationParams $params)
             $errors[] = new RequestError('GraphQL Request parameters "query" and "queryId" are mutually exclusive');
         }
 
-        if ($params->query !== null && ! is_string($params->query)) {
+        if ($params->query !== null && ! (is_string($params->query) || $params->query instanceof DocumentNode || AST::isAst($params->query))) {
```
spawnia, line 169:
```suggestion
        if ($params->query !== null
            && ! (
                is_string($params->query)
                || $params->query instanceof DocumentNode
                || AST::isAst($params->query)
            )
        ) {
```

shmax:
@spawnia Any idea what's going on with some of those checks in the matrix? A few of them fail, but I don't see any information on the details page; it just says "This check failed": https://github.com/webonyx/graphql-php/pull/725/checks?check_run_id=1097480255

Something to do with exercising different executors? I couldn't figure out how to switch them around in my local environment; I tried setting a the "executor" env var to "coroutine", but I don't see any calls to `getEnv` in the test flow. Any advice?

simPod:
might be github status issue 🤔

shmax:
> might be github status issue 🤔

Oh, hey @simPod, haven't seen you around in a while! You think something is awry on Github's end?

simPod:
hi @shmax :P been busy with some personal things lately. I tried to retrigger here and all passes https://github.com/simPod/graphql-php/actions/runs/249356988 so must be github. I restarted ur checks.

shmax:
I guess you've got the magic touch! Thanks very much.

simPod (approved):

vladar (commented):
Makes sense to me in general (as I mentioned in the original issue). One concern I have is security. `AST::fromArray` was originally implemented for "safe" usage scenarios (like persisting AST on disk and then reading back).

I don't feel confident enough to expose it out of the box as we don't know how it will behave with malicious inputs.

So what we need with this PR:

1. Make this feature opt-in and only in the `StandardServer` (should not be added to `GraphQL` facade - we can document how to convert an array to DocumentNode for those who really needs it)

2. Do some kind of security audit for `AST::fromArray` and probably add tests for various malicious and invalid inputs.

shmax:
> So what we need with this PR:

Eh, I'm going to pass. I don't have time to be run through the wringer over this. I've already removed `StandardServer` from my own code. I was just trying to be helpful.

vladar:
I understand. But if you use this `AST::fromArray` utility in your project with `graphql-tag`, keep in mind that it can be unsafe.

shmax:
Well, never say never, but I use persisted queries, so in my particular case there are no queries being formed on the client side and in principle there should be no risk.

PR #726, opened 2020-09-01
Fix the `@var` type hint for various `NodeList<Node>` usages.

zimzat:
- Remove ` = []` from a couple of usages, as this is now considered invalid typing.
- Adjust a few boolean checks to explicitly check for ` === null` now
    - (the default when not set during manual construction prior to typed properties).

### Discussion:
With regard to ` = []` and/or `|null`/` = null`, what approach would you want the library to take?

If you want the library to start using `public NodeList $interfaces`, after PHP 7.4 becomes the minimum version, then attempting to access it directly would cause uninitialized errors unless it's `public ?NodeList $interfaces = null`. Ensuring it's always initialized with the expected value would allow for fluid object access. At initial glance I don't see anywhere things are _explicitly_ set to null (though I could be very wrong) so maintaining nullability doesn't seem necessary?

Direction:
- Find a way to ensure they're always initialized as NodeList during construction?
    - Extending the `__construct()` method to do something like `$this->interfaces = $this->interfaces ?? new NodeList([])` is one possibility.
- Mark _all_ of them as `|null` and default to ` = null`?
- Other?

I'd lean toward ensuring they're initialized as NodeList, but could see going some other way depending on where you see the architecture going.

Thanks!

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -311,17 +311,7 @@ parameters:
 			path: src/Utils/ASTDefinitionBuilder.php
 
 		-
-			message: "#^Only booleans are allowed in an if condition, array\\<GraphQL\\\\Language\\\\AST\\\\NamedTypeNode\\> given\\.$#"
-			count: 1
-			path: src/Utils/ASTDefinitionBuilder.php
-
-		-
 			message: "#^Only booleans are allowed in a ternary operator condition, GraphQL\\\\Language\\\\AST\\\\NodeList\\<GraphQL\\\\Language\\\\AST\\\\EnumValueDefinitionNode\\>\\|null given\\.$#"
-			count: 1
-			path: src/Utils/ASTDefinitionBuilder.php
-
-		-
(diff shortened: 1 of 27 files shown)


simPod (approved):
About the initialization, here's some context https://github.com/webonyx/graphql-php/pull/675

spawnia:
> Remove = [] from a couple of usages, as this is now considered invalid typing.

Nicely done!

> With regard to = [] and/or |null/ = null, what approach would you want the library to take?

Some properties can legimately not exist in some nodes, in which case defaulting initialization to `null` seems fine to me.

For properties that must not be `null` but may not always be initialized after construction, we should avoid typing them as nullable. We can use `isset()` to safely check if it is initialized (or null):

```php
class Foo {
    var $bar;
}

$foo = new Foo();
var_dump(isset($foo->bar));
// bool(false)

$foo->bar = null;
var_dump(isset($foo->bar));
// bool(false)
```

This plays nicely with PHP 7.4 typed properties, which I would like to use at some point - probably far into the future.

PR #731, opened 2020-11-14
Update Type.php

murtukov:
Is there a reason to do this weird double check:
```php
if ((static::$standardTypes[self::ID] ?? null) === null)
```
instead of just this:
```php
if (!isset(static::$standardTypes[self::ID]))
```

`$value ?? $anotherValue` is just a sugar syntax for `isset($value) ? $value : $anotherValue`


spawnia (approved):
I think this was added as a result of applying more strict static validation, transitioning from `if ($x[$y] ?? null)`.

Using `isset()` seems cleaner to me.

simPod (approved):

PR #740, opened 2020-11-26
Implement support for interfaces implementing interfaces

### LINKED ISSUE #728: Allow interfaces to implement other interfaces
v15 of graphql-js implemented an addition to the GraphQL spec which allows interfaces to implement other interfaces. It would be nice to be able to do this in PHP as well. For example when there is a generic Node interface and a specialised Media Node type needs to be created for images, videos, etc.

graphql-js implementation: https://github.com/graphql/graphql-js/pull/2084
spec addition: https://github.com/graphql/graphql-spec/pull/373

Article on why this is useful: https://dev.to/mikemarcacci/intermediate-interfaces-generic-utility-types-in-graphql-50e8

The following GraphQL schema snippet currently doesn't parse but would be expected to parse after implementation.

```gql
interface Node {
  uuid: ID!
}

interface Media implements Node {
  uuid: ID!
  url: String!
}

type Image implements Media & Node {
  uuid: ID!
  url: String!
}
```

Kingdutch:
**What's new**
This adds the option to have interfaces implement other interfaces. I’ve mostly looked at the interface implementation of normal `type`. A lot of the work in the parser will continue to work similarly because the specification forces objects to implement all interfaces that a parent interface may implement.

**Tests**
Tests have been implemented according to https://github.com/graphql/graphql-js/pull/2084/files

**Backwards Compatibility**
There was some discussion in #728 about whether this should be postponed to 15.0 or could be included included in 14.x. From what I can tell so far, all code changes are either backwards compatible, or internal to the library (private in a class). Additionally, the changes don't make any schema that's currently valid, invalid. It does allow schema's that were previously invalid to be valid, but that's not a breaking change :)

**Cheers!**
I want to commend the previous contributors to this project on the comments littered throughout the codebase. It has helped me using the library in extending projects but also made it easy to understand what is going on while working in the project. I think this project could be a great way for people to learn about ASTs and parsers :D 

Closes #728


Kingdutch:
Tests are currently blocked by #741

spawnia:
This library sticks pretty close to the reference implementation. For the tests, you should be able to pretty much copy https://github.com/graphql/graphql-js/pull/2084

Kingdutch:
I've added the tests from the linked PR :D Let's see if this works in one go or I made a booboo somewhere (or it reveals an implementation error) ^^

Kingdutch:
Wow, that process was a lot rougher than I expected. Who would've guessed it isn't easy to reliably apply a JavaScript diff to a PHP project ^^

It looks like the only thing left is fixing a bug that's now present in the parser and a missing validation for breaking changes. I'll see if I can get to that after dinner. If you don't see activity from me in the next few hours, then feel free to pick it up this weekend, or I'll continue on Monday :)

Kingdutch:
It looks like the implementation of the schema parser for interfaces becamse a bit more complicated. All that's needed is to reset the changes to `src/Type/Schema.php` and properly implement the changes from https://github.com/graphql/graphql-js/pull/2084/files#diff-a01d9baf9db0301e57d4671db353a2a9f07c56236ad7eee4a2fa980672afdeb3

If no one beats me to it, I'll do that with a clear head on Monday :)

Kingdutch:
I can't figure out the following errors.
```
"Abstract type Named must resolve to an Object type at runtime for field Person.friends with value "instance of GraphQL\Tests\Executor\TestClasses\Person", received "null". Either the Named type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function."

"Abstract type Named must resolve to an Object type at runtime for field Person.friends with value "instance of GraphQL\Tests\Executor\TestClasses\Dog", received "null". Either the Named type should provide a "resolveType" function or each possible type should provide an "isTypeOf" function."
```

Both in `src/Executor/ReferenceExecutor.php(976)` for test `GraphQL\Tests\Executor\UnionInterfaceTest::testAllowsFragmentConditionsToBeAbstractTypes`

This could be related to

```
"Call to a member function isTypeOf() on bool"
```

Which is called twice at `/src/Executor/ReferenceExecutor.php(944)` in `AbstractPromiseTest::testIsTypeOfUsedToResolveRuntimeTypeForUnion` it shows up in that same locaiton for `AbstractPromiseTest::testIsTypeOfCanBeRejected` and `AbstractPromiseTest::testIsTypeOfUsedToResolveRuntimeTypeForInterface  ` but I'm unable to find where the value is set to a boolean.

Kingdutch:
Okay it looks like I had made a mistake in a copy paste of one of the tests as well as missed changing `Schema::getPossibleTypes` which are now both fixed. It looks like that's it!

Given that most of the tests have failed at some point of me trying to port this from JS I'm pretty confident that it works now that all the tests (including the new ones) are passing! 🚀

src/Type/Definition/ImplementingType.php
```diff
@@ -0,0 +1,21 @@
...
+GraphQLObjectType |
+GraphQLInterfaceType;
+*/
+
+interface ImplementingType
+{
+    public function implementsInterface(InterfaceType $interfaceType) : bool;
+
+    /**
+     * @return InterfaceType[]
```
spawnia, line 18:
```suggestion
     * @return array<int, InterfaceType>
```

...assuming the result is not supposed to be a map?

src/Type/Definition/InterfaceType.php
```diff
@@ -27,6 +30,20 @@ class InterfaceType extends Type implements AbstractType, OutputType, CompositeT
...
+     * Lazily initialized.
+     *
+     * @var InterfaceType[]
+     */
+    private $interfaces;
+
+    /**
+     * Lazily initialized.
+     *
+     * @var InterfaceType[]
```
spawnia, line 43:
```suggestion
     * @var array<string, InterfaceType>
```

src/Type/Definition/InterfaceType.php
```diff
@@ -27,6 +30,20 @@ class InterfaceType extends Type implements AbstractType, OutputType, CompositeT
      */
     private $fields;
 
+    /**
+     * Lazily initialized.
+     *
+     * @var InterfaceType[]
```
spawnia, line 36:
```suggestion
     * @var array<int, InterfaceType>
```

src/Type/Definition/InterfaceType.php
```diff
@@ -27,6 +30,20 @@ class InterfaceType extends Type implements AbstractType, OutputType, CompositeT
...
+     */
+    private $interfaces;
```
spawnia, line 38: Maybe rename to `interfaceList` to differentiate from the map?

src/Type/Schema.php
```diff
@@ -416,55 +426,105 @@ public static function resolveType($type) : Type
...
-        if ($abstractType instanceof InterfaceType) {
-            return $possibleType->implementsInterface($abstractType);
-        }
+        return $this->isSubType($abstractType, $possibleType);
+    }
 
-        if ($abstractType instanceof UnionType) {
-            return $abstractType->isPossibleType($possibleType);
+    /**
+     * Returns true if maybe sub type is a sub type of given abstract type.
```
spawnia, line 500: I don't mind the variable name, but *maybe* confused me in this sentence. It reads as if the method just makes a guess.

src/Type/SchemaValidationContext.php
```diff
@@ -654,30 +659,47 @@ private function getFieldArgNode($type, $fieldName, $argName)
         return $nodes[0] ?? null;
     }
 
-    private function validateObjectInterfaces(ObjectType $object)
+    /**
+     * @param ObjectType|InterfaceType $type
+     */
+    private function validateInterfaces(ImplementingType $type)
```
spawnia, line 665:
```suggestion
    private function validateInterfaces(ImplementingType $type) : void
```

src/Type/SchemaValidationContext.php
```diff
@@ -694,23 +716,25 @@ private function getDirectives($object)
...
 
     /**
-     * @param InterfaceType $iface
+     * @param ObjectType|InterfaceType $type
+     * @param InterfaceType            $iface
      *
      * @return NamedTypeNode|null
      */
-    private function getImplementsInterfaceNode(ObjectType $type, $iface)
+    private function getImplementsInterfaceNode($type, $iface)
```
spawnia, line 724:
```suggestion
    private function getImplementsInterfaceNode($type, InterfaceType $iface) : ?NamedTypeNode
```

src/Type/SchemaValidationContext.php
```diff
@@ -845,28 +870,62 @@ private function validateObjectImplementsInterface(ObjectType $object, $iface)
...
                 );
             }
         }
     }
 
+    /**
+     * @param ObjectType|InterfaceType $type
+     * @param InterfaceType            $iface
+     */
+    private function validateTypeImplementsAncestors(ImplementingType $type, $iface)
```
spawnia, line 899:
```suggestion
    private function validateTypeImplementsAncestors(ImplementingType $type, InterfaceType $iface) : void
```

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -329,7 +329,10 @@ private function getDeprecationReason($node)
         return $deprecated['reason'] ?? null;
     }
 
-    private function makeImplementedInterfaces(ObjectTypeDefinitionNode $def)
+    /**
+     * @param ObjectTypeDefinitionNode | InterfaceTypeDefinitionNode $def
+     */
+    private function makeImplementedInterfaces($def)
```
spawnia, line 335: What does this return?

spawnia (changes requested):
Looks quite solid overall, great effort.

I am assuming the naming somewhat follows the reference implementation where it is new. At least it matches the existing conventions well.

Type definitions can be refined quite a bit. Distinguishing between `array<int, T>` and `array<string, T>` cleans up one of PHPs biggest sins.

src/Type/Definition/ImplementingType.php
Kingdutch: This has the same return type as `ObjectType:getInterfaces` (and `ObjectType::$interfaces`) since that's what this interface is based on (and then implemented by both `ObjectType` and `InterfaceType`).

src/Type/Definition/InterfaceType.php
Kingdutch: `ObjectType` follows the convention of having `$interfaces` and `$interfaceMap` so I think consistency would point to keeping `interfaces` for the list type (alternatively, we'd also rename the `$interfaces` variable in ObjectType, which is private).

src/Type/Schema.php
Kingdutch: It could be that I was the one making a guess to a description. I'm pretty clear on what the method does but not necessarily on how to describe it (the reference JS implementation is quite lacking in the comments area)

src/Type/Definition/InterfaceType.php
spawnia: Alright. We don't really suffix with `List` anywhere else, so I think it can stay like this.

src/Type/Definition/ImplementingType.php
spawnia: Alright, then my suggestion seems right.

src/Type/Schema.php
spawnia:
```suggestion
     * Returns true if the given type is a sub type of the given abstract type.
```

src/Utils/ASTDefinitionBuilder.php
Kingdutch: This either returns `null` or an array `Type` instances (which is what `$this->buildType` officially returns). If we assume a correct implementation of `$this->buildType` then we expect the return to be `?array<int,InterfaceType>`

spawnia: Given we have a comprehensive test suite, you can just put in the proper type hints and we will see if they hold true. If not, that is even better and gives us a chance to clarify or fix a bug.

src/Type/Definition/ImplementingType.php
Kingdutch: Cool, I'll include the change for the ObjectType as well then, since it implements the interface.

src/Utils/ASTDefinitionBuilder.php
Kingdutch:
```suggestion
    private function makeImplementedInterfaces($def) : ?array
```

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -329,7 +329,10 @@ private function getDeprecationReason($node)
         return $deprecated['reason'] ?? null;
     }
 
-    private function makeImplementedInterfaces(ObjectTypeDefinitionNode $def)
+    /**
+     * @param ObjectTypeDefinitionNode | InterfaceTypeDefinitionNode $def
+     */
```
Kingdutch, line 334:
```suggestion
     * @param ObjectTypeDefinitionNode | InterfaceTypeDefinitionNode $def
     *
     * @return array<int, InterfaceType> | null
     */
```

src/Type/SchemaValidationContext.php
```diff
@@ -694,23 +716,25 @@ private function getDirectives($object)
...
 
     /**
-     * @param InterfaceType $iface
+     * @param ObjectType|InterfaceType $type
+     * @param InterfaceType            $iface
      *
      * @return NamedTypeNode|null
      */
-    private function getImplementsInterfaceNode(ObjectType $type, $iface)
+    private function getImplementsInterfaceNode($type, InterfaceType $iface) : ?NamedTypeNode
```
Kingdutch, line 724: @spawnia This function is used in case of interface extension errors in SchemaValdationContext.php:676 (`validateInterfaces`) which causes a type error in our tests. There is a test that specifically checks that a schema where an object implements a non-interface doesn't validate (that scenario would hit this code path).

I think this is a bug in the `validateObjectInterfaces` implementation which shouldn't call `getImplementsInterfaceNodes` there with a known non-interface. However, I'm unsure what the context for `reportError` in terms of `$nodes` should be instead.

Kingdutch: The only thing `getAllImplementsInterfaceNodes` actually needs from the `$iface`-like object is a name property. The method was providing the relevant `NamedTypeNode` instance (rather than not returning anything at all) so removing that context would be a step backwards.

I've widened the type to `Type` to make a step in the direction of type safety without breaking existing functionality. I've had some other places where I needed `ImplementingType` but also needed a `name` property and thus had to fallback to an annotation of `InterfaceType | ObjectType`. A good follow-up to this issue would be to introduce a generic `TypeInterface` which defines things shared by all type implementations. The `ImplementingInterface` could extend this so we can accept that interface directly instead of a union of interface and object. (That's not needed in PHP 8 where we could just do `ObjectType | InterfaceType` which may be more desirable since it's debatable whether custom implementations of those types are useful 🤷‍♂️ )

spawnia:
@Kingdutch mind if I modify the PR to clarify type signatures? That helps me understand the changes better.

Kingdutch:
@spawnia Go for it! That's why "allow modifications by maintainers" is enabled :)

src/Type/SchemaValidationContext.php
spawnia: Alright. We can stick with `Type`, that works fine. I adapted the naming to indicate it is not always an `InterfaceType`.

spawnia (approved):
Nice work, big effort 👍

Kingdutch:
@spawnia It looks like your last changes introduced a coding standards error, any idea why?

spawnia:
> @spawnia It looks like your last changes introduced a coding standards error, any idea why?

I think the array shape was the culprit: https://github.com/webonyx/graphql-php/pull/740/commits/602bf4eb8ad208f5deedf43f24a3d57ce1f75d44

Kingdutch:
Cool! Then this is all green again and ready to be committed :D

src/Utils/BreakingChangesFinder.php
```diff
@@ -41,14 +42,14 @@ class BreakingChangesFinder
...
-    public const BREAKING_CHANGE_INTERFACE_REMOVED_FROM_OBJECT = 'INTERFACE_REMOVED_FROM_OBJECT';
+    public const BREAKING_CHANGE_IMPLEMENTED_INTERFACE_REMOVED = 'IMPLEMENTED_INTERFACE_REMOVED';
     public const BREAKING_CHANGE_DIRECTIVE_REMOVED             = 'DIRECTIVE_REMOVED';
     public const BREAKING_CHANGE_DIRECTIVE_ARG_REMOVED         = 'DIRECTIVE_ARG_REMOVED';
     public const BREAKING_CHANGE_DIRECTIVE_LOCATION_REMOVED    = 'DIRECTIVE_LOCATION_REMOVED';
     public const BREAKING_CHANGE_REQUIRED_DIRECTIVE_ARG_ADDED  = 'REQUIRED_DIRECTIVE_ARG_ADDED';
     public const DANGEROUS_CHANGE_ARG_DEFAULT_VALUE_CHANGED    = 'ARG_DEFAULT_VALUE_CHANGE';
     public const DANGEROUS_CHANGE_VALUE_ADDED_TO_ENUM          = 'VALUE_ADDED_TO_ENUM';
-    public const DANGEROUS_CHANGE_INTERFACE_ADDED_TO_OBJECT    = 'INTERFACE_ADDED_TO_OBJECT';
+    public const DANGEROUS_CHANGE_IMPLEMENTED_INTERFACE_ADDED  = 'IMPLEMENTED_INTERFACE_ADDED';
```
vladar, line 52: Sadly, changing those constant is a subtle breaking change.

vladar (commented):
That's awesome. Such a huge work! Thank you 👍 

There is one small breaking change in `BreakingChangesFinder` 😄 (see my comment below) It makes me hesitant about publishing it as a minor. What if we revert constant names for 14.x?

This way we'll be able to publish it as minor - then we can rename them in 15.0 and document properly in the migration guide.

So my recommendation would be: 
1. Rename those constraints back in this PR - then we can merge it and publish as a minor
2. Open a separate PR with new names for those constants and a note in the UPGRADE.md. We will merge this PR in 15.0.0

src/Utils/BreakingChangesFinder.php
Kingdutch: Would you accept an alternative:

Add the new name and use it in our internal code. However, change the value of the old name to the new name and add a deprecation notice for the use of that variant.

This allows existing 14.x consumers to start using the new constant and should not break any existing users of the current constant (if they're using matching the string literal rather than the content then they're already violating the API, since that should be considered an internal representation).

Then the only change to make in 15.0 is to remove the old constant. Any consumer that is already using the new constant then doesn't need to make any additional changes and is compatible with 14.x and 15.0.

Kingdutch: I've added my proposed change in a new commit :)

vladar: Good idea 👍 

But I've seen cases where values of similar constants were used as translation keys. So while technically it is an implementation detail it also easily leaks outside.

Anyway, I think it is fine for a minor version as long as we mention it in the changelog. I can live with that.

Kingdutch:
@vladar will you add the changelog or should I do it?

src/Utils/BreakingChangesFinder.php
```diff
@@ -41,17 +42,19 @@ class BreakingChangesFinder
     public const BREAKING_CHANGE_ARG_CHANGED_KIND              = 'ARG_CHANGED_KIND';
     public const BREAKING_CHANGE_REQUIRED_ARG_ADDED            = 'REQUIRED_ARG_ADDED';
     public const BREAKING_CHANGE_REQUIRED_INPUT_FIELD_ADDED    = 'REQUIRED_INPUT_FIELD_ADDED';
-    public const BREAKING_CHANGE_INTERFACE_REMOVED_FROM_OBJECT = 'INTERFACE_REMOVED_FROM_OBJECT';
+    public const BREAKING_CHANGE_IMPLEMENTED_INTERFACE_REMOVED = 'IMPLEMENTED_INTERFACE_REMOVED';
```
vladar, line 45: This one too. Sorry, I should've highlighted this earlier.

Kingdutch: No problem, here you go :D

vladar (approved):
Thank you!

PR #741, opened 2020-11-27
Update to shivammathur/setup-php@2.9.0

Kingdutch:
**Problem**
`set-env` and `add-path` are disabled in GitHub actions. 
See https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

This causes the tests in all PRs to fail until resolved.

**Solution**
This requires an update to the action using those
commands. This means at least version 2.6.0 is
needed. See https://github.com/shivammathur/setup-php/pull/300

However there seem to be no breaking changes so we
can go straight to 2.9.0 instead. This also allows testing
with version 8.x of PHP.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -58,7 +58,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Install PHP
-      uses: shivammathur/setup-php@2.2.2
+      uses: shivammathur/setup-php@2.9.0
       with:
         php-version: 7.1
         coverage: none
@@ -89,7 +89,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Install PHP
-      uses: shivammathur/setup-php@2.2.2
+      uses: shivammathur/setup-php@2.9.0
       with:
         php-version: 7.1
         coverage: none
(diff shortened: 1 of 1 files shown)


Kingdutch:
The coding Standard, PHPStan and Code Coverage failures are caused by using the `master` as source for the workflow, thus using the older, broken, version of setup-php. All other tests successfully passed with the 2.9.0 version from this PR.

simPod (commented):
The version need to be updated for each job

Kingdutch:
I had totally missed there were more jobs, fixed it now. Couldn't amend the previous commit locally so had to edit through the GitHub UI. Please squash this during merge :D

simPod (approved):

.github/workflows/ci-build.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Install PHP
-      uses: shivammathur/setup-php@2.2.2
+      uses: shivammathur/setup-php@2.9.0
```
mfn, line 27: FTR, ever since I've been using setup-php it's been extremely stable and the author is very diligent. Aka never have I experienced issues by just using `setup-php@2` and benefit from minor updates on the fly.

simPod: Yup but 💩  happens, I guess it's better to lock to minor at least to have reproducible builds.

Issue #742, opened 2020-11-30
PHP 8 syntax error


brandonfarber:
https://php.watch/versions/8.0/deprecate-required-param-after-optional

Required parameters declared after optional parameters now throw a deprecation notice.

https://github.com/webonyx/graphql-php/blob/master/src/Error/FormattedError.php#L280

You can just get rid of the `= null` in the argument list. `callable` is already nullable via the `?` mark

spawnia:
Can you create a PR?

PR #743, opened 2020-12-01
Fix PHP 8 deprecation: Required parameters after optional parameter

brandonfarber:
PHP8 produces a deprecation notice when required parameters are defined following optional parameters. This PR addresses this concern in `Error\FormatteError::prepareFormatter()`, the only place I identified the issue occurring in my tests.

Fixes: #742

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Fix:
 - Parse `DirectiveDefinitionNode->locations` as `NodeList<NamedNode>` (fixes AST::fromArray conversion) (#723)
 - Parse `Parser::implementsInterfaces` as `NodeList<NamedTypeNode>` (fixes AST::fromArray conversion)
 - Fix signature of `Parser::unionMemberTypes` to match actual `NodeList<NamedTypeNode>`
+- Fix signature of `Error\FormattedError::prepareFormatter()` to address PHP8 deprecation (#742)
 
 #### v14.3.0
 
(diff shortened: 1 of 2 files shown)


simPod (approved):

mfn:
The reason this probably wasn't caught yet is that the suite does not test on PHP 8.0 yet.

There seem to be some (dev!) dependencies, which need to updated to newer versions:
- phpbench
- phpstan
- phpstan-phpunit
- phpstan-strict-rules
- doctrine/coding-standards

simPod:
@mfn IMO it's out of scope here tho. We're eg. blocked by supporting PHP EOLed versions https://github.com/webonyx/graphql-php/pull/470

mfn:
https://github.com/webonyx/graphql-php/pull/745

mfn:
Interestingly https://github.com/webonyx/graphql-php/pull/745 does not trigger this particular issue fixed here

spawnia (approved):
Nicely done, this does not break anything so should be good to go.

mfn (approved):
Agreed

konradoboza:
Hi @spawnia, @brandonfarber would it be possible to backport this fix to 0.13 series of the bundle? I would be up for creating a PR, but wanted to have some confirmation first that this would be even released. ;)

spawnia:
I am not maintaining 0.13 any more.

PR #744, opened 2020-12-01
Drop PHP 7.2 support

simPod:
Same as https://github.com/webonyx/graphql-php/pull/470 but for EOLed PHP 7.2


vladar (approved):
Thanks! Let's get rid of those old versions finally 👍

PR #747, opened 2020-12-03
Fixed SchemaPrinter so that it uses late static binding

fabis94:
Making class methods protected isn't enough to be able to override them, because all of them call each other using `self::` not `static::`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
+#### Unreleased
+
 Fix:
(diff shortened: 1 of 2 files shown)


spawnia (approved):
That opens the door for easier extensibility, nice.

simPod (approved):

fabis94:
@spawnia when can I expect this to be merged/published?

spawnia:
We typically wait for @vladar to show up and merge. This is a small change which I do feel confident about including, but so far that has been our modus operandi.

vladar:
Thank you!

PR #751, opened 2020-12-11
Reify AST node types and remove unneeded nullability

spawnia:
This achieves a few things:

- make the code easier to understand both by humans and static analyzers
- improve performance by eliminating unnecessary null checks
- improve usability through a clarified API

### DIFF
--- a/src/Utils/ASTDefinitionBuilder.php
+++ b/src/Utils/ASTDefinitionBuilder.php
@@ -181,7 +181,7 @@ class ASTDefinitionBuilder
     }
 
     /**
-     * @param string|(Node&NamedTypeNode)|(Node&TypeDefinitionNode) $ref
+     * @param string|(Node &NamedTypeNode)|(Node&TypeDefinitionNode) $ref
      */
     public function buildType($ref) : Type
     {
@@ -193,7 +193,7 @@ class ASTDefinitionBuilder
     }
 
     /**
-     * @param (Node&NamedTypeNode)|(Node&TypeDefinitionNode)|null $typeNode
+     * @param (Node &NamedTypeNode)|(Node&TypeDefinitionNode)|null $typeNode
      *
      * @throws Error
      */
--- a/src/Utils/ASTDefinitionBuilder.php
+++ b/src/Utils/ASTDefinitionBuilder.php
@@ -59,7 +59,9 @@ class ASTDefinitionBuilder
     private $cache;
 
     /**
-     * @param array<string, (Node&TypeDefinitionNode)> $typeDefinitionsMap
+     * code sniffer doesn't understand this syntax. Pr with a fix here: waiting on https://github.com/squizlabs/PHP_CodeSniffer/pull/2919
+     * phpcs:disable Squiz.Commenting.FunctionComment.SpacingAfterParamType
+     * @param array<string, Node&TypeDefinitionNode> $typeDefinitionsMap
      * @param array<string, bool> $options
      */
     public function __construct(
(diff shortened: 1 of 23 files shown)


src/Utils/SchemaExtender.php
```diff
@@ -615,13 +616,13 @@ static function (string $typeName) use ($schema) {
         $types = array_merge(
             // Iterate through all types, getting the type definition for each, ensuring
             // that any type not directly referenced by a field will get created.
-            array_map(static function ($type) {
+            array_map(static function (Type $type) : Type {
                 return static::extendNamedType($type);
-            }, array_values($schema->getTypeMap())),
+            }, $schema->getTypeMap()),
```
simPod, line 621: didn't resetting of those array indexes through `array_values()` play some part?

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -318,95 +319,89 @@ public function buildField(FieldDefinitionNode $field)
...
-                ? function () use ($def) {
-                    return Utils::map(
-                        $def->types,
-                        function ($typeNode) : Type {
-                            return $this->buildType($typeNode);
-                        }
-                    );
-                }
-                : [],
+            'types'       => function () use ($def) {
```
simPod, line 392: we can add return type hint here as well

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -41,28 +43,28 @@
...
     private $resolveType;
 
-    /** @var Type[] */
+    /** @var array<string, Type> */
     private $cache;
 
     /**
-     * @param Node[] $typeDefinitionsMap
-     * @param bool[] $options
+     * @param array<string, Node  &TypeDefinitionNode> $typeDefinitionsMap
```
simPod, line 62:
```suggestion
     * @param array<string, Node&TypeDefinitionNode> $typeDefinitionsMap
```

simPod (approved):
👌🏾

src/Utils/SchemaExtender.php
spawnia: I don't see how, given the index is not used in the `array_map()`

simPod: true, it's not `+`...

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -41,28 +43,28 @@
...
     private $resolveType;
 
-    /** @var Type[] */
+    /** @var array<string, Type> */
     private $cache;
 
     /**
-     * @param Node[] $typeDefinitionsMap
-     * @param bool[] $options
+     * @param array<string, Node&TypeDefinitionNode> $typeDefinitionsMap
```
spawnia, line 62:
```suggestion
     * @param array<string, (Node&TypeDefinitionNode)> $typeDefinitionsMap
```

vladar (approved):
Nice change 👍 But do you think it is safe to publish as a patch? I am a bit hesitant. Inclining to include it in the next minor.

spawnia:
The actual code changes are very minor and would be best classified as bug fixes. If anything, it would uncover implementations which are already bugged in some way.

vladar:
We have conflicts now after merging #740. Mind fixing? Then we can ship it and safely publish it in the new minor 🚢

spawnia:
@vladar done 🚢

vladar (approved):
:shipit:

shmax:
@spawnia @vladar  Great work, guys. 👍 By the way, sorry I haven't been very active lately; work has gotten a bit hectic, and my Big Project using graphql is basically done, but please don't hesitate to ping me if you would like me to look at any issues.

PR #752, opened 2020-12-13
Remove unused imports

spawnia:


mfn:
I was about to suggest why not enable `no_unused_imports` when I realized, it's not using php-cs-fixer; a feature sadly not supported by phpcs 😞

simPod:
@mfn phpcs supports this, I guess it's just not enabled here. #466 blocked by #470

simPod (approved):

mfn:
> phpcs supports this

Does it? Granted, I'm not familiar with phpcs but https://github.com/squizlabs/PHP_CodeSniffer/issues/801 leads me to assume otherwise.

simPod:
@mfn see https://github.com/slevomat/coding-standard#slevomatcodingstandardnamespacesunuseduses-

In PHPCS there IMO should not be any sniffs but the CS tool. Might lead to confusions like this.

mfn:
> confusions

I'm already confused by linking to coding-standard vs. the phpcs issue I linked but .. I totally trust your judgement, I feel already bad for derailing here so much. Apologies.

simPod:
There's phpcs tool for running sniffs and then there are coding standards that are basically rulesets composed of sniffs. Those rulesets are run via phpcs. And one such ruleset that is used here is maintained by Doctrine https://github.com/doctrine/coding-standard. 

One confusing thing is that there are sniffs in phpcs repository https://github.com/squizlabs/PHP_CodeSniffer, I'd separate it into two 🤷🏾‍♀️ . Hope that clears it up a bit ;)

PR #755, opened 2020-12-15
Implement Utils::getOperationAST

### LINKED ISSUE #754: `Utils::getOperation` deviates from graphql-js' getOperationAST
## Problem
The [`Utils::getOperation`](https://github.com/webonyx/graphql-php/blame/master/src/Utils/AST.php#L595) implementation in this library deviates from [`getOperationAST`](https://github.com/graphql/graphql-js/blob/master/src/utilities/getOperationAST.js) in two important ways:

1. It accepts documents with multiple operations without specifying an operationName. This is considered invalid by the graphql-js implementation.
2. It only provides the name of the operation whereas users of the library (such as myself ^^) may want the entire operation instead of just the name.

## Proposed resolution
We have a few options here depending on how much/little we want to break and how we want to approach releasing this in a semver compatible manner.

### Add a new method
Since the name of the operation that's currently implemented also differs we can implement the method from `graphql-js` with the name that it has over there, this also makes discovery easier.

```
    /**
     * Returns the operation given a document and operation name.
     *
     * If a name is not provided, an operation is only returned if only one is
     * provided in the document.
     *
     * @api
     */
    public static function getOperationAST(DocumentNode $document, string $operationName = null) : ?OperationDefinitionNode
    {
        $operation = NULL;
        foreach ($document->definitions->getIterator() as $node) {
            if ($node instanceof OperationDefinitionNode) {
                if ($operationName === NULL) {
                    // If no operation name was provided, only return an Operation if there
                    // is one defined in the document. Upon encountering the second, return
                    // null.
                    if ($operation !== NULL) {
                        return null;
                    }
                    $operation = $node;
                }
                elseif ($node->name instanceof NameNode && $node->name->value === $msg->getOperationName()) {
                    return $node;
                }
            }
        }
        return $operation;
    }
```

The original method (`getOperation`) is left unaffected because using the new method would cause it to fail for documents that contain multiple operations, without an operation name specified. Currently this would default to the first operation.

### Fix the behaviour of `getOperation` too

This would implement `getOperationAST` and adjust `getOperation` to match the expected behaviour of rejecting documents by calling `getOperationAST` and only returning the name. This fixes the behaviour but requires no changes in existing call sites.

### Deprecate `getOperation`
Not necessarily a resolution on its own but one that can be combined with either of the above approaches. This would require changing the call sites in the GraphQL PHP library itself. This allows to remove `getOperation` in the future ensuring we don't have methods that aren't present in the implementation that we follow.

We can also decide not to deprecate this method and let them live side-by-side depending on what's needed, but the difference in behaviour may then become confusing.

Kingdutch:
This follows the getOperationAST from graphql-js and adds the related
test suite. `getOperation` is deprecated in favor of using
`getOperationAST` and getting the name from a non-null result.

Source version: https://github.com/graphql/graphql-js/blob/7b3241329e1ff49fb647b043b80568f0cf9e1a7c/src/utilities/getOperationAST.js
Tests version: https://github.com/graphql/graphql-js/blob/7b3241329e1ff49fb647b043b80568f0cf9e1a7c/src/utilities/__tests__/getOperationAST-test.js

Fixes #754

### DIFF
--- a/src/Utils/AST.php
+++ b/src/Utils/AST.php
@@ -612,10 +612,9 @@ class AST
     }
 
     /**
-     * Returns the operation given a document and operation name.
+     * Returns the operation within a document by name.
      *
-     * If a name is not provided, an operation is only returned if only one is
-     * provided in the document.
+     * If a name is not provided, an operation is only returned if the document has exactly one.
      *
      * @api
      */
@@ -628,9 +627,7 @@ class AST
             }
 
             if ($operationName === null) {
-                // If no operation name was provided, only return an Operation if there
-                // is one defined in the document. Upon encountering the second, return
-                // null.
+                // We found a second operation, so we bail instead of returning an ambiguous result.
                 if ($operation !== null) {
                     return null;
                 }
--- a/src/Utils/AST.php
+++ b/src/Utils/AST.php
@@ -584,6 +584,8 @@ class AST
     }
 
     /**
+     * @deprecated use getOperationAST instead.
+     *
      * Returns operation type ("query", "mutation" or "subscription") given a document and operation name
      *
      * @param string $operationName
@@ -607,5 +609,37 @@ class AST
         }
 
         return false;
+    }
+
+    /**
+     * Returns the operation given a document and operation name.
+     *
+     * If a name is not provided, an operation is only returned if only one is
+     * provided in the document.
+     *
+     * @api
+     */
+    public static function getOperationAST(DocumentNode $document, ?string $operationName = null) : ?OperationDefinitionNode
+    {
+        $operation = null;
+        foreach ($document->definitions->getIterator() as $node) {
+            if (! $node instanceof OperationDefinitionNode) {
+                continue;
+            }
+
+            if ($operationName === null) {
+                // If no operation name was provided, only return an Operation if there
+                // is one defined in the document. Upon encountering the second, return
+                // null.
+                if ($operation !== null) {
+                    return null;
+                }
+                $operation = $node;
+            } elseif ($node->name instanceof NameNode && $node->name->value === $operationName) {
+                return $node;
+            }
+        }
+
+        return $operation;
     }
 }
--- a/tests/Utils/AstGetOperationAstTest.php
+++ b/tests/Utils/AstGetOperationAstTest.php
@@ -18,7 +18,10 @@ class AstGetOperationAstTest extends TestCase
     public function testGetsAnOperationFromASimpleDocument() : void
     {
         $doc = Parser::parse('{ field }');
-        self::assertEquals(AST::getOperationAST($doc), $doc->definitions->offsetGet(0));
+        self::assertEquals(
+            AST::getOperationAST($doc),
+            $doc->definitions->offsetGet(0)
+        );
     }
 
     /**
--- a/tests/Utils/AstGetOperationAstTest.php
+++ b/tests/Utils/AstGetOperationAstTest.php
@@ -1,0 +1,105 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Tests\Utils;
+
+use GraphQL\Language\Parser;
+use GraphQL\Utils\AST;
+use PHPUnit\Framework\TestCase;
+
+class AstGetOperationAstTest extends TestCase
+{
+    // Describe: getOperationAST
+
+    /**
+     * @see it('Gets an operation from a simple document')
+     */
+    public function testGetsAnOperationFromASimpleDocument() : void
+    {
+        $doc = Parser::parse('{ field }');
+        self::assertEquals(AST::getOperationAST($doc), $doc->definitions->offsetGet(0));
+    }
+
+    /**
+     * @see it('Gets an operation from a document with named op (mutation)')
+     */
+    public function testGetsAnOperationFromADcoumentWithNamedOpMutation() : void
+    {
+        $doc = Parser::parse('mutation Test { field }');
+        self::assertEquals(AST::getOperationAST($doc), $doc->definitions->offsetGet(0));
+    }
+
+    /**
+     * @see it('Gets an operation from a document with named op (subscription)')
+     */
+    public function testGetsAnOperationFromADcoumentWithNamedOpSubscription() : void
+    {
+        $doc = Parser::parse('subscription Test { field }');
+        self::assertEquals(AST::getOperationAST($doc), $doc->definitions->offsetGet(0));
+    }
+
+    /**
+     * @see it('Does not get missing operation')
+     */
+    public function testDoesNotGetMissingOperation() : void
+    {
+        $doc = Parser::parse('type Foo { field: String }');
+        self::assertEquals(AST::getOperationAST($doc), null);
+    }
+
+    /**
+     * @see it('Does not get ambiguous unnamed operation')
+     */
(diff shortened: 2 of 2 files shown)


src/Utils/AST.php
```diff
@@ -608,4 +610,36 @@ public static function getOperation(DocumentNode $document, $operationName = nul
 
         return false;
     }
+
+    /**
+     * Returns the operation given a document and operation name.
```
spawnia, line 615:
```suggestion
     * Returns the operation within a document by name.
```

src/Utils/AST.php
```diff
@@ -608,4 +610,36 @@ public static function getOperation(DocumentNode $document, $operationName = nul
...
+     *
+     * If a name is not provided, an operation is only returned if only one is
+     * provided in the document.
```
spawnia, line 618:
```suggestion
     * If a name is not provided, an operation is only returned if the document has exactly one.
```

src/Utils/AST.php
```diff
@@ -608,4 +610,36 @@ public static function getOperation(DocumentNode $document, $operationName = nul
...
+        $operation = null;
+        foreach ($document->definitions->getIterator() as $node) {
+            if (! $node instanceof OperationDefinitionNode) {
+                continue;
+            }
+
+            if ($operationName === null) {
+                // If no operation name was provided, only return an Operation if there
+                // is one defined in the document. Upon encountering the second, return
+                // null.
```
spawnia, line 633:
```suggestion
                // We found a second operation, so we bail instead of returning an ambiguous result.
```

tests/Utils/AstGetOperationAstTest.php
```diff
@@ -0,0 +1,105 @@
...
+{
+    // Describe: getOperationAST
+
+    /**
+     * @see it('Gets an operation from a simple document')
+     */
+    public function testGetsAnOperationFromASimpleDocument() : void
+    {
+        $doc = Parser::parse('{ field }');
+        self::assertEquals(AST::getOperationAST($doc), $doc->definitions->offsetGet(0));
```
spawnia, line 21:
```suggestion
        self::assertEquals(
            AST::getOperationAST($doc),
            $doc->definitions->offsetGet(0)
        );
```

src/Utils/AST.php
```diff
@@ -608,4 +610,36 @@ public static function getOperation(DocumentNode $document, $operationName = nul
...
+                if ($operation !== null) {
+                    return null;
```
spawnia, line 635: I would consider throwing an Exception here.

spawnia (commented):
I don't love the design of the reference implementation, consider the following:

<table>
<tr>
	<td>Amount of operations within the DocumentAST
	<td>Result of calling getOperationAST($doc)
<tr>
	<td>0
	<td>null
<tr>
	<td>1
	<td>Operation
<tr>
	<td>>1
	<td>null
</table>

src/Utils/AST.php
Kingdutch: Distinguishing this case with an exception doesn't make much sense I think since it mostly increases the calling site's burden without providing a better error then the other times that null gets returned.

This is the calling logic that I use in a server-like implementation:
```
            // If no operation was found we can't continue.
            $operation = AST::getOperationAST($document, $msg->getOperationName());
            if ($operation === NULL) {
              throw new OperationNotFoundException('Could not identify operation.');
            }
```

The [current usage in this library](https://github.com/webonyx/graphql-php/blob/521efa8f089e865404afc339c56d1e1077f92e10/src/Server/Helper.php#L292) is as follows and when switching over to the new (undeprecated) method would also be unnecessarily complicated by moving to an exception instead of sticking to the reference implementation.

```
           $operationType = AST::getOperation($doc, $op->operation);

            if ($operationType === false) {
                throw new RequestError('Failed to determine operation type');
            }

            if ($operationType !== 'query' && $op->isReadOnly()) {
                throw new RequestError('GET supports only query operation');
            }
            ```

Regarding your comment
> This special case is a somewhat hidden implementation detail, I believe it does not lead to robust usage.
The [formal specification](https://spec.graphql.org/June2018/#sec-Lone-Anonymous-Operation) for lone anonymous operations actually doesn't allow multiple anonymous operations, and covers this case. 

So it's not just an implementation detail but spec compliant. Although the reporting to a client is left for other functions (which is mirrored in the examples above).

Given multiple anonymous operations (and no named operations) then the set of `operations` would be bigger then `1` thus `anonymous` must be empty, which means this is an invalid document.

Comment (spawnia) [reply src/Utils/AST.php:632]: There is a substantial amount of duplication for functionality that is basically:

```
(DocumentNode $document, ?string $operationName): OperationDefinitionNode | null | Error
```

https://github.com/webonyx/graphql-php/blob/521efa8f089e865404afc339c56d1e1077f92e10/src/Experimental/Executor/Collector.php#L72-L102

https://github.com/webonyx/graphql-php/blob/521efa8f089e865404afc339c56d1e1077f92e10/src/Executor/ReferenceExecutor.php#L140-L167

...and now this implementation.

Comment (Kingdutch) [reply src/Utils/AST.php:632]: Ah I see 🤔 The js implementation contains [the same duplication](https://github.com/graphql/graphql-js/blob/d6e760cf7060217bd25fa934bd104ff2400aad96/src/execution/execute.js#L312). It's only made more noticeable in graphql-php since it contains two Executor implementations with the logic.

A few weeks ago I would've been all for cleaning up the PHP code sites, but now that I'm at my second contribution I'm actually leaning towards sticking to the duplication that's present in the graphql-js library (and if need be, updating there before here). Since it makes development a lot easier since all the JS examples out there suddenly map nicely to PHP, even for more complex things.

**Comment by Kingdutch** [comment]
Has the feedback provided been sufficiently addressed or is more work needed here? :) Would love to get this merged.

**Review by spawnia — APPROVED** [review]
I would like a more principled approach, but this implementation seems to have proven its worth in `graphql-js` and is better than nothing.

**Comment by Kingdutch** [comment]
> I would like a more principled approach, but this implementation seems to have proven its worth in graphql-js and is better than nothing.

That's a raving review if I've ever seen one 😂 I'll take it! 👍  

Perhaps we can take the feedback to graphql-js and create an issue for change there, then it should find its way back here eventually :D

File: tests/Utils/AstGetOperationAstTest.php [hunk 6]
```diff
@@ -0,0 +1,108 @@
...
+
+declare(strict_types=1);
+
+namespace GraphQL\Tests\Utils;
+
+use GraphQL\Language\Parser;
+use GraphQL\Utils\AST;
+use PHPUnit\Framework\TestCase;
+
+class AstGetOperationAstTest extends TestCase
```
simPod, line 11: I would add a `final` so it is not offered for extension.

tests/Utils/AstGetOperationAstTest.php
```diff
@@ -0,0 +1,108 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Tests\Utils;
+
+use GraphQL\Language\Parser;
+use GraphQL\Utils\AST;
+use PHPUnit\Framework\TestCase;
+
```
simPod, line 10: Would it make sense to add `@covers` annotation to allow this to influence `AST`'s coverage only?

simPod (approved):

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v14.8.0

PR #756, opened 2020-12-16
fix query plan for wrapped types

crissi:
Was getting error: "Call to undefined method GraphQL\Type\Definition\ListOfType::getField()" when using QueryPlan::lookAhead()

### DIFF
--- a/tests/Type/QueryPlanTest.php
+++ b/tests/Type/QueryPlanTest.php
@@ -298,6 +298,60 @@ final class QueryPlanTest extends TestCase
         self::assertFalse($queryPlan->hasType('Test'));
     }
 
+    public function testQueryPlanForWrappedTypes() : void
+    {
+        $article = new ObjectType([
+            'name'   => 'Article',
+            'fields' => [
+                'id' => ['type' => Type::string()],
+            ],
+        ]);
+
+        $doc               = '
+      query Test {
+        articles {
+            id
+        }
+      }
+';
+        $expectedQueryPlan = [
+            'id' => [
+                'type' => Type::string(),
+                'fields' => [],
+                'args' => [],
+            ],
+        ];
+
+        /** @var QueryPlan|null $queryPlan */
+        $queryPlan = null;
+
+        $blogQuery = new ObjectType([
+            'name'   => 'Query',
+            'fields' => [
+                'articles' => [
+                    'type'    => Type::nonNull(Type::listOf($article)),
+                    'resolve' => static function (
+                        $value,
+                        $args,
+                        $context,
+                        ResolveInfo $info
+                    ) use (
+                        &$queryPlan
+                    ) : array {
+                        $queryPlan = $info->lookAhead();
+
+                        return [];
+                    },
+                ],
+            ],
+        ]);
+
+        $schema = new Schema(['query' => $blogQuery]);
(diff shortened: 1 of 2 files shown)


tests/Type/QueryPlanTest.php
```diff
@@ -298,6 +298,78 @@ public function testQueryPlan() : void
...
+        $expectedQueryPlan = [
+            'id' => [
+                'type' => Type::string(),
+                'args' => [],
+                'fields' => [],
+            ],
+        ];
+
+        $hasCalled = false;
+        /** @var QueryPlan $queryPlan */
```
spawnia, line 326:
```suggestion
        /** @var QueryPlan|null $queryPlan */
```

tests/Type/QueryPlanTest.php
```diff
@@ -298,6 +298,78 @@ public function testQueryPlan() : void
...
+                        ];
+                    },
+                ],
+            ],
+        ]);
+
+        $schema = new Schema(['query' => $blogQuery]);
+        $result = GraphQL::executeQuery($schema, $doc)->toArray();
+
+        self::assertTrue($hasCalled);
```
spawnia, line 358: I don't think this is necessary.

tests/Type/QueryPlanTest.php
```diff
@@ -298,6 +298,78 @@ public function testQueryPlan() : void
...
+        self::assertEquals(
```
spawnia, line 359: That is also superfluous. How about you just return `null` from the resolver and don't even check the result? That makes the test more focused.

spawnia (changes requested):

tests/Type/QueryPlanTest.php
```diff
@@ -298,6 +298,60 @@ public function testQueryPlan() : void
...
+                        $queryPlan = $info->lookAhead();
+
+                        return [];
+                    },
+                ],
+            ],
+        ]);
+
+        $schema = new Schema(['query' => $blogQuery]);
+        $result = GraphQL::executeQuery($schema, $doc)->toArray();
```
spawnia, line 350:
```suggestion
        GraphQL::executeQuery($schema, $doc);
```

spawnia (approved):

tests/Type/QueryPlanTest.php
```diff
@@ -298,6 +298,60 @@ public function testQueryPlan() : void
...
+                        return [];
+                    },
+                ],
+            ],
+        ]);
+
+        $schema = new Schema(['query' => $blogQuery]);
+        GraphQL::executeQuery($schema, $doc);
+
+        self::assertEquals($expectedQueryPlan, $queryPlan->queryPlan());
```
simPod, line 352: well mb we should use `assertSame` or another set of assertions. `assertEquals` might incorrectly compare objects which might lead to bugs when test fixtures are modified in the future.

simPod (approved):

tests/Type/QueryPlanTest.php
crissi: okay will change

spawnia:
The implementation of is fine, but i would like @vladar to confirm that this is indeed how the query plan should behave for wrapped types.

vladar (approved):
Agree, this is a valid bugfix! Thanks! 👍

Issue #759, opened 2020-12-28
Executor fetches data twice: offsetExists on ArrayAccess


brunobg:
I'm finding a few duplicated queries on my logs and debugging. I'm using lighthouse-php but the stack trace shows they seem related to graphql-php, so I'm posting them here. I think they have different causes so I'll open multiple issues.

First chapter: I'm using Laravel with an accessor that hits the DB. The query is called twice. 

More or less this:

```php
class Stuff extends Model {
    public function getFancyAttribute() {
       return DB::select('SELECT sometable.id FROM sometable');  // some fancy query here
    }
}
```

I fetch this on graphql, which is like this:

```graphql
type Stuff {
   id: ID!
   fancy: [Fancytype!]
}

extend type Query {
   stuff(id: String @id): Stuff @find
```

So when I query I see two selects, caused by calling `Model::offsetExists` and `Model::offsetGet`. Here are the relevant stack traces for both calls:

```[2020-12-28 15:59:33] local.ERROR: array (
  0 => 
  array (
    'file' => '/server/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php',
    'line' => 473,
    'function' => 'getFancyAttribute',
    'class' => 'App\\Models\\Stuff',
    'type' => '->',
  ),
  1 => 
  array (
    'file' => '/server/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php',
    'line' => 1444,
    'function' => 'mutateAttribute',
    'class' => 'Illuminate\\Database\\Eloquent\\Model',
    'type' => '->',
  ),
  2 => 
  array (
    'file' => '/server/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php',
    'line' => 386,
    'function' => 'transformModelValue',
    'class' => 'Illuminate\\Database\\Eloquent\\Model',
    'type' => '->',
  ),
  3 => 
  array (
    'file' => '/server/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php',
    'line' => 365,
    'function' => 'getAttributeValue',
    'class' => 'Illuminate\\Database\\Eloquent\\Model',
    'type' => '->',
  ),
  4 => 
  array (
    'file' => '/server/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php',
    'line' => 1660,
    'function' => 'getAttribute',
[...]

PR #760, opened 2020-12-29
Fix double access to ArrayAccess methods

brunobg:
This is a proposal to fix https://github.com/webonyx/graphql-php/issues/759

### DIFF
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -179,6 +179,10 @@ class Executor
                 $property = $objectValue[$fieldName];
             }
         } elseif ($objectValue instanceof ArrayAccess) {
+            // handles #759: ArrayAccess::offsetExists() can call offsetGet() internally,
+            // which results in calling offsetGet() twice.
+            // This avoids the double call and handles Exceptions to have the same 
+            // behavior as isset() in the pure array case.
             try {
                 $property = $objectValue[$fieldName];
             } catch (Throwable $e) {
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -6,6 +6,7 @@ namespace GraphQL\Executor;
 
 use ArrayAccess;
 use Closure;
+use Throwable;
 use GraphQL\Executor\Promise\Adapter\SyncPromiseAdapter;
 use GraphQL\Executor\Promise\Promise;
 use GraphQL\Executor\Promise\PromiseAdapter;
@@ -180,7 +181,7 @@ class Executor
         } elseif ($objectValue instanceof ArrayAccess) {
             try {
                 $property = $objectValue[$fieldName];
-            } catch (\Throwable $e) {
+            } catch (Throwable $e) {
                 // pass
             }
         } elseif (is_object($objectValue)) {
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -178,7 +178,11 @@ class Executor
                 $property = $objectValue[$fieldName];
             }
         } elseif ($objectValue instanceof ArrayAccess) {
-            $property = $objectValue[$fieldName];
+            try {
+                $property = $objectValue[$fieldName];
+            } catch (\Exception $e) {
+                // pass
+            }
         } elseif (is_object($objectValue)) {
             if (isset($objectValue->{$fieldName})) {
                 $property = $objectValue->{$fieldName};
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -180,7 +180,7 @@ class Executor
         } elseif ($objectValue instanceof ArrayAccess) {
             try {
                 $property = $objectValue[$fieldName];
-            } catch (\Exception $e) {
+            } catch (\Throwable $e) {
                 // pass
             }
         } elseif (is_object($objectValue)) {
--- a/src/Executor/Executor.php
+++ b/src/Executor/Executor.php
@@ -173,15 +173,15 @@ class Executor
         $fieldName = $info->fieldName;
         $property  = null;
 
-        if (is_array($source)) {
-            if (isset($source[$fieldName])) {
-                $property = $source[$fieldName];
+        if (is_array($objectValue)) {
+            if (isset($objectValue[$fieldName])) {
+                $property = $objectValue[$fieldName];
             }
-        } elseif ($source instanceof ArrayAccess) {
-            $property = $source[$fieldName];
-        } elseif (is_object($source)) {
-            if (isset($source->{$fieldName})) {
-                $property = $source->{$fieldName};
+        } elseif ($objectValue instanceof ArrayAccess) {
+            $property = $objectValue[$fieldName];
+        } elseif (is_object($objectValue)) {
+            if (isset($objectValue->{$fieldName})) {
+                $property = $objectValue->{$fieldName};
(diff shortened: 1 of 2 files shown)


brunobg:
Hm, breaks a lot of tests. Let me see if it's `property_exists` that broke things..

src/Executor/Executor.php
```diff
@@ -173,10 +173,12 @@ public static function defaultFieldResolver($objectValue, $args, $contextValue,
         $fieldName = $info->fieldName;
         $property  = null;
 
-        if (is_array($objectValue) || $objectValue instanceof ArrayAccess) {
+        if (is_array($objectValue)) {
             if (isset($objectValue[$fieldName])) {
                 $property = $objectValue[$fieldName];
             }
+        } elseif ($objectValue instanceof ArrayAccess) {
```
spawnia, line 180: An explicit test case for this would be great.

brunobg: Sure, but I need some pointers. There's no unit test for `defaultFieldResolver()`. Any pointers to test code that I could use as a base?

spawnia: > Sure, but I need some pointers.

Here you go:

```c
int *foo;
char *bar;
```

*Sorry, couldn't resist.*

I am assuming there is a test class that covers this entire class.

spawnia: I just dug around https://github.com/webonyx/graphql-php/blob/master/tests/Executor/ExecutorTest.php a bit. Given this library is mostly a port of the JS reference implementation, it is possible there are no explicit tests for the default resolver, given that JS only has objects and regular property access.

brunobg:
```
Segmentation fault
```

brunobg: Perhaps `testExecutesArbitraryCode` could have yet another element added to its already huge array, with a ArrayAccessor instance instead of a pure array. Something like `deep`, a `deepAccessor`. Since there is no coverage there were no previous tests checking this if branch. If you have a better idea give me some extra pointers. But make them a pointer to a reference to a pointer this time.
I could also write a basic unit test for it, but it'd be completely useless.

spawnia: A dedicated test method would be better, since the existing method should mirror the JS implementation. Something like `testDefaultResolverHandlesArrayAccess`?

mfn:
I'm supporting this change 👍 

I discovered this some time ago and AFAIK suggested to lighthouse to adapt this but it wasn't picked up (and I also don't use lighthouse, I just noticed this in different Laravel  / GraphQL adapter).

I tested this on a private project with lots of GraphQL tests, no issues found.

Are we concerned about a BC break here? Maybe we should considering target the next major version? #JustBeingCautious here

src/Executor/Executor.php
spawnia: I added a comprehensive test case to reify the current behaviour: https://github.com/webonyx/graphql-php/pull/761

spawnia:
@mfn great to see you around here.

This change indeed causes breakage when `ArrayAccess::offsetGet()` is implemented to throw when unset offsets are accessed. This behaviour is in line with how plain arrays work, I suspect many implementations in the wild work like this.

brunobg:
@spawnia see my reply on the original issue and the new patch. 

@mfn I think that should be as safe as the currently released code and doesn't even need a major version, could be part of a patch release. There are no changes to array and object.

There's a worst case scenario, with a trashy implementation of ArrayAccess. Let's say one that just wraps over a normal array, calling `isset()` on `offsetExists()` and accessing [] with `offsetGet()` with no checks, which generates a notice. But we shouldn't avoid this change because of that IMHO.

brunobg:
Hey @spawnia anything else I can do about this? I'd love to get this accepted.

spawnia:
The implementation looks fine, I would like to avoid feedback. @vladar usually comes by every few weeks and merges a bunch of PRs.

simPod (approved):

src/Executor/Executor.php
```diff
@@ -173,10 +174,16 @@ public static function defaultFieldResolver($objectValue, $args, $contextValue,
...
```
spawnia, line 181: I think we should add a few lines of comment here to explain the rationale behind why this is treated differently from plain arrays.

brunobg: Extra docs added.

spawnia: Nicely done, thank you

brunobg:
Considering other reports such as  https://github.com/nuwave/lighthouse/issues/1671, who knows how many people are having this behavior without noticing, please consider this PR and a new release ASAP @vladar and @spawnia . Thanks!

spawnia (approved):
This change does smooth over the weird semantics of `ArrayAccess` in a way that is beneficial to most if not all users.

vladar:
This PR is trying to fix a problem in the wrong place. `ArrayAccess` is designed after arrays and it is quite expected to call `isset` (`offsetExists`) before trying to access an element. Take this simplest implementation of a class implementing `ArrayAccess`:

```php
<?php
class Foo implements ArrayAccess
{
    private $foo = [];

    public function __construct(array $foo)
    {
        $this->foo = $foo;
    }

    public function offsetExists($offset)
    {
        return isset($this->foo[$offset]);
    }

    public function offsetGet($offset)
    {
        return $this->foo[$offset];
    }

    public function offsetSet($offset, $value)
    {
        $this->foo[$offset] = $value;
    }

    public function offsetUnset($offset)
    {
        unset($this->foo[$offset]);
    }
}
```

Now with the proposed change if you try to access a missing key like this:

```php
$d = new Foo([ 'foo' => 'bar' ]);
$property = null;

try {
    $property = $d['bar'];
} catch (Throwable $e) {
    // pass
}
```

you will get PHP notice:

```
PHP Notice:  Undefined index: bar in Foo.php on line 18
```

Obviously, you can have an additional `isset` check in your `offsetGet` method but you are not required to! It is the responsibility of the `ArrayAccess` implementation to be smart enough not to do expensive operations multiple times.

So this is a breaking change and also an invalid behavior.

The generic implementation for `defaultResolver` should stay logically correct and consume interfaces like `ArrayAccess` in a way they were designed. You can always override the default resolver for your specific situation by passing fieldResolver argument in [GraphQL::executeQuery](https://webonyx.github.io/graphql-php/executing-queries/) or passing `fieldResolver` in [StandardServer options](https://webonyx.github.io/graphql-php/executing-queries/#server-configuration-options).

Or alternatively you can set it globally via `GraphQL\Executor\Executor::setDefaultFieldResolver($myDefaultResolver)`

spawnia (changes requested):
The semantics of `ArrayAccess` are not that weird after all, having checked how they work in https://3v4l.org/OLqvr.

I think this change can be valid if you know you are dealing with inefficient implementations, such as Laravel's. As @vladar noted, it is easily possible to overwrite the default resolver in such a case.

brunobg:
Closing this though I disagree as explained in the issue. Also, note that a detailed analysis of what happens in Laravel https://github.com/laravel/framework/issues/36026 shows that real implementations can throw in offsetExists(). So if you're worried that a trivial implementation can throw a notice in my PR, you should be worried that it can also throw exceptions.

Issue #762, opened 2020-12-31
ListOfType could contain NonNull


brunobg:
**Describe the bug**

Close this if I'm wrong, but `GraphQL\Type\Definition\ListOfType` lists a number of types on its phpdocs but is missing `NonNull` on `$ofType` and `getWrapperType`:

```
/** @var ObjectType|InterfaceType|UnionType|ScalarType|InputObjectType|EnumType */
```

You could have something like `[MyType!]` which makes a `ListOfType[NonNull[ObjectType]]`. This is not a bug per se, but stan is complaning about a if that should never happen in my code when it actually does. Am I missing something? If this issue is correct I can send the trivial PR to fix it.

spawnia:
The newest version already has a different annotation there.

PR #763, opened 2021-01-04
Simplify type hints in wrapping types

spawnia:
Resolves https://github.com/webonyx/graphql-php/issues/762

### DIFF
--- a/src/Type/Definition/ListOfType.php
+++ b/src/Type/Definition/ListOfType.php
@@ -30,9 +30,6 @@ class ListOfType extends Type implements WrappingType, OutputType, NullableType,
         return Schema::resolveType($this->ofType);
     }
 
-    /**
-     * @return ObjectType|InterfaceType|UnionType|ScalarType|InputObjectType|EnumType|(Type&WrappingType)
-     */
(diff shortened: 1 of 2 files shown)


src/Type/Definition/ListOfType.php
```diff
@@ -30,9 +30,6 @@ public function getOfType()
         return Schema::resolveType($this->ofType);
     }
 
-    /**
-     * @return ObjectType|InterfaceType|UnionType|ScalarType|InputObjectType|EnumType|(Type&WrappingType)
```
simPod, line 34: im not sure about this change 🤷🏾

simPod (approved):

src/Type/Definition/ListOfType.php
spawnia: It is really not necessary, since any `Type` can be wrapped within a list. I guess we could clarify the return type of `NonNull::getWrappedType()`, which can not hold another `NonNull`.

simPod: Ah right 👍🏾

PR #766, opened 2021-01-13
Do not add errors key when empty

TamasSzigeti:
If the error handler filters out all the errors so that none remains, we should not add the errors key to the result.

### DIFF
--- a/src/Executor/ExecutionResult.php
+++ b/src/Executor/ExecutionResult.php
@@ -148,7 +148,7 @@ class ExecutionResult implements JsonSerializable
             );
 
             // While we know that there were errors initially, they might have been discarded
-            if ($handledErrors) {
+            if ($handledErrors !== []) {
                 $result['errors'] = $handledErrors;
             }
         }
--- a/src/Executor/ExecutionResult.php
+++ b/src/Executor/ExecutionResult.php
@@ -142,10 +142,14 @@ class ExecutionResult implements JsonSerializable
                 return array_map($formatter, $errors);
             };
 
-            $result['errors'] = $errorsHandler(
+            $formattedErrors = $errorsHandler(
                 $this->errors,
                 FormattedError::prepareFormatter($this->errorFormatter, $debug)
             );
+
+            if ($formattedErrors) {
+                $result['errors'] = $formattedErrors;
+            }
         }
 
         if ($this->data !== null) {
--- a/src/Executor/ExecutionResult.php
+++ b/src/Executor/ExecutionResult.php
@@ -142,13 +142,14 @@ class ExecutionResult implements JsonSerializable
                 return array_map($formatter, $errors);
             };
 
-            $formattedErrors = $errorsHandler(
+            $handledErrors = $errorsHandler(
                 $this->errors,
                 FormattedError::prepareFormatter($this->errorFormatter, $debug)
             );
 
-            if ($formattedErrors) {
-                $result['errors'] = $formattedErrors;
+            // While we know that there were errors initially, they might have been discarded
+            if ($handledErrors) {
+                $result['errors'] = $handledErrors;
             }
         }
 
--- a/tests/Executor/ExecutionResultTest.php
+++ b/tests/Executor/ExecutionResultTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace GraphQL\Tests\Executor;
 
+use GraphQL\Error\Error;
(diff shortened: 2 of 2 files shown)


src/Executor/ExecutionResult.php
```diff
@@ -142,10 +142,14 @@ public function toArray(int $debug = DebugFlag::NONE) : array
                 return array_map($formatter, $errors);
             };
 
-            $result['errors'] = $errorsHandler(
+            $formattedErrors = $errorsHandler(
```
spawnia, line 145: Not just formatted, also filtered etc.

```suggestion
            $handledErrors = $errorsHandler(
```

spawnia (commented):
I just added some minutiae for clarity, the change makes sense to me.

TamasSzigeti:
> I just added some minutiae for clarity, the change makes sense to me.

Thank you, done both.

src/Executor/ExecutionResult.php
```diff
@@ -142,10 +142,15 @@ public function toArray(int $debug = DebugFlag::NONE) : array
...
             };
 
-            $result['errors'] = $errorsHandler(
+            $handledErrors = $errorsHandler(
                 $this->errors,
                 FormattedError::prepareFormatter($this->errorFormatter, $debug)
             );
+
+            // While we know that there were errors initially, they might have been discarded
+            if ($handledErrors) {
```
simPod, line 151: I'd evaluate boolean expression here. Guess we want to check for non-emptiness?

```suggestion
            if ($handledErrors !== []) {
```

tests/Executor/ExecutionResultTest.php
```diff
@@ -26,4 +27,16 @@ public function testToArrayExtensions() : void
...
+    public function testNoEmptyErrors() : void
+    {
+        $executionResult = new ExecutionResult(null, [new Error()]);
+        $executionResult->setErrorsHandler(
+            static function () : array {
+                return [];
+            }
+        );
+
+        self::assertEquals([], $executionResult->toArray());
```
simPod, line 40: assertSame can be used

src/Executor/ExecutionResult.php
TamasSzigeti: It's the same thing, empty array is falsy

tests/Executor/ExecutionResultTest.php
TamasSzigeti: Done, replaced the other cases too.

src/Executor/ExecutionResult.php
simPod: Yup but it's not type safe. When type changes it does not get caught. Triggers error with phpstan with strict types extension on.

https://phpstan.org/r/d5fe9f4e-cdb7-49ce-88ed-8f17f85cb691

TamasSzigeti: I feel it's unnecessarily verbose and has no real benefit, but if the plan is to turn on phpstan strict, I don't mind

simPod: At least in my opinion it's better to rely on types so we need less type-related testing in the long run 🤷🏾‍♀️ .

simPod (approved):
Just checked and complies with spec 👍🏾

spawnia (approved):
Nice

TamasSzigeti:
Can we merge and tag this?

src/Executor/ExecutionResult.php
```diff
@@ -142,10 +142,15 @@ public function toArray(int $debug = DebugFlag::NONE) : array
...
-            $result['errors'] = $errorsHandler(
+            $handledErrors = $errorsHandler(
                 $this->errors,
                 FormattedError::prepareFormatter($this->errorFormatter, $debug)
             );
+
+            // While we know that there were errors initially, they might have been discarded
+            if ($handledErrors !== []) {
+                $result['errors'] = $handledErrors;
+            }
```
Kingdutch, line 153: I realise that this may be a tad late (since it's merged), but the comment suggests this array should be named `$unhandledErrors` for clarity? Since discarded errors are probably handled by an error handler and recovered from. The remaining errors are unrecoverable/unhandled, those are returned to the client.

spawnia: Handling the errors could also mean to translate the messages, format them nicely, whatever!

Kingdutch: Ah I see :D Yeah then I suppose it makes sense 🙂

PR #767, opened 2021-01-20
Allow pushing nodes to NodeList via []=

spawnia:
This can come up when manipulating the AST programmatically.
When pushing to a `NodeList` using `[]=`, the first argument
to `offsetExists` is `null`.

Thus, the element is set into the internal array with `(string) null === ""`.
Now, additional pushes replace the node at key `""` instead of appending as expected.

### DIFF
--- a/src/Language/AST/NodeList.php
+++ b/src/Language/AST/NodeList.php
@@ -95,6 +95,7 @@ class NodeList implements ArrayAccess, IteratorAggregate, Countable
             $value = AST::fromArray($value);
         }
 
+        // Happens when a Node is pushed via []=
         if ($offset === null) {
             $this->nodes[] = $value;
 
--- a/src/Language/AST/NodeList.php
+++ b/src/Language/AST/NodeList.php
@@ -91,17 +91,14 @@ class NodeList implements ArrayAccess, IteratorAggregate, Countable
     public function offsetSet($offset, $value) : void
     {
         if (is_array($value)) {
-            if (isset($value['kind'])) {
-                /** @phpstan-var T $node */
-                $node                 = AST::fromArray($value);
-                $this->nodes[$offset] = $node;
+            /** @phpstan-var T $value */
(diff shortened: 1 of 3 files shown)


src/Language/AST/NodeList.php
```diff
@@ -91,17 +91,15 @@ public function offsetGet($offset)// : Node
     public function offsetSet($offset, $value) : void
     {
         if (is_array($value)) {
-            if (isset($value['kind'])) {
-                /** @phpstan-var T $node */
-                $node                 = AST::fromArray($value);
-                $this->nodes[$offset] = $node;
+            /** @phpstan-var T $value */
+            $value = AST::fromArray($value);
```
simPod, line 95: wouldn't it be better to keep var name as $node so we don't change type of `$value argument` to `T`?

tests/Language/NodeListTest.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace GraphQL\Tests\Language;
+
+use GraphQL\Error\InvariantViolation;
+use GraphQL\Language\AST\NameNode;
+use GraphQL\Language\AST\NodeList;
+use PHPUnit\Framework\TestCase;
+use function get_class;
+
+class NodeListTest extends TestCase
```
simPod, line 13:
```suggestion
final class NodeListTest extends TestCase
```

simPod (approved):

src/Language/AST/NodeList.php
spawnia: Well, we *do* change its type. Since it was `array|T` before and headed into this branch because it `is_array()`, we now successfully narrowed it.

simPod: aight 👍🏾

spawnia:
@vladar can i merge this and cut a bugfix release? Just ran into this bug in Lighthouse.

vladar (approved):
Nice! 👍

Issue #769, opened 2021-01-25
phpstan, \GraphQL\Language\AST\FieldNode::$arguments and `??`


mfn:
Upgraded to phpstan 0.12.69 with level 6 today and I got this:
```
  112    Property GraphQL\Language\AST\FieldNode::$arguments
         (GraphQL\Language\AST\NodeList<GraphQL\Language\AST\ArgumentNode>) on
         left side of ?? is not nullable.
```
I'm using `\GraphQL\Language\AST\FieldNode::$arguments` in a code like this:
```php
                foreach ($selectionNode->arguments ?? [] as $argumentNode) {
                    $fields[$name]['args'][$argumentNode->name->value] = $this->getValue($argumentNode);
                }
```

The fix is easy but when I got curious and checked the project, I found out it also uses the pattern `?? []` a few times in combination with this field:
![image](https://user-images.githubusercontent.com/87493/105658474-d11ff480-5ec6-11eb-94cb-7a3fd2069727.png)

I found that field also on other node types:
![image](https://user-images.githubusercontent.com/87493/105658583-088ea100-5ec7-11eb-995a-580fda6ba7b2.png)

None of them specifies `null`ability.

However I also see that this project uses level 3 and I can't say when/if this analysis kicks.

I'm trying to figure out if removing the `?? []` is truly safe or not or maybe the library also could use adjustments here?

simPod:
IIRC null checks kick in lvl 4.

Though there's some magic in Node constructor so it would be problematic to check statically with current code state.

```php
    /**
     * @param (NameNode|NodeList|SelectionSetNode|Location|string|int|bool|float|null)[] $vars
     */
    public function __construct(array $vars)
    {
        if (count($vars) === 0) {
            return;
        }

        Utils::assign($this, $vars);
    }
```

spawnia:
The specific typehints you are referring to are indeed never `null`. Some related properties were, so I went ahead and created https://github.com/webonyx/graphql-php/pull/770 to ensure the typehints for those are accurate as well.

PR #770, opened 2021-01-25
Ensure properties annotated to hold NodeList are not null

spawnia:
- Resolves https://github.com/webonyx/graphql-php/issues/769
- Fixes 2 bugs related to schema validation of directives.


vladar (commented):
This could lead to subtle breaking changes if someone is doing things like `if(isset($type->extensionASTNodes)) { /**/ }`

spawnia:
I would argue this is a bugfix. The annotations said the property is an array, but it was falsely set to `null`.

spawnia:
Regarding your example, most likely the code inside the `isset()` will loop after the extension nodes.

spawnia:
Alright, since we are now moving towards a new major version the slight potential breakage of this PR should no longer be an issue.

src/Type/Definition/Directive.php
```diff
@@ -31,7 +31,7 @@ class Directive
     /** @var string|null */
     public $description;
 
-    /** @var FieldArgument[] */
+    /** @var array<int, FieldArgument> */
```
simPod, line 34: Can we use `list<>`?

src/Type/Definition/EnumType.php
```diff
@@ -43,7 +43,7 @@ class EnumType extends Type implements InputType, OutputType, LeafType, Nullable
     /** @var ArrayObject<string, EnumValueDefinition> */
     private $nameLookup;
 
-    /** @var EnumTypeExtensionNode[] */
+    /** @var array<int, EnumTypeExtensionNode> */
```
simPod, line 46: Can we use `list<>`?

src/Type/Definition/FieldArgument.php
```diff
@@ -56,9 +56,9 @@ public function __construct(array $def)
     }
 
     /**
-     * @param mixed[] $config
+     * @param array<mixed> $config
      *
-     * @return FieldArgument[]
+     * @return array<int, FieldArgument>
```
simPod, line 61: Can we use `list<>`?

src/Type/Definition/FieldDefinition.php
```diff
@@ -25,7 +25,7 @@ class FieldDefinition
     /** @var string */
     public $name;
 
-    /** @var FieldArgument[] */
+    /** @var array<int, FieldArgument> */
```
simPod, line 28: Can we use `list<>`?

src/Type/Definition/InputObjectType.php
```diff
@@ -26,7 +26,7 @@ class InputObjectType extends Type implements InputType, NullableType, NamedType
      */
     private $fields;
 
-    /** @var InputObjectTypeExtensionNode[] */
+    /** @var array<int, InputObjectTypeExtensionNode> */
```
simPod, line 29: Can we use `list<>`?

src/Type/Definition/ObjectType.php
```diff
@@ -60,7 +60,7 @@ class ObjectType extends Type implements OutputType, CompositeType, NullableType
     /** @var ObjectTypeDefinitionNode|null */
     public $astNode;
 
-    /** @var ObjectTypeExtensionNode[] */
+    /** @var array<int, ObjectTypeExtensionNode> */
```
simPod, line 63: Can we use `list<>`?

src/Type/Definition/UnionType.php
```diff
@@ -33,7 +33,7 @@ class UnionType extends Type implements AbstractType, OutputType, CompositeType,
      */
     private $possibleTypeNames;
 
-    /** @var UnionTypeExtensionNode[] */
+    /** @var array<int, UnionTypeExtensionNode> */
```
simPod, line 36: Can we use `list<>`?

src/Type/Schema.php
```diff
@@ -84,7 +84,7 @@ class Schema
     /** @var Error[] */
     private $validationErrors;
 
-    /** @var SchemaTypeExtensionNode[] */
+    /** @var array<int, SchemaTypeExtensionNode> */
```
simPod, line 87: Can we use `list<>`?

src/Type/SchemaConfig.php
```diff
@@ -52,7 +52,7 @@ class SchemaConfig
     /** @var bool */
     public $assumeValid = false;
 
-    /** @var SchemaTypeExtensionNode[] */
+    /** @var array<int, SchemaTypeExtensionNode> */
```
simPod, line 55: Can we use `list<>`?

src/Type/SchemaConfig.php
```diff
@@ -37,7 +37,7 @@ class SchemaConfig
     /** @var ObjectType|null */
     public $subscription;
 
-    /** @var Type[]|callable */
+    /** @var array<Type>|(callable(): array<Type>) */
```
simPod, line 40: what is the array key type here?

src/Type/SchemaConfig.php
```diff
@@ -216,13 +216,13 @@ public function getTypes()
     }
 
     /**
-     * @param Type[]|callable $types
+     * @param array<Type>|(callable(): array<Type>) $types
```
simPod, line 219: Missing array key type perhaps

src/Type/SchemaValidationContext.php
```diff
@@ -521,9 +521,9 @@ private function validateFields($type)
     /**
      * @param Schema|ObjectType|InterfaceType|UnionType|EnumType|InputObjectType|Directive $obj
      *
-     * @return ObjectTypeDefinitionNode[]|ObjectTypeExtensionNode[]|InterfaceTypeDefinitionNode[]|InterfaceTypeExtensionNode[]
+     * @return array<ObjectTypeDefinitionNode>|array<ObjectTypeExtensionNode>|array<InterfaceTypeDefinitionNode>|array<InterfaceTypeExtensionNode>
```
simPod, line 524: Isn't this 

```suggestion
     * @return list<ObjectTypeDefinitionNode|ObjectTypeExtensionNode|InterfaceTypeDefinitionNode|InterfaceTypeExtensionNode>
```

?

src/Validator/Rules/QueryComplexity.php
```diff
@@ -54,7 +56,8 @@ public function getVisitor(ValidationContext $context)
     {
         $this->context = $context;
 
-        $this->variableDefs     = new ArrayObject();
+        // @phpstan-ignore-next-line Initializing with an empty array does not set the generic type
```
simPod, line 59: I know there were some fixes related to this in phpstan recently. Can you recheck it still applies?

simPod (commented):
_needs rebase_

src/Type/Definition/Directive.php
spawnia: Perhaps, let's discuss this further in https://github.com/webonyx/graphql-php/issues/811. If we decide to use it, we can make a sweeping change throughout the codebase and replace every instance of `array<int, ` with `list<`.

As far as this PR is concerned, i don't want it blocked by this issue and am thus using the same notation that we already use in a few dozen places.

src/Type/SchemaConfig.php
spawnia: It is unspecified, because it does not matter. The index is only used for debugging, we don't have to constrain it unnecessarily.

src/Type/SchemaConfig.php
spawnia: It is the same type we are discussing in https://github.com/webonyx/graphql-php/pull/770#discussion_r602849388

src/Type/Definition/Directive.php
simPod: > i don't want it blocked by this issue

yup, np. 

_But we can't just change `array<int, T>` to `list<T>` as it's not the same._

src/Type/SchemaConfig.php
simPod: IMO we should use as much narrow type as possible. If index is always int, we should specify it. `array<Type>` is equal to `array<mixed, Type>` so the type of key is as narrow as possible.

src/Type/SchemaValidationContext.php
spawnia: It was even more complex, but i found and fixed a bug while writing out the type.

src/Validator/Rules/QueryComplexity.php
spawnia: Still fails.

src/Type/SchemaConfig.php
spawnia: This property accepts an array given by users of this library, they can use whatever keys they want. It is not significant for our purposes.

tests/Type/ValidationTest.php
```diff
@@ -3103,6 +3103,26 @@ public function testRejectsASchemaWithDirectiveDefinedMultipleTimes()
         );
     }
 
+    public function testRejectsASchemaWithDirectivesWithWrongArgs()
```
spawnia, line 3106: This case was bugged before, the combination of strict typing and code coverage led me to find this.

spawnia: Coverage really was essential here, found one more bug.

src/Type/SchemaValidationContext.php
simPod: hah, now it's ultimate 😄

vladar:
@spawnia Is it ready to go?

@simPod are you good with this PR?

simPod (approved):
👌🏾

PR #772, opened 2021-01-29
Fix keys ordering in input object

maxperrimond:
Little fix to keep same keys ordering in object inputs and to not be based on fields order in schema definition.


spawnia:
The order in which a client passes arguments is prescribed to be non-significant by the GraphQL spec. This change goes in the wrong direction.

PR #777, opened 2021-02-12
Clarify how NodeList accepts array's but returns Node's

spawnia:
The current version of PHPStan flagged this code as wrong, but it is not: https://github.com/nuwave/lighthouse/blob/7fb892638e8bcbcbfc369ba5349f34764ff0760f/src/Schema/AST/DocumentAST.php#L119-L127

This PR relaxes the type requirements for what is passed into `NodeList` and simplifies its implementation and thus performance in one part.


simPod (approved):

Kingdutch:
I think this would break some of the code I've recently written, which relies on the return type of a `NodeList` to always be a `Node` as per the annotations, rather than possibly a nested array (which would be valid in the new type annotations). If I have to account for nested arrays or other types that code would become wildly more complex.

With that in mind it feels as if this change may be moving us in the wrong direction?

There's a TODO in the code above which argues for using a different type/code in the visitor so that this class itself can be more strictly typed: https://github.com/webonyx/graphql-php/pull/777/files#diff-c1a0fb68d110ee6e15a8de8cb304abd3a708b68accc690cae1db46076e5dc1e1L61 (which would have my preference).

spawnia:
@Kingdutch look more closely. The return type is still always `Node`, it is just that for performance reasons the `NodeList` also accepts plain `array`, which are lazily converted to `Node`.

Kingdutch:
Ah, I see now. I thought `merge` provided a way to get around `offsetGet`, but crucially all item access for NodeList goes through `offsetGet` to ensure nested arrays are unwrapped 👍

PR #778, opened 2021-02-13
Add typehints to SchemaPrinter

spawnia:
I am currently working on Apollo Federation support for Lighthouse, that requires extending the `SchemaPrinter`. In places, it was hard to understand what was going on, so i added types.


src/Utils/SchemaPrinter.php
```diff
@@ -211,7 +218,7 @@ protected static function printDescription(array $options, $def, $indentation =
     }
 
     /**
-     * @return string[]
+     * @return array<int, string>
```
simPod, line 221: we can use list<> eventually

simPod (approved):

src/Utils/SchemaPrinter.php
spawnia: Still not in PHPStan, but it would be nice.

PR #788, opened 2021-03-12
Update the debug flag name

AdrienPoupa:
When passing 'debug' to the StandardServer it fails:

```
Unknown server config option "debug" {"exception":"[object] (GraphQL\\Error\\InvariantViolation(code: 0): Unknown server config option \"debug\" at /vol/website/vendor/webonyx/graphql-php/src/Server/ServerConfig.php:50)"}
```

This is because the option is now called debugFlag


simPod (approved):

spawnia:
@vladar what is the process for actually updating the website?

vladar:
@spawnia It should be as simple as installing [mkdocs](https://www.mkdocs.org/user-guide/deploying-your-docs/) and running `mkdocs gh-deploy` (in the repository root folder).

simPod:
So we can MB convert it into an Action to be run either on Release or Push to master 🤔

PR #791, opened 2021-03-24
Bump phpstan

simPod:
- fix new and old Missing return native typehint errors
- old phpstan versioned did not run under php 8


spawnia (approved):
Straightforward change

PR #800, opened 2021-03-25
Validation messages translation

### LINKED ISSUE #711: Validation messages translation

LastDragon-ru:
Probably the most simple solution :)

1) Add

```php
<?php declare(strict_types = 1);

namespace App\GraphQL;

use GraphQL\Validator\Rules\ValuesOfCorrectType;

use function sprintf;

class ValuesOfCorrectTypeTranslated extends ValuesOfCorrectType {
    public function getName() {
        return ValuesOfCorrectType::class;
    }

    public static function badArgumentValueMessage($typeName, $valueName, $fieldName, $argName, $message = null) {
        return sprintf('Свойство "%s" аргумент "%s" должен быть %s, но передан %s', $fieldName, $argName, $typeName, $valueName);
    }
}

```

2) Add rule

```php
DocumentValidator::addRule(new ValuesOfCorrectTypeTranslated());
```

3) For lighthouse:

```php
<?php declare(strict_types = 1);

namespace App\GraphQL;

use GraphQL\Validator\Rules\ValuesOfCorrectType;
use Nuwave\Lighthouse\Execution\ValidationRulesProvider;

class ProvidesValidationRulesTranslated extends ValidationRulesProvider {
    public function validationRules(): ?array {
        return [
                ValuesOfCorrectType::class => new ValuesOfCorrectTypeTranslated(),
            ] + parent::validationRules();
    }
}
```

And in AppServiceProvider:

```php
$this->app->bind(ProvidesValidationRules::class, ProvidesValidationRulesTranslated::class);
```

Voila 
image

Closes #711, overrides #713 and #718

### DIFF
--- a/tests/Validator/ValuesOfCorrectTypeTest.php
+++ b/tests/Validator/ValuesOfCorrectTypeTest.php
@@ -1651,9 +1651,6 @@ class ValuesOfCorrectTypeTest extends ValidatorTestCase
         self::assertTrue($errors[0]->isClientSafe());
     }
 
-    /**
-     * @see it('error messages can be overwriten')
-     */
     public function testOverwriting() : void
     {
         DocumentValidator::addRule(
--- a/tests/Validator/ValuesOfCorrectTypeTest.php
+++ b/tests/Validator/ValuesOfCorrectTypeTest.php
@@ -6,6 +6,7 @@ namespace GraphQL\Tests\Validator;
 
 use GraphQL\Error\FormattedError;
 use GraphQL\Language\SourceLocation;
+use GraphQL\Validator\DocumentValidator;
 use GraphQL\Validator\Rules\ValuesOfCorrectType;
 
 class ValuesOfCorrectTypeTest extends ValidatorTestCase
@@ -1644,6 +1645,44 @@ class ValuesOfCorrectTypeTest extends ValidatorTestCase
         ',
             [
                 $this->badValue('String', '2', 2, 50),
+            ]
+        );
+
+        self::assertTrue($errors[0]->isClientSafe());
+    }
+
+    /**
+     * @see it('error messages can be overwriten')
+     */
+    public function testOverwriting() : void
+    {
+        DocumentValidator::addRule(
+            new class() extends ValuesOfCorrectType
+            {
+                public function getName()
+                {
+                    return (new ValuesOfCorrectType())->getName();
+                }
+
+                public static function badArgumentValueMessage($typeName, $valueName, $fieldName, $argName, $message = null)
+                {
+                    return 'overwritten';
+                }
+            }
+        );
+
+        $errors = $this->expectInvalid(
+            self::getTestSchema(),
+            null,
+            '
+            {
+              complicatedArgs {
+                stringArgField(stringArg: 1)
+              }
+            }
+            ',
+            [
+                $this->badValueWithMessage('overwritten', 4, 43),
             ]
         );
 
--- a/src/Validator/DocumentValidator.php
+++ b/src/Validator/DocumentValidator.php
@@ -300,7 +300,7 @@ class DocumentValidator
         $emptyDoc    = new DocumentNode(['definitions' => []]);
         $typeInfo    = new TypeInfo($emptySchema, $type);
         $context     = new ValidationContext($emptySchema, $emptyDoc, $typeInfo);
-        $validator   = new ValuesOfCorrectType();
+        $validator   = self::getRule(ValuesOfCorrectType::class);
         $visitor     = $validator->getVisitor($context);
         Visitor::visit($valueNode, Visitor::visitWithTypeInfo($typeInfo, $visitor));
 
--- a/src/Validator/DocumentValidator.php
+++ b/src/Validator/DocumentValidator.php
@@ -300,7 +300,7 @@ class DocumentValidator
         $emptyDoc    = new DocumentNode(['definitions' => []]);
         $typeInfo    = new TypeInfo($emptySchema, $type);
         $context     = new ValidationContext($emptySchema, $emptyDoc, $typeInfo);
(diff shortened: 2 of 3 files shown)


spawnia:
Can you add a test to ensure overwriting `ValuesOfCorrectType`?

LastDragon-ru:
> Can you add a test to ensure overwriting ValuesOfCorrectType?

Do you mean for DocumentValidator?

spawnia:
Yeah. Basically what you wrote in the description:
- extend `ValuesOfCorrectType`
- register your custom implementation
- trigger a document validation error
- assert the message comes from your custom implementation

LastDragon-ru:
> trigger a document validation error

Could you please give me an example how I can do it?

tests/Validator/ValuesOfCorrectTypeTest.php
```diff
@@ -1649,4 +1650,42 @@ public function testListVariablesWithInvalidItem() : void
 
         self::assertTrue($errors[0]->isClientSafe());
     }
+
+    /**
+     * @see it('error messages can be overwriten')
```
spawnia, line 1655: Is there such a test in the graphl-js reference implementation?

tests/Validator/ValuesOfCorrectTypeTest.php
```diff
@@ -1649,4 +1650,42 @@ public function testListVariablesWithInvalidItem() : void
...
+     */
+    public function testOverwriting() : void
+    {
+        DocumentValidator::addRule(
+            new class() extends ValuesOfCorrectType
+            {
+                public function getName()
+                {
+                    return (new ValuesOfCorrectType())->getName();
```
spawnia, line 1664: Why is this function overwritten at all?

LastDragon-ru: Because `getName()` used by `DocumentValidator::addRule()` and we want to replace `ValuesOfCorrectType` completely.

tests/Validator/ValuesOfCorrectTypeTest.php
LastDragon-ru: Probably no 😁

tests/Validator/ValuesOfCorrectTypeTest.php
LastDragon-ru: And the default realization of `getName()` uses `static::class`

https://github.com/webonyx/graphql-php/blob/65428ce4da4ba6918a20f8f638345e894d2575d7/src/Validator/Rules/ValidationRule.php#L16-L19

tests/Validator/ValuesOfCorrectTypeTest.php
spawnia: Then this comment can go. They are only added to other tests that directly mirror their implementation.

tests/Validator/ValuesOfCorrectTypeTest.php
spawnia: That seems like an annoying gotcha for users that extend this class.

LastDragon-ru: 🤷‍♂️ Maybe something like?

```
class ValuesOfCorrectType extends ValidationRule
{
    // (1)
    /** @var string */
    protected $name = self::class;
    
    // (2) or this (probably a breaking change) 
    public function __construct() { 
        $this->name = self::class;
    }
```

spawnia:
Thanks for this POC, I will make sure this way of overwriting rules keeps working - see https://github.com/webonyx/graphql-php/blob/a5d64c77e77c9d35e6179e5791b23a50c301c760/tests/Validator/CustomRuleTest.php#L19-L55

PR #801, opened 2021-03-26
Able to extends ReferenceExecutor and overwrite  resolveFieldValueOrError method

4n70w4:
Case: need overwrite `resolveFieldValueOrError` without try catch block and handle exception in my own error handler from application. Without this patch, it is impossible to override this behavior.

```php

class MyExecutor extends ReferenceExecutor {



    /**
     * Isolates the "ReturnOrAbrupt" behavior to not de-opt the `resolveField` function.
     * Returns the result of resolveFn or the abrupt-return Error object.
     *
     * @param mixed $rootValue
     *
     * @return Throwable|Promise|mixed
     */
    protected function resolveFieldValueOrError(
        FieldDefinition $fieldDef,
        FieldNode $fieldNode,
        callable $resolveFn,
        $rootValue,
        ResolveInfo $info
    ) {
        // Build a map of arguments from the field.arguments AST, using the
        // variables scope to fulfill any variable references.
        $args         = Values::getArgumentValues(
            $fieldDef,
            $fieldNode,
            $this->exeContext->variableValues
        );
        $contextValue = $this->exeContext->contextValue;

        return $resolveFn($rootValue, $args, $contextValue, $info);
    }



}
```


spawnia:
I highly recommend you to not change that behaviour and instead read the GraphQL specification thoroughly and try to understand why error handling in GraphQL works in this particular way: http://spec.graphql.org/draft/#sec-Execution. In short: GraphQL queries are complex and may contain many fields. An error in one field should not abort the entire query.

Regardless of that, I generally think that opening up library code for extensiblity is a good thing. We should apply that consistently throughout the class though.

4n70w4:
I make it not for error in fields. For example I have query:


```
class Status {

    /**
     * @param  null  $_
     * @param  array<string, mixed>  $args
     */
    public function __invoke($_, array $args) {
        throw new \Exception('dcdcd');
    }
 ```
 
 With default `ReferenceExecutor` I get error response:
 
 ```json
{
  "errors": [
    {
      "debugMessage": "dcdcd",
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "status"
      ],
      "trace": [
        {
```

I don't know what the format is, this is not my error handler. I want to turn it off.

But with `MyExecutor` i get:

```json
{
  "errors": [
    {
      "id": "e84880e179a64be98d1f421b83252f23",
      "status": 500,
      "title": "Internal Server Error",
      "detail": "An error has occurred and this resource cannot be displayed."
    }
  ]
}
```

This is my application's error handler. I need it.

spawnia:
Please read the spec. Your format is not compliant.

4n70w4:
Even Facebook and Instagram display errors in his own format without compliant the spec.
Do I need to remind that grapql was developed by facebook.

spawnia:
Standard tools are not going to work as well when deviating from the specification.

Anyways, concerning this PR: every instance of `private` in this class should be `protected`, same with `self` to `static`.

4n70w4:
Okay, I change every instance of `private` to `protected`, and `self` to `static`.

A little about my case. I am using only a small part of the graphQL capabilities for one of the projects. First of all, the schema as a contract and documentation for developers.

All my graphQL requests are stored in the repository, they are verified, tested and approved. The server is under authorization. Requests to it only from trusted services. All queries always return only 1 or 0 entities by ID. The entity either returns completely or does not return. There are no other types of requests and are not planned.

GraphQL is no more than 5% of the code and traffic from the entire volume of the projects. It is more important to have the same error handling system for all projects and components.

During the 2 years of the project's existence, a graphQL spec error format has never been required and has not been useful.

4n70w4:
I suppose this line remains as a private.

https://github.com/webonyx/graphql-php/blob/f93879b5a0033db1662c79e71756effc2cd35b7c/src/Executor/ReferenceExecutor.php#L107

spawnia (approved):

4n70w4:
@spawnia Thanks. It would be very nice to make a release.
To update the dependency through the composer.
The last release was almost 2 months ago.
Many changes have accumulated.

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v14.6.0

PR #808, opened 2021-03-27
Prettify markdown in GitHub action

spawnia:


.github/workflows/format.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+        with:
+          ref: "${{ github.head_ref }}"
+
+      - name: "Run prettier on markdown files"
+        uses: "creyD/prettier_action@v2.2"
+        with:
+          prettier_options: "--write --tab-width=2 *.md **/*.md"
```
simPod, line 20: I'd prefer it did not write and only notify me instead. Personally, I prefer to commit on my own.

spawnia: That would require setting up prettier locally. I think we should rather make contributing as simple as possible, even if that means giving up a bit of control. My experience with automatic code style fixes has been very positive, virtually all projects I participate in have them.

simPod: I have the exactly opposite experience with projects. Can't prettier be just dry run with non zero error code and some output?

mfn: FWIW I'm with @spawnia here: I've the same experience as a contributor as well as a maintainer. 

Why cause friction if we already have the tools to fix it?

simPod: All I'm saying it should not touch my work without asking. We can allow people to run it but it should not modify their branches. It forces me to change my workflow so personally I'm against anything like it.

We can run in only for master pushes eventually. That way the markdown would always be fixed there while keeping user PRs pure.

spawnia: Running it for `master` only is definitely an option. The tradeoff there would be that the commit history gets uglier - would be a shame, considering how nice it is now: https://github.com/webonyx/graphql-php/commits/master

mfn: LGTM! needs a different token though (e.g. a PAT), the default can't push to protected branches AFAIK

spawnia: Pushing to `master` worked fine.

PR #809, opened 2021-03-27
Fix codestyle everywhere

spawnia:


simPod (approved):

simPod (commented):
We might exclude "native type hint" sniff for now and do it separately so the scope of this PR is smaller.

spawnia:
> We might exclude "native type hint" sniff for now and do it separately so the scope of this PR is smaller.

Good call on splitting the PR. I am chopping of smaller bits by directory now.

spawnia:
Superseded by smaller PRs.

PR #812, opened 2021-03-29
Open ReferenceExecutor for extending (#801)

spawnia:
(cherry picked from commit c19cd7c10e4cad63a1fb74a95ef3fdbeeabe5672)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 #### Unreleased
 
+Feat:
+- Open ReferenceExecutor for extending
+
 #### 14.5.1
 
 Fix:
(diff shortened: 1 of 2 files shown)


simPod (commented):
_I don't fancy changes like this as it increases the scope where we have to worry about BC._

_Though I have not investigated possible alternatives so can't offer a solution._

spawnia:
Methods are properties that are not marked with `@api` are not part of the BC guarantees. If someone wants to mess with internals, we should let them - but they are on their own when things go wrong.

simPod:
Aight so we can eventually revert this while not breaking BC promise. 🆗

PR #817, opened 2021-03-29
Fix docs deployment

spawnia:


mkdocs.yml
```diff
@@ -23,4 +27,5 @@ pages:
 # - Performance tips: performance.md
 - How it works: how-it-works.md
 - Class Reference: reference.md
-theme: readthedocs
+theme:
+  name: material
```
simPod, line 31: It looks modern now :D

spawnia: Wasn't planning to change it, but my first try did error out with something related to the theme and i just copied their example. I like the style

simPod:
@spawnia not sure since when this happens but I see 404 at https://webonyx.github.io/

PR #819, opened 2021-03-29
Only report Codecov status on pulls

spawnia:
Otherwise, it makes the repository page look really bad.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        only_pulls: true
+comment: false
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,4 +3,3 @@ coverage:
     project:
       default:
         only_pulls: true
-comment: false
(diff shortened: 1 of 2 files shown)


.codecov.yml
```diff
@@ -0,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        only_pulls: true
+comment: false
```
simPod, line 6: this is IMO useful, provides quick overview on what's changed in a comment

simPod (commented):
How does it look bad? Not sure what this change does exactly 🤔

spawnia:
For some reason, codecov keeps failing even on config only changes and shows the ci of the project page to be failing. When I am shopping for packages, an always green pipeline is something I value.

.codecov.yml
spawnia: I am fine with keeping them in, sure

simPod:
Got it. I'll try to come up with a fix and meanwhile we can ship this.

PR #823, opened 2021-03-29
Remove deprecated GraphQL::executeAndReturnResult()

simPod:

PR #824, opened 2021-03-29
Remove deprecated GraphQL::execute()

simPod:


src/GraphQL.php
```diff
@@ -177,53 +177,6 @@ public static function promiseToExecute(
...
-            $result = $result->then(static function (ExecutionResult $r): array {
-                return $r->toArray();
-            });
-        }
-
-        return $result;
-    }
-
     /**
      * @deprecated renamed to executeQuery()
```
spawnia, line 181: Even worse 😆

simPod: That should be in https://github.com/webonyx/graphql-php/pull/823, hm? 🙃

spawnia (approved):

PR #825, opened 2021-03-29
Drop Licence badge

simPod:
It's useless as every decent UI shows Licence by default

Github:

image

Packagist:

image


spawnia:
Agreed. The repository itself also has the `LICENSE` file, in case no UI is available.

PR #818, opened 2021-03-29
Restructure docs

spawnia:

PR #826, opened 2021-03-30
README ci.yml

kitikhunp:


simPod (changes requested):
I think this is a mistake and should be closed. Can you eventually explain why it should not?

spawnia:
I don't know what this or the nonsensical comments in https://github.com/webonyx/graphql-php/pull/818 should accomplish. To me it looks like spam and I reported it as such.

PR #831, opened 2021-04-03
Add support for __typename field in QueryPlan

simPod:
I have found this old commit (committed 1y ago) and I'm not entirely sure what I've been trying to achieve 🤷🏾‍♀️ Though, it still makes sense to me. Having `__typename` in query should also show it in QueryPlan, right?


spawnia:
It could also be considered a feature to not have it there. This field is resolved by the library, not the user - why should they care?

simPod:
@spawnia what do you exactly mean by 

> This field is resolved by the library, not the user - why should they care?

? Personally I often have `__typename` in queries I write for client apps (useful for typeguards).

spawnia:
Sure, it is queried by clients. But when developing the server, you do not have to care about it. This library adds it for you.

simPod:
Aight, let's close this for now. I wonder I must have read a query plan on server and the `__typename` field was missing though I don't remember why I needed it back then. Thanks!

crissi:
if you have __typename in your query and you use queryPlan() it will fail right now, so I would say this PR is essential.

See also: https://github.com/webonyx/graphql-php/issues/544

spawnia:
Let's differentiate. I think that:
1. `QueryPlan` should not crash when `__typename` is used in the query
2. The output of `QueryPlan` should not contain `__typename`

simPod:
I think you're right. I remember what I tried to use QueryPlan for, will try to implement it this week and resolve 2.

simPod:
So "this week" is now.

simPod:
https://github.com/webonyx/graphql-php/pull/863

PR #834, opened 2021-04-04
Fix "Only booleans are allowed" PHPStan errors

simPod:


src/Server/Helper.php
```diff
@@ -79,17 +81,17 @@ public function parseHttpRequest(?callable $readRawBodyFn = null)
...
-                $rawBody    = $readRawBodyFn ?
-                    $readRawBodyFn()
-                    : $this->readRawBody();
+                $rawBody    = $readRawBodyFn === null
+                    ? $this->readRawBody()
+                    : $readRawBodyFn();
                 $bodyParams = json_decode($rawBody ?? '', true);
 
-                if (json_last_error()) {
+                if (json_last_error() !== JSON_ERROR_NONE) {
```
spawnia, line 94: Off-topic: We could consider using https://github.com/thecodingmachine/safe.

src/Validator/Rules/PossibleFragmentSpreads.php
```diff
@@ -142,14 +142,13 @@ public static function typeIncompatibleAnonSpreadMessage($parentType, $fragType)
     private function getFragmentType(ValidationContext $context, $name)
     {
         $frag = $context->getFragment($name);
-        if ($frag) {
-            $type = TypeInfo::typeFromAST($context->getSchema(), $frag->typeCondition);
-            if ($type instanceof CompositeType) {
-                return $type;
-            }
+        if ($frag === null) {
+            return null;
```
spawnia, line 146: A test to cover this case would be great.

spawnia:
Not quite confident the changes are correct in places where there is no test coverage.

src/Server/Helper.php
simPod: Not sure if worth though. There's some lack of static analysis tools support / it's often behind with updates. 

Eg. there are some dynamic extensions in phpstan/psalm that are not working for Safe and must be ported there. Also this example https://github.com/hectorj/safe-php-psalm-plugin/pull/5. 

On my private projects I use it but for libraries I'm trying to consider these cons.

shmax:
Nice! Way to chip away at the baseline file!

simPod:
> Not quite confident the changes are correct in places where there is no test coverage.

@spawnia can you point at them?

src/Validator/Rules/PossibleFragmentSpreads.php
simPod: Will add though this is covered by types (improved in b2d560e2af9cd7512e326005079264ffa447c41e)

simPod: Looking at it I don't think it's currently possible to invoke that line through our code in src.

spawnia: If there is some invariant making this state impossible, we can add a commented type assertion and remove the condition.

spawnia:
> > Not quite confident the changes are correct in places where there is no test coverage.
> 
> @spawnia can you point at them?

https://github.com/webonyx/graphql-php/pull/834/files shows me inline Codecov comments on some specific lines, can you see those?

src/Validator/Rules/PossibleFragmentSpreads.php
simPod: What is the way to do it here? `assert` is usually not evaluated in runtime environments. In my projects I use runtime assertion libs. Do we have an exception for it?

simPod:
Ah yes, opened it again and some appeared, they were not there before 😄 . Will review it.

src/Validator/Rules/PossibleFragmentSpreads.php
spawnia: We are using `assert` in exactly one place as of now. I would go with a PHPDoc to override the type if I know it can not ever be `null` because of previously executed code.

simPod: Well, it can be executed but one must try very hard and won't happen when using the typical flow.

Either overwriting types using PHPDoc is wrong as it might rot over time. I'm using psalm to prevent me doing it.

TBH I'd keep it as is. It's type covered and method consumers handle `null` value. `getFragmentType()` can return `null` in theory. In practice it won't.

src/Server/Helper.php
```diff
@@ -79,17 +81,17 @@ public function parseHttpRequest(?callable $readRawBodyFn = null)
...
+                $rawBody    = $readRawBodyFn === null
+                    ? $this->readRawBody()
+                    : $readRawBodyFn();
                 $bodyParams = ['query' => $rawBody ?? ''];
             } elseif (stripos($contentType, 'application/json') !== false) {
-                $rawBody    = $readRawBodyFn ?
-                    $readRawBodyFn()
-                    : $this->readRawBody();
+                $rawBody    = $readRawBodyFn === null
+                    ? $this->readRawBody()
```
simPod, line 90: In order to test these lines we would have to modify PHP globals and mock `php://input`. I think it's not worth and out of scope here.

src/Type/SchemaValidationContext.php
```diff
@@ -403,9 +401,9 @@ static function ($schemaLocation) use ($location): bool {
                 }
             );
             if (! $includes) {
-                $errorNodes = $schemaDirective->astNode
-                    ? [$directive, $schemaDirective->astNode]
-                    : [$directive];
+                $errorNodes = $schemaDirective->astNode === null
+                    ? [$directive]
```
simPod, line 405: This has type coverage

PR #839, opened 2021-04-05
Use default codecov config but keep running project status check only on PRs

simPod:
https://docs.codecov.com/docs/codecov-yaml#default-yaml


simPod:
@vladar is something set in https://codecov.io/account/gh/webonyx/yaml

PR #842, opened 2021-04-09
Add types to Type

spawnia:


src/Type/Schema.php
```diff
@@ -410,11 +410,16 @@ private function defaultTypeLoader(string $typeName): ?Type
...
+     * @param Type|callable $type
+     *
+     * @template T of Type
+     * @phpstan-param T|callable():T $type
+     * @phpstan-return T
      */
     public static function resolveType($type): Type
     {
         if ($type instanceof Type) {
+            /** @phpstan-var T $type */
```
simPod, line 422: I'd expect phpstan to remember the type information from 

```php
* @param Type|callable $type
* @phpstan-param T|callable():T $type
```

🤔

src/Type/Definition/NonNull.php
```diff
@@ -26,16 +26,20 @@ public function toString(): string
...
+
+        return $type;
     }
 
-    /**
-     * @return (NullableType&Type)
-     */
     public function getWrappedType(bool $recurse = false): Type
     {
+        /** @var NullableType&Type $type */
```
simPod, line 42: is this needed? It's typed above

image

tests/Executor/ExecutorTest.php
```diff
@@ -54,7 +54,7 @@ public function testExecutesArbitraryCode(): void
             });
         };
 
-        $data = [
+        $data     = [
```
simPod, line 57: wasn't it compliant with CS already before?

simPod (approved):

src/Type/Schema.php
spawnia: Me too, it seems as though the `instanceof` check erases the generic type. I still think it is better to have the function be well-typed for consumers.

src/Type/Definition/NonNull.php
spawnia: I would rather omit it, PHPStan will not let me.

tests/Executor/ExecutorTest.php
spawnia: I tried reverting, seems like the newest version changes this. This is one of the reasons why I do not like alignment, my preference would be to left-align everything always.

simPod: After some time, I agree. There's even no proper IDE support for this kind of alignment. It some cases it's really handy but overall not worth I guess.

src/Type/Definition/NonNull.php
simPod: Aight, let's skip this for now.

Eg. psalm can detect useless `@var` annotations so we can automatically clean this up later

tests/Executor/ExecutorTest.php
spawnia: As far as I can tell, PHP_CodeSniffer has no option to force left-alignment. Wading through their docs is a total pain in the neck though, perhaps we should go with php-cs-fixer.

PR #843, opened 2021-04-15
Validate default values match argument type

spawnia:
See https://github.com/graphql/graphql-js/issues/2606 for discussion.

Since the pace of development in `graphql-js` has not been great lately and we just hit this issue, I think we could charge ahead and pioneer a solution for it 💪


spawnia:
Relevant: https://github.com/graphql/graphql-js/pull/3049

spawnia:
See https://github.com/graphql/graphql-js/pull/3814.

PR #847, opened 2021-04-23
Fix overly eager validation of repeatable directive usage

spawnia:
@stayallive

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
 # Changelog
 
 #### Unreleased
+Changed:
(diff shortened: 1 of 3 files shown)


simPod (approved):
I see nothing wrong at the first sight though it's kind of over my graphql knowledge

spawnia:
Thank you @simPod. As avid directive users, we recently ran into this validation problem in a newer release of Lighthouse. Having `repeatable` on a directive specifically says it is fine to use multiple times.

simPod:
I guess that's it, I have not used a directive yet :D Thanks

PR #854, opened 2021-04-26
Improve extendability of validator rules

### LINKED ISSUE #851: Remove static from GraphQL\Validator\Rules\ValuesOfCorrectType
A lot of `static` methods are used in `ValuesOfCorrectType`, but when I take a closer look I can't see why. There seems to be no added benefit in having these static. Would it be possible to remove all the `static` usage from this class?

The reason I ask is because I want to extend it in order to change the way some error messages are created. This feels related to #711, but I want to do more than a simple string replacement.

wimski:
Resolves #851 

> Yes. Accepting PRs, but preferrably one that opens up all validation rules for extensibility to ensure consistency.

### DIFF
--- a/src/Validator/Rules/CustomValidationRule.php
+++ b/src/Validator/Rules/CustomValidationRule.php
@@ -10,7 +10,7 @@ use GraphQL\Validator\ValidationContext;
 class CustomValidationRule extends ValidationRule
 {
     /** @var callable */
-    private $visitorFn;
+    protected $visitorFn;
 
     public function __construct($name, callable $visitorFn)
     {
--- a/src/Validator/Rules/CustomValidationRule.php
+++ b/src/Validator/Rules/CustomValidationRule.php
@@ -4,7 +4,8 @@ declare(strict_types=1);
 
 namespace GraphQL\Validator\Rules;
 
-use GraphQL\Error\Error;
+use GraphQL\Language\AST\Node;
+use GraphQL\Language\VisitorOperation;
 use GraphQL\Validator\ValidationContext;
 
 class CustomValidationRule extends ValidationRule
@@ -12,19 +13,17 @@ class CustomValidationRule extends ValidationRule
     /** @var callable */
     protected $visitorFn;
 
-    public function __construct($name, callable $visitorFn)
+    /**
+     * @param callable(ValidationContext): (array<string, callable(Node): VisitorOperation|mixed|null>|array<string, array<string, callable(Node): VisitorOperation|mixed|null>>) $visitorFn
+     */
+    public function __construct(string $name, callable $visitorFn)
     {
         $this->name      = $name;
         $this->visitorFn = $visitorFn;
     }
 
-    /**
-     * @return Error[]
-     */
     public function getVisitor(ValidationContext $context)
     {
-        $fn = $this->visitorFn;
-
-        return $fn($context);
+        return ($this->visitorFn)($context);
     }
 }
--- a/src/Validator/Rules/DisableIntrospection.php
+++ b/src/Validator/Rules/DisableIntrospection.php
@@ -52,6 +52,6 @@ class DisableIntrospection extends QuerySecurityRule
 
     protected function isEnabled()
     {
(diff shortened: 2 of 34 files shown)


mfn (commented):
I love opening up the things, aligns with my philosophy.

I guess however it's important to keep in mind that only `@api` marked elements are covered by semver/backwards compatibility promises.
(I kind of made this up from observation on this repository,  could not quickly find a proper doc if this is the case -> all my search for `@api` only yielded code-use)

src/Validator/Rules/DisableIntrospection.php
```diff
@@ -52,6 +52,6 @@ public static function introspectionDisabledMessage()
 
     protected function isEnabled()
     {
-        return $this->isEnabled !== self::DISABLED;
+        return $this->isEnabled !== static::DISABLED;
```
spawnia, line 55: @simPod can we turn off this sniffer?

src/Validator/Rules/CustomValidationRule.php
```diff
@@ -10,7 +10,7 @@
 class CustomValidationRule extends ValidationRule
 {
     /** @var callable */
-    private $visitorFn;
+    protected $visitorFn;
```
spawnia, line 13:
```suggestion
    protected callable $visitorFn;
```

Same with other properties, let's do it now to save you the inevitable breakage later ;)

mcg-web: It is not possible to use callable to type properties because it is [inconsistent](https://wiki.php.net/rfc/consistent_callables). AI'm not a 100% sure of this for php 8.0 but in 7.4 this not work for me, in my projects.

src/Validator/Rules/DisableIntrospection.php
simPod: Hard to say and evaluate for me as I use `final` everywhere so this is not my daily bread to deal with.

But personally I'm against this as constants are not meant to be overridden and abstract static methods should be used instead. Not sure what LSB for constants implies in the future.

src/Validator/Rules/DisableIntrospection.php
```diff
@@ -14,7 +14,7 @@ class DisableIntrospection extends QuerySecurityRule
     public const ENABLED = 1;
 
     /** @var bool */
-    private $isEnabled;
+    protected $isEnabled;
```
spawnia, line 17:
```suggestion
    protected bool $isEnabled;
```

Generally, we should switch to fully typed properties and methods before opening up this class.

src/Validator/Rules/DisableIntrospection.php
spawnia: I generally agree with that sentiment and am moving towards doing the same in projects that are self-contained. It is always possible to change the code there. Consumers of this library have no choice but to copy-paste code if we close a class for extension.

The late static binding might be a footgun here, but I feel like it only adds to the usefulness of this library and does not take away from it.

wimski: I could easily change this back to `self`, because I agree that constants shouldn't be overridden. Someone could then still extend/override the `isEnabled` method without changing the constant.

src/Validator/Rules/DisableIntrospection.php
wimski: I've typed all properties that changed from private to protected except for callable's.

src/Validator/Rules/DisableIntrospection.php
simPod: > Consumers of this library have no choice but to copy-paste code if we close a class for extension.

That would exactly be my suggestion :D to copy-paste it and do whatever you want with it, you're on your own.

My way is fully typed, fully strict, etc. blabla code so I can't encourage doing this. Don't consider me a blocker though, I'd stayed silent if you didn't mention me 🙃

spawnia:
There is more work to be done with those classes in regards to strict typing. At least the properties are done now and the member visibility is fine, so this is a good point to merge as is.

PR #856, opened 2021-04-28
Update examples/01-blog with latest best practices

spawnia:
Update the example to:
- PHP 7.4
- strict typing practices
- actually work
- be easier to setup

### DIFF
--- a/examples/01-blog/Blog/Data/DataSource.php
+++ b/examples/01-blog/Blog/Data/DataSource.php
@@ -7,11 +7,11 @@ namespace GraphQL\Examples\Blog\Data;
 use function array_filter;
 use function array_keys;
 use function array_map;
-use function array_pop;
 use function array_search;
 use function array_slice;
 use function array_values;
 use function count;
+use function end;
 use function in_array;
 use function rand;
 
@@ -22,36 +22,43 @@ use function rand;
 class DataSource
 {
     /** @var array<int, User> */
-    private static array $users          = [];
+    private static array $users = [];
 
     /** @var array<int, Story> */
-    private static array $stories        = [];
+    private static array $stories = [];
 
-    /** @var array<int, int> */
-    private static array $storyLikes     = [];
+    /** @var array<int, array<int, int>> */
+    private static array $storyLikes = [];
 
-    private static $comments       = [];
-    private static $storyComments  = [];
-    private static $commentReplies = [];
-    private static $storyMentions  = [];
+    /** @var array<int, Comment> */
+    private static array $comments = [];
+
+    /** @var array<int, array<int, int>> */
+    private static array $storyComments = [];
+
+    /** @var array<int, array<int, int>> */
+    private static array $commentReplies = [];
+
+    /** @var array<int, array<int, User|Story>> */
+    private static array $storyMentions = [];
 
     public static function init(): void
     {
         self::$users = [
-            '1' => new User([
-                'id' => '1',
+            1 => new User([
+                'id' => 1,
                 'email' => 'john@example.com',
                 'firstName' => 'John',
                 'lastName' => 'Doe',
             ]),
-            '2' => new User([
-                'id' => '2',
+            2 => new User([
+                'id' => 2,
                 'email' => 'jane@example.com',
                 'firstName' => 'Jane',
                 'lastName' => 'Doe',
             ]),
-            '3' => new User([
-                'id' => '3',
+            3 => new User([
+                'id' => 3,
                 'email' => 'john@example.com',
                 'firstName' => 'John',
                 'lastName' => 'Doe',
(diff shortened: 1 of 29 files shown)


examples/01-blog/Blog/Data/DataSource.php
```diff
@@ -1,202 +1,250 @@
...
+                'id' => 2,
                 'email' => 'jane@example.com',
                 'firstName' => 'Jane',
-                'lastName' => 'Doe'
+                'lastName' => 'Doe',
             ]),
-            '3' => new User([
-                'id' => '3',
+            3 => new User([
+                'id' => 3,
```
simPod, line 60: Having those IDs as strings were IMO better. The fact it's numeric is a coincidence.

examples/01-blog/Blog/Type/Enum/ContentFormatType.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace GraphQL\Examples\Blog\Type\Enum;
+
+use GraphQL\Type\Definition\EnumType;
+
+class ContentFormatType extends EnumType
+{
+    const FORMAT_TEXT = 'TEXT';
```
simPod, line 11: maybe we can also encourage to specify constant visibility

examples/01-blog/Blog/Type/ImageType.php
```diff
@@ -1,62 +1,63 @@
...
+                        throw new Exception('Non-null field with exception');
+                    },
+                ],
+            ],
+        ]);
     }
 
-    public function resolveUrl(Image $value, $args, AppContext $context)
+    /**
+     * @param array<void> $args
```
simPod, line 49: what does `array<void>` mean?

examples/01-blog/Blog/Type/QueryType.php
```diff
@@ -12,85 +18,97 @@ class QueryType extends ObjectType
...
 
-    public function user($rootValue, $args)
+    /**
+     * @param null              $rootValue
+     * @param array{id: string} $args
+     */
+    public function user($rootValue, array $args): ?User
     {
-        return DataSource::findUser($args['id']);
+        return DataSource::findUser((int) $args['id']);
```
simPod, line 78: If IDs were always strings, we could drop casts

examples/01-blog/graphql.php
```diff
@@ -1,74 +1,71 @@
...
+use GraphQL\Examples\Blog\Types;
+use GraphQL\GraphQL;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Schema;
 
-// Disable default PHP error reporting - we have better one for debug mode (see below)
-ini_set('display_errors', 0);
+// Replace default PHP error reporting with a better handler
+ini_set('display_errors', '0');
+set_error_handler(static function ($severity, $message, $file, $line): void {
```
simPod, line 22: Maybe this can be dropped entirely? PHP error handling is out of scope. And if we don't want to drop it, it's not clear why it's a "better handler" 🧐

simPod (approved):
👌🏾

examples/01-blog/Blog/Data/DataSource.php
spawnia: GraphQL generally seems to push towards this with the ID type. In practice, I found that databases are more performant when using numeric IDs and tend to use them pervasively.

In this example, there is no big difference. Would be fine as it was before, is also fine like this.

examples/01-blog/Blog/Type/ImageType.php
spawnia: My intention was to specify an empty array type. `array{}` did not work, do you have other suggestions?

examples/01-blog/graphql.php
spawnia: I thought about just using the standard server for this example. We have `00-hello-world` as an example of handling the HTTP aspect ourselves, but I would not recommend anyone to actually do this in a real application.

examples/01-blog/Blog/Type/ImageType.php
simPod: There are only `non-empty-array`s so no actual way to express empty array AFAIK.

The best way to type $args is IMO `array<string, mixed>`, it is not guaranteed it's empty anyway, is it?

examples/01-blog/graphql.php
simPod: So if we want to keep this, maybe we should specify why it is better handler. 🤷🏾

spawnia: I don't actually know.

simPod: hm, me neither :D umm

examples/01-blog/Blog/Type/ImageType.php
spawnia: In this case it is, since the field has no args. When well-known, I am starting to type `$args` with array shapes such as `array{foo: int, bar: string}`.

examples/01-blog/Blog/Data/DataSource.php
simPod: > I found that databases are more performant when using numeric IDs and tend to use them pervasively.

side note, depends significantly on the database, no?

examples/01-blog/Blog/Type/ImageType.php
simPod: Yup but the super type is `array<string, mixed>` and it is the closest valid type we have for this no-args thing currently I guess.

examples/01-blog/Blog/Data/DataSource.php
spawnia: Indices on numeric values are generally more efficient, subsequently `WHERE IN` or `JOIN` or faster.

examples/01-blog/Blog/Type/ImageType.php
spawnia: PHPStan seems to be ok with `array<void>` though, and it does express the intent.

simPod: PHPStan is still ok with many things like [inserting under string key when int key is declared](https://phpstan.org/r/1f31a564-7fff-475a-9875-c658e1b6078e). Would not rely on it much in cases like this.

I just remember there's `array<empty, empty>` :)

simPod: PHPStan does not seem to know it yet. I think I opened some related issue there about it but can't find it now.

https://psalm.dev/r/3c66dc8bcf

https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/#empty

spawnia: I went with `array<never>` now, `never` is the bottom type in PHPStan.

examples/01-blog/graphql.php
spawnia: I switched to StandardServer.

PR #859, opened 2021-04-30
Increase docs readability and remove outdated PHP version range

Khartir:
This MR contains two commits.

The first commit adds PHP 8 to the list of supported PHP versions in the docs. I assume this was missed when the PHP 8 support was added and not a deliberate choice? Also I didn't remove `5.5+` from the list of supported versions since it's technically still supported with older versions, but maybe it should be removed for brevity?

The second commit contains some slight changes to wording in the docs in an attempt to improve readability. But then I'm not a native speaker. Is this OK or should I drop it?

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -3,7 +3,7 @@ GraphQL is data-centric. On the very top level it is built around three major co
 **Schema**, **Query** and **Mutation**.
  
 You are expected to express your application as a **Schema** (aka Type System) and expose it
-with single a HTTP endpoint (e.g. using our [standard server](executing-queries.md#using-server)). 
+as a single HTTP endpoint (e.g. using our [standard server](executing-queries.md#using-server)). 
 Application clients (e.g. web or mobile clients) send **Queries** 
 to this endpoint to request structured data and **Mutations** to perform changes (usually with HTTP POST method).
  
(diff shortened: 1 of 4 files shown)


simPod (approved):
lgtm

docs/concepts.md
```diff
@@ -2,8 +2,8 @@
 GraphQL is data-centric. On the very top level it is built around three major concepts: 
 **Schema**, **Query** and **Mutation**.
  
-You are expected to express your application as **Schema** (aka Type System) and expose it
-with single HTTP endpoint (e.g. using our [standard server](executing-queries.md#using-server)). 
+You are expected to express your application as a **Schema** (aka Type System) and expose it
+with single a HTTP endpoint (e.g. using our [standard server](executing-queries.md#using-server)).
```
spawnia, line 6:
```suggestion
as a single HTTP endpoint (e.g. using our [standard server](executing-queries.md#using-server)). 
```

docs/index.md
```diff
@@ -18,7 +18,7 @@ All of them equally apply to this PHP implementation.
 
 # About graphql-php
 
-**graphql-php** is a feature-complete implementation of GraphQL specification in PHP (5.5+, 7.0+). 
+**graphql-php** is a feature-complete implementation of GraphQL specification in PHP (5.5+, 7.0+, 8.0).
```
spawnia, line 21:
```suggestion
**graphql-php** is a feature-complete implementation of GraphQL specification in PHP. 
```

Let's just defer that to `composer.json`, it lists the supported versions.

spawnia (changes requested):

spawnia (approved):

PR #860, opened 2021-05-07
Improve performance of subtype checks

stayallive:
When using a fragment on a interface like:

```graphql
interface Character {
  name: String!
}

fragment CharacterFragment on Character {
  name
}
```

The check in `Schema::isSubType` uses `Schema::getImplementations` which is rightly so marked `This operations requires full schema scan. Do not use in production environment.`...

This PR adds a quick short circuit to check on a `ImplementingType` object if it implements the interface which is also cached within the `ImplementingType` to prevent calling the very costly `Schema::getImplementations` preventing a full schema scan for a simple check.

Even the tests needed a little changes because less code was executed 😄 

This (in my rough tests) on a large schema (200+ types) and with a broadly used interface (20+ implementors) improved performance 4x (using Laravel Lighthouse). Using `composer bench` (reading the avg times before and after changes) I saw 5 to 10% performance gains.

### DIFF
--- a/src/Type/Schema.php
+++ b/src/Type/Schema.php
@@ -20,9 +20,11 @@ use GraphQL\Type\Definition\UnionType;
 use GraphQL\Utils\InterfaceImplementations;
 use GraphQL\Utils\TypeInfo;
 use GraphQL\Utils\Utils;
+use InvalidArgumentException;
 use Traversable;
 
 use function array_map;
+use function get_class;
 use function implode;
 use function is_array;
 use function is_callable;
@@ -525,7 +527,7 @@ class Schema
             return $abstractType->isPossibleType($maybeSubType);
         }
 
-        return false;
+        throw new InvalidArgumentException(sprintf('$abstractType must be of type UnionType|InterfaceType got: %s.', get_class($abstractType)));
     }
 
     /**
--- a/src/Type/Schema.php
+++ b/src/Type/Schema.php
@@ -524,7 +524,7 @@ class Schema
      */
     public function isSubType(AbstractType $abstractType, ImplementingType $maybeSubType): bool
     {
-        if ($abstractType instanceof InterfaceType && $maybeSubType instanceof ImplementingType) {
+        if ($abstractType instanceof InterfaceType) {
             return $maybeSubType->implementsInterface($abstractType);
         }
 
--- a/src/Type/Schema.php
+++ b/src/Type/Schema.php
@@ -524,6 +524,10 @@ class Schema
      */
     public function isSubType(AbstractType $abstractType, ImplementingType $maybeSubType): bool
     {
+        if ($abstractType instanceof InterfaceType && $maybeSubType instanceof ImplementingType) {
+            return $maybeSubType->implementsInterface($abstractType);
+        }
+
         if (! isset($this->subTypeMap[$abstractType->name])) {
             $this->subTypeMap[$abstractType->name] = [];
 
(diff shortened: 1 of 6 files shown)


spawnia:
Can we add a similar fast path for unions?

stayallive:
@spawnia yes actually, I'm not even sure `isSubType` would ever be called for another type... according to the test suite it doesn't because the test suite never comes out the other side of the 2 fast paths added now.

src/Type/Schema.php
```diff
@@ -524,22 +524,24 @@ public function isPossibleType(AbstractType $abstractType, ObjectType $possibleT
      */
     public function isSubType(AbstractType $abstractType, ImplementingType $maybeSubType): bool
     {
+        if ($abstractType instanceof InterfaceType && $maybeSubType instanceof ImplementingType) {
```
spawnia, line 527:
```suggestion
        if ($abstractType instanceof InterfaceType) {
```

spawnia:
I think we can get rid of that other code. Perhaps some unrelated change (interfaces implementing interfaces?) enabled us calling the abstract types themselves. Much nicer that way!

src/Type/Schema.php
```diff
@@ -524,22 +524,24 @@ public function isPossibleType(AbstractType $abstractType, ObjectType $possibleT
...
     public function isSubType(AbstractType $abstractType, ImplementingType $maybeSubType): bool
     {
+        if ($abstractType instanceof InterfaceType) {
+            return $maybeSubType->implementsInterface($abstractType);
+        }
+
+        if ($abstractType instanceof UnionType) {
+            return $abstractType->isPossibleType($maybeSubType);
+        }
+
```
spawnia, line 527: I think we do not need the other code, and can throw here with a message reminding future developers of this library to handle a possible 3rd `AbstractType` (not needed now).

stayallive: Wouldn't you not throw so that you can still send impossible types to this method and it just returning false for types it cannot handle? Since technically users could also use this method (it's part of `@api`).

spawnia: `AbstractType` must be a `UnionType` or `InterfaceType`. It is not part of `@api` so technically users should not implement it 😉 In more capable type systems (e.g. TypeScript), `AbstractType` is modeled as a Union type and thus impossible to use wrong.

I think returning `false` here is failing silently.

stayallive: Yep, okay I agree, something something PHP 8 right 😉 I'll change it to throw!

spawnia: Yeah, PHP 8 would allow defining the argument type as `UnionType | InterfaceType`. Still not quite as ergonomic as TypeScript - perhaps https://wiki.php.net/rfc/union_types_v2#type_aliases will be picked up at some point 🤞

spawnia (approved):
Nicely done, this will be an improvement for many users.

simPod (approved):
🚢

spawnia:
@stayallive if you need this sooner rather then later, you can backport this change to v14.

spawnia:
We have not build consensus on when we would release v15, but I guess it will be a few weeks or months until it happens. Same with a new major version of Lighthouse which would allow using it.

So far #861 looks like it is non-breaking, so if you are on a timeline then backporting both probably makes sense.

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v14.6.3

PR #861, opened 2021-05-08
Lazy field definitions

stayallive:
This PR aims to allow field definitions to be provided as a callable allowing them to defer any processing to resolve a field to as late as possible (or just in time).

Within Laravel Lighthouse for example this has huge performance improvements (especially on larger schema's) because the process of resolving a field can be quite some work and there is a potential huge cost to be paid for each field. Allowing them to be loaded just in time solves that by paying the cost only when needed instead of always on every query.

This feature is completely optional by the implementor of course no functionality has actually changed except for that now a callable is a valid type for a field definition which allows the implementor to lazily resolve the field config if required.

### DIFF
--- a/src/Type/Definition/HasFieldsType.php
+++ b/src/Type/Definition/HasFieldsType.php
@@ -15,7 +15,7 @@ interface HasFieldsType
 
     public function hasField(string $name): bool;
 
-    public function tryGetField(string $name): ?FieldDefinition;
+    public function findField(string $name): ?FieldDefinition;
 
     /**
      * @return array<string, FieldDefinition>
--- a/src/Type/Definition/HasFieldsType.php
+++ b/src/Type/Definition/HasFieldsType.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Type\Definition;
+
+use GraphQL\Error\InvariantViolation;
+
+interface HasFieldsType
+{
+    /**
+     * @throws InvariantViolation
+     */
+    public function getField(string $name): FieldDefinition;
+
+    public function hasField(string $name): bool;
+
+    public function tryGetField(string $name): ?FieldDefinition;
+
+    /**
+     * @return FieldDefinition[]
+     *
+     * @throws InvariantViolation
+     */
+    public function getFields(): array;
+
+    /**
+     * @return string[]
+     *
+     * @throws InvariantViolation
+     */
+    public function getFieldNames(): array;
+}
--- a/src/Type/Definition/HasFieldsType.php
+++ b/src/Type/Definition/HasFieldsType.php
@@ -25,7 +25,7 @@ interface HasFieldsType
     public function getFields(): array;
 
     /**
-     * @return string[]
+     * @return array<int, string>
      *
      * @throws InvariantViolation
      */
--- a/src/Type/Definition/HasFieldsType.php
+++ b/src/Type/Definition/HasFieldsType.php
@@ -18,7 +18,7 @@ interface HasFieldsType
     public function tryGetField(string $name): ?FieldDefinition;
 
     /**
-     * @return FieldDefinition[]
+     * @return array<string, FieldDefinition>
      *
      * @throws InvariantViolation
      */
--- a/src/Type/Definition/HasFieldsTypeImpl.php
+++ b/src/Type/Definition/HasFieldsTypeImpl.php
@@ -1,0 +1,94 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Type\Definition;
+
+use GraphQL\Error\InvariantViolation;
+use GraphQL\Utils\Utils;
+
+use function array_keys;
+
+trait HasFieldsTypeImpl
+{
+    /**
+     * Lazily initialized.
+     *
+     * @var FieldDefinition[]|null
+     */
+    private $fields;
+
+    /**
+     * @throws InvariantViolation
+     */
+    public function getField(string $name): FieldDefinition
+    {
+        Utils::invariant($this->hasField($name), 'Field "%s" is not defined for type "%s"', $name, $this->name);
+
+        return $this->tryGetField($name);
+    }
+
+    public function tryGetField(string $name): ?FieldDefinition
+    {
+        $this->initializeFields();
+
+        if (! isset($this->fields[$name])) {
+            return null;
+        }
+
+        if ($this->fields[$name] instanceof UnresolvedFieldDefinition) {
+            $this->fields[$name] = $this->fields[$name]->resolve();
+        }
+
+        return $this->fields[$name];
+    }
+
+    public function hasField(string $name): bool
+    {
+        $this->initializeFields();
+
+        return isset($this->fields[$name]);
+    }
+
+    /**
+     * @return FieldDefinition[]
+     *
+     * @throws InvariantViolation
+     */
+    public function getFields(): array
+    {
+        $this->initializeFields();
+
+        foreach ($this->fields as $name => $field) {
+            if (! ($field instanceof UnresolvedFieldDefinition)) {
+                continue;
+            }
+
+            $this->fields[$name] = $field->resolve();
+        }
+
+        return $this->fields;
+    }
+
+    /**
+     * @return string[]
+     *
+     * @throws InvariantViolation
+     */
+    public function getFieldNames(): array
+    {
+        $this->initializeFields();
+
+        return array_keys($this->fields);
+    }
+
+    protected function initializeFields(): void
+    {
+        if ($this->fields !== null) {
+            return;
+        }
+
+        $fields       = $this->config['fields'] ?? [];
+        $this->fields = FieldDefinition::defineFieldMap($this, $fields);
+    }
+}
--- a/src/Type/Definition/HasFieldsTypeImpl.php
+++ b/src/Type/Definition/HasFieldsTypeImpl.php
@@ -14,9 +14,9 @@ trait HasFieldsTypeImpl
     /**
      * Lazily initialized.
      *
-     * @var FieldDefinition[]|null
+     * @var array<string, FieldDefinition>|null
      */
-    private $fields;
+    private ?array $fields;
 
     /**
      * @throws InvariantViolation
@@ -51,7 +51,7 @@ trait HasFieldsTypeImpl
     }
 
     /**
-     * @return FieldDefinition[]
+     * @return array<string, FieldDefinition>
      *
      * @throws InvariantViolation
      */
@@ -84,7 +84,7 @@ trait HasFieldsTypeImpl
 
     protected function initializeFields(): void
     {
-        if ($this->fields !== null) {
+        if (isset($this->fields)) {
             return;
         }
 
--- a/src/Type/Definition/HasFieldsTypeImpl.php
+++ b/src/Type/Definition/HasFieldsTypeImpl.php
@@ -14,9 +14,9 @@ trait HasFieldsTypeImpl
     /**
      * Lazily initialized.
      *
-     * @var array<string, FieldDefinition>|null
+     * @var array<string, FieldDefinition>
      */
-    private ?array $fields;
+    private array $fields;
 
     /**
      * @throws InvariantViolation
@@ -71,7 +71,7 @@ trait HasFieldsTypeImpl
     }
 
     /**
-     * @return string[]
+     * @return array<int, string>
      *
      * @throws InvariantViolation
      */
--- a/src/Type/Definition/TypeWithFields.php
+++ b/src/Type/Definition/TypeWithFields.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace GraphQL\Type\Definition;
 
-use GraphQL\Error\InvariantViolation;
 use GraphQL\Utils\Utils;
 
 use function array_keys;
@@ -28,9 +27,6 @@ abstract class TypeWithFields extends Type implements HasFieldsType
         $this->fields = FieldDefinition::defineFieldMap($this, $fields);
     }
 
-    /**
-     * @throws InvariantViolation
-     */
     public function getField(string $name): FieldDefinition
     {
         Utils::invariant($this->hasField($name), 'Field "%s" is not defined for type "%s"', $name, $this->name);
@@ -60,11 +56,7 @@ abstract class TypeWithFields extends Type implements HasFieldsType
         return isset($this->fields[$name]);
     }
 
-    /**
-     * @return array<string, FieldDefinition>
-     *
-     * @throws InvariantViolation
-     */
+    /** @inheritDoc */
     public function getFields(): array
     {
         $this->initializeFields();
@@ -80,11 +72,7 @@ abstract class TypeWithFields extends Type implements HasFieldsType
         return $this->fields;
     }
 
-    /**
-     * @return array<int, string>
-     *
-     * @throws InvariantViolation
-     */
+    /** @inheritDoc */
     public function getFieldNames(): array
     {
         $this->initializeFields();
--- a/src/Type/Definition/TypeWithFields.php
+++ b/src/Type/Definition/TypeWithFields.php
@@ -9,7 +9,7 @@ use GraphQL\Utils\Utils;
 
 use function array_keys;
 
-trait HasFieldsTypeImpl
+abstract class TypeWithFields extends Type implements HasFieldsType
 {
     /**
      * Lazily initialized.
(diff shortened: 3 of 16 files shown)


src/Type/Definition/UnresolvedFieldDefinition.php
```diff
@@ -0,0 +1,57 @@
...
+
+        if (! is_array($field)) {
+            return FieldDefinition::create(['name' => $this->name, 'type' => $field]);
+        }
+
+        if (! isset($field['name'])) {
+            $field['name'] = $this->name;
+        } elseif ($field['name'] !== $this->name) {
+            throw new InvariantViolation(
+                sprintf('%s.%s should not dynamically change it\'s name when resolved lazily.', $this->type->name, $this->name)
```
shmax, line 45: it's > its

src/Type/Definition/HasFieldsType.php
```diff
@@ -0,0 +1,33 @@
...
+     * @throws InvariantViolation
+     */
+    public function getField(string $name): FieldDefinition;
+
+    public function hasField(string $name): bool;
+
+    public function tryGetField(string $name): ?FieldDefinition;
+
+    /**
+     * @return FieldDefinition[]
```
spawnia, line 21: List or map? Use `array<?, FieldDefinition` accordingly.

src/Type/Definition/HasFieldsTypeImpl.php
```diff
@@ -0,0 +1,94 @@
...
+use function array_keys;
+
+trait HasFieldsTypeImpl
+{
+    /**
+     * Lazily initialized.
+     *
+     * @var FieldDefinition[]|null
+     */
+    private $fields;
```
spawnia, line 19:
```suggestion
    private array $fields;
```

src/Type/Definition/HasFieldsTypeImpl.php
```diff
@@ -0,0 +1,94 @@
...
+    public function getFieldNames(): array
+    {
+        $this->initializeFields();
+
+        return array_keys($this->fields);
+    }
+
+    protected function initializeFields(): void
+    {
+        if ($this->fields !== null) {
```
spawnia, line 87:
```suggestion
        if (isset($this->fields)) {
```

spawnia (changes requested):
Annotations like `T[]` should be replaced by either `array<int, T>` or `array<string, T>`.

Can you add a benchmark to highlight the benefits?

stayallive:
I'd like to add a benchmark, but it depends on the implementation of the fields resolver and I'm having a hard time coming up with a good benchmark that compares lazy vs not-lazy programatically.

There are also no benchmarks for the other lazy approaches (like for lazy types for example) so if you have an idea on how to properly and synthetically add a useful benchmark, let me know.

I have quite some blackfire.io traces I can show on how this improves real world performance (when using Lighthouse) on a large schema where the fields do quite some work in when generating the `FieldDefinition`.

src/Type/Definition/HasFieldsTypeImpl.php
```diff
@@ -0,0 +1,94 @@
...
+use function array_keys;
+
+trait HasFieldsTypeImpl
+{
+    /**
+     * Lazily initialized.
+     *
+     * @var array<string, FieldDefinition>|null
+     */
+    private ?array $fields;
```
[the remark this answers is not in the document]
stayallive: Shouldn't the type also include the default which is `null` because it's a possible type even though you would never explicitly assign it that type?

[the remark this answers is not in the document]
spawnia: No. There is no default, the field is uninitialized first.

src/Type/Definition/HasFieldsType.php
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * @return array<string, FieldDefinition>
+     *
+     * @throws InvariantViolation
+     */
+    public function getFields(): array;
+
+    /**
+     * @return string[]
```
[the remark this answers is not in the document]
stayallive: Thanks 👍

src/Type/Definition/UnresolvedFieldDefinition.php
```diff
@@ -0,0 +1,70 @@
...
+
+    private string $name;
+
+    /** @var (callable(): FieldDefinition|array<string, mixed>|Type) $resolver */
+    private $resolver;
+
+    /**
+     * @param (callable(): FieldDefinition|array<string, mixed>|Type) $resolver
+     */
+    public function __construct(Type $type, string $name, callable $resolver)
```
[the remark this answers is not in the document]
stayallive: Dang... yep that was it... should have seen that :D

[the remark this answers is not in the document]
stayallive: Hehe, should have figured that out... good call anyway, callable makes more sense instead of just a `Closure`!

src/Type/SchemaValidationContext.php
```diff
@@ -440,7 +440,7 @@ private function validateFields(Type $type): void
         $fieldMap = $type->getFields();
 
         // Objects and Interfaces both must define one or more fields.
-        if (! $fieldMap) {
+        if (count($fieldMap) === 0) {
```
simPod, line 443: We can use

```suggestion
        if ($fieldMap === []) {
```

src/Type/Definition/HasFieldsTypeImpl.php
```diff
@@ -0,0 +1,94 @@
...
+declare(strict_types=1);
+
+namespace GraphQL\Type\Definition;
+
+use GraphQL\Error\InvariantViolation;
+use GraphQL\Utils\Utils;
+
+use function array_keys;
+
+trait HasFieldsTypeImpl
```
simPod, line 12: Personally I'd rather design it as 

`class InterfaceType extends TypeWithFields`

`abstract class TypeWithFields extends Type implements HasFieldsType`

1. Traits often cause issues for static analyser tools and IDEs.
2. Also it would allow us to use a better name than `Impl`.
3. We could safely use `@inheritDoc` instead of duplicating phpdoc.

Though it's also a matter of preference I guess.

src/Type/Definition/HasFieldsTypeImpl.php
```diff
@@ -0,0 +1,94 @@
...
+     * @throws InvariantViolation
+     */
+    public function getField(string $name): FieldDefinition
+    {
+        Utils::invariant($this->hasField($name), 'Field "%s" is not defined for type "%s"', $name, $this->name);
+
+        return $this->tryGetField($name);
+    }
+
+    public function tryGetField(string $name): ?FieldDefinition
```
simPod, line 31: what about using `get` for `FieldDefinition` and `find` for `FieldDefinition|null`?

```suggestion
    public function findField(string $name): ?FieldDefinition
```

src/Type/Definition/HasFieldsTypeImpl.php
```diff
@@ -0,0 +1,94 @@
...
+     * @throws InvariantViolation
+     */
+    public function getFieldNames(): array
+    {
+        $this->initializeFields();
+
+        return array_keys($this->fields);
+    }
+
+    protected function initializeFields(): void
```
simPod, line 85: this can be eventually private, no?

src/Type/Definition/HasFieldsTypeImpl.php
stayallive: I'm fine with this, it's a preference thing indeed, I'll make the adjustments 👍

src/Type/Definition/HasFieldsTypeImpl.php
stayallive: Sounds good to me!

src/Type/Definition/TypeWithFields.php
```diff
@@ -0,0 +1,94 @@
...
+        $this->initializeFields();
+
+        return isset($this->fields[$name]);
+    }
+
+    /**
+     * @return array<string, FieldDefinition>
+     *
+     * @throws InvariantViolation
+     */
```
simPod, line 67: I'd use this 👇🏾  where comments are useful but are in interface already

```suggestion
    /** @inheritDoc */
```

src/Utils/TypeInfo.php
```diff
@@ -204,8 +205,8 @@ public static function extractTypes($type, ?array $typeMap = null)
             $nestedTypes = array_merge($nestedTypes, $type->getInterfaces());
         }
 
-        if ($type instanceof ObjectType || $type instanceof InterfaceType) {
-            foreach ($type->getFields() as $fieldName => $field) {
+        if ($type instanceof HasFieldsType) {
```
simPod, line 208: This popped out nicely 👍🏾

simPod (approved):

src/Utils/TypeInfo.php
stayallive: Yeah that is arguably why the interface is the better choice here!

src/Type/Definition/TypeWithFields.php
stayallive: Ah yeah, interestingly with this docblock:

```
    /**
     * @throws InvariantViolation
     */
```

`phpcs` removes the `@inhertiDoc` (it leaves it where there is also an `@return`)... that is the `@inheritDoc` even needed?

simPod: No. It's needed only for array types IIRC. Otherwise it's considered to be already inherited automatically from the interface without the need to specify the inheritance.

spawnia (approved):
Looks good to me. Is there any chance of breakage when backporting this?

stayallive:
I've looked through the PR again, but I believe only some code was moved to introduce a new abstract type but the method names were not changed so all public code remains unchanged except for added methods and of course a field definition now being able to be a callable. So I think this can be backported safely.

spawnia:
Alright. Can you add a PR for the backport? I would like to try a slight adjustment to our workflow: merge the backport first, document and release the change, then incorporate the changelog entry in this PR before merging it.

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v14.7.0

stayallive:
Thanks everyone for the review and support for this feature, this gonna make apps fly 🚀

sebastienbarre:
@stayallive This looks great and I'm trying to evaluate if we should start declaring all our types using callables. Could you give us some real world example as to how this helped you? Typically what really takes time for us is resolving the *value* of the field (since this often involves hitting some kind of database or backend), not its type definition. What kind of field *definition* would be significantly less performant than your typical definition? Thanks.

stayallive:
Hey @sebastienbarre, it depends (of course) but I think this from the description describes it the best for me / our situation the best:

> Within Laravel Lighthouse for example this has huge performance improvements (especially on larger schema's) because the process of resolving a field can be quite some work and there is a potential huge cost to be paid for each field. Allowing them to be loaded just in time solves that by paying the cost only when needed instead of always on every query.

So when implementing a raw schema in just simply this package and some sauce there might not be any improvement to be made. But in case you do work or resolve things from a container etc. this can shave off a lot of busy work the schema is doing for no reason when a field is not requested in the query.

Hope that helps?

sebastienbarre:
Thanks. So maybe specific fields are not the culprits, but it's more the sheer size of your schema. Our is getting pretty large too, hence my curiosity. I'll have to try to benchmark. Which tool did you use to benchmark this specific improvement, if I may ask?

stayallive:
I had a hard time benchmarking it since any synthetic benchmark is pretty fast and setting up a realistic benchmark just took too much time. So I used the biggest schema I had available and ran queries with and without this change a few hundred times. Not very scientific but good enough for me since it was a clear improvement. 

After putting it in production the results were also pretty good (as measured by Sentry). 

image

It's a little hard to tell but simple requests went from 200-300ms to 100-150ms or even lower. Bottleneck was shifting towards the queries instead of processing done before the query could be executed (building the schema).

In the end profiling locally with Blackfire set me on this path & change. After tinkering around this took away the biggest chunk of time in the profile. 

But again, this is highly specific to Laravel Lighthouse and I have no clue if this helps in other situations too 🙂

sebastienbarre:
Out of curiosity, were you already using the feature that allows `fields` itself to be a callable? As opposed to each field being a callable, which is the feature added in this PR.

stayallive:
I know because I wrote the PR 😜

And yes, field values were already lazily resolved. 

The only change was this PR and the changes required for Lighthouse to use it: https://github.com/nuwave/lighthouse/pull/1851.

sebastienbarre:
Thanks. 
Only one way to find out :)
I started converting our schema, and I'm hitting some snags tho.

Are input types not supported?


```php
class SignInCredentialsInputType extends InputObjectType
{
  public function __construct()
  {
    $types = TypeRegistry::getRegistry();
    $config = [
      'description' => 'Employee credentials',
      'fields' => static fn () => [
        'email' => static fn () => [
          'type' => $types->nonNull($types->string()),
          'description' => 'Employee email',
        ],
        'password' => static fn () => [
          'type' => $types->nonNull($types->string()),
          'description' => 'Employee password',
        ],
      ],
    ];
    parent::__construct($config);
  }
}
```

The code above used to be:
```php
        'email' => [
          'type' => $types->nonNull($types->string()),
          'description' => 'Employee email',
        ],
        'password' => [
          'type' => $types->nonNull($types->string()),
          'description' => 'Employee password',
        ],
```

With the callables in place, I'm now getting an error using v14.9.0:
```
TypeError: Return value of GraphQL\Type\Schema::resolveType() must be an instance of GraphQL\Type\Definition\Type, array returned

Vendor/webonyx/graphql-php/src/Type/Schema.php:406
Vendor/webonyx/graphql-php/src/Type/Definition/InputObjectField.php:121
Vendor/webonyx/graphql-php/src/Validator/Rules/ValuesOfCorrectType.php:107
Vendor/webonyx/graphql-php/src/Language/Visitor.php:414
Vendor/webonyx/graphql-php/src/Language/Visitor.php:470
Vendor/webonyx/graphql-php/src/Language/Visitor.php:277
Vendor/webonyx/graphql-php/src/Validator/DocumentValidator.php:224
Vendor/webonyx/graphql-php/src/Validator/DocumentValidator.php:116
```

stayallive:
I think you went a little too far, IIRC it should be something like this:

```php
class SignInCredentialsInputType extends InputObjectType
{
  public function __construct()
  {
    $types = TypeRegistry::getRegistry();
    $config = [
      'description' => 'Employee credentials',
      'fields' => static fn () => [
        'email' => [
          'type' => $types->nonNull($types->string()),
          'description' => 'Employee email',
        ],
        'password' => [
          'type' => $types->nonNull($types->string()),
          'description' => 'Employee password',
        ],
      ],
    ];
    parent::__construct($config);
  }
}
```

Notice that the fields array can be an array or closure but the fields themself should be an array.

The `fields` are lazy loaded for a type not the individual `field`s within.

By the way looking at the simplicity of your definition I'm really interested too see if you book any performance results :) My guess is that it won't be slower but also won't be (measurable) faster. But maybe a bit better on the memory pressure.

sebastienbarre:
OK now I'm very confused. 

That's why I asked earlier, " Out of curiosity, were you already using the feature that allows `fields` itself to be a callable? " (as opposed to individual fields). 

All my `fields` have been callable for a long time now, at least a year ago according to git, way before that PR. This has been doable for a while, unless I'm completely misunderstanding when this PR was created?

Furthermore, I'm looking at the code in this PR, and the tests seem to imply just that, that the individual fields themselves can callables. See [https://github.com/webonyx/graphql-php/pull/861/files#diff-7bbb5376d72f8561023190dadb349c5acf905edf96ef2763b7164887f44fe7a6R1891](here) for example:
```php
objType = new ObjectType([
            'name'   => 'SomeObject',
            'fields' => [
                'f' => static function (): array {
                    return ['type' => Type::string()];
                },
            ],
        ]);
```
`fields` here is returning an array, but the field itself, `f`, is using a callable, which is returning an array with the `type`. This is the same, unless I'm misreading, as my `email` field which used to return an array with the `type`, and now uses a callable that returns that very array. 

Am I misunderstanding this PR completely? Because making the field itself a callable seems to work fine for me in some examples I just tried, based on your PR. It's for input object that it seems to be having troubles.

For example, this below works fine, I have been `curl`-ing at it all morning to benchmark:
```php
      'fields' => fn () => [
        'ping' => static fn () => [
          'type' => $types->string(),
          'description' => 'Ping the service (test query, no authentication required)',
          'resolve' => static fn () => 'pong',
        ],
```
note how `'ping'` is a callable. It used to be just an array before this morning:
```php
      'fields' => fn () => [
        // -----------------------------------------------------------------------------------
        'ping' => [
          'type' => $types->string(),
```

stayallive:
Whoops now I'm confused too, handy that everything is called field(s) and I think you are right and I was to hastily with my reply :)

It's possible that for `InputObjectType` there is something broken then maybe (maybe for others too)... I do not have the code handy but maybe it's a great idea to open up a new issue with this finding and or PR with a failing test case if you are willing?

I'd have to take a look at it since it's been a few months since I was this deep in the core but from a quick browse around it might be very true that this was just implemented for object (and interface) types...

sebastienbarre:
Thanks, this is done in #911, with a failing test case.

PR #866, opened 2021-05-17
[v14] Lazy field definitions (#861)

stayallive:
As requested by @spawnia, the backported changes of #861 to v14.


spawnia (approved):
Looks like a very nice and safe addition to this library. Having this in v14 is nice, it allows us to test it in practice and perhaps make changes before releasing v15.

PR #867, opened 2021-05-20
 change cache file content

3alisaki:
i'm changed the cache content from standard php parsable files to unserializeable php value


spawnia:
I reckon the PHP approach is faster, it allows the OPCache to optimize the operation. Why did you change it?

3alisaki:
I did not know it was faster, I made this change because I feel it is safer from errors.

PR #869, opened 2021-05-27
[FEATURE] Added status code in error formatter

lewisvoncken:


src/Error/FormattedError.php
```diff
@@ -216,6 +216,7 @@ static function (SourceLocation $loc): array {
             $formattedError = self::addDebugEntries($formattedError, $exception, $debug);
         }
 
+        $formattedError['code'] = $e->getCode();
```
spawnia, line 219: This is discouraged by the GraphQL specification, see http://spec.graphql.org/draft/#note-5c13b

src/Error/Error.php
```diff
@@ -100,7 +100,7 @@ public function __construct(
         $previous = null,
         array $extensions = []
     ) {
-        parent::__construct($message, 0, $previous);
+        parent::__construct($message, $previous->getCode(), $previous);
```
spawnia, line 103: `$previous` can be `null`.

spawnia (changes requested):

src/Error/FormattedError.php
simPod: We should not pursue this, right? So it stays protocol agnostic.

spawnia: Well the `code` in this case is an internal PHP error code, not an HTTP status code. But as the spec says, if anywhere it should go under the `extensions` key.

spawnia:
Closing because the reasons I outlined in https://github.com/webonyx/graphql-php/pull/869#discussion_r641364409

PR #871, opened 2021-06-01
Add support for type config decorator in SchemaExtender

### LINKED ISSUE #461: Config Decoration of Types from Extended Schema
@Torsten85 introduced `SchemaExtender` with #362 and #186, allowing to extend an existing schema with additional SDL.

But how do you decorate the config of new types added by an extension?
`BuildSchema::build` accepts a `$typeConfigDecorator`, but `SchemaExtender::extend` does not.

Decorating new types in the original Schema does not work, since they do not exist yet when building the unextended schema.

Example:
```php
$sdl = '
  type Query {
    defaultValue: String
  }
  type Foo {
    value: Int
  }
';

$typeConfigDecorator = function ($typeConfig) {
  switch ($typeConfig['name']) {
    case 'Foo': // works as expected
        $typeConfig['resolveField'] => function () { /* whatever */ };
        break;
    case 'Bar': // Never the case as this type does not exist 
                // in the original unextended Schema
        $typeConfig['resolveField'] => function () { /* whatever */ };
        break;
  }

  return $typeConfig;
};

$documentNode = \GraphQL\Language\Parser::parse($sdl);
$schema = \GraphQL\Utils\BuildSchema::build($documentNode, $typeConfigDecorator);

// how to decorate the config of type `Bar`?
$extensionSdl = '
  type Bar {
    foo: Foo
  }
';

$extendedDocumentNode = \GraphQL\Language\Parser::parse($extensionSdl);
$extendedSchema = \GraphQL\Utils\SchemaExtender::extend($schema, $extendedDocumentNode);
```

The way I see it it's not possible to pass a configDecorator for newly introduced types to the `SchemaExtender` or am I missing something?

AdrienPoupa:
This adds support for the type config decorator when merging GraphQL schema files.

I added the config decorator option at the end to keep retro compatibility even though it should be before options to remain consistent with the rest of the codebase.

### DIFF
--- a/src/Utils/SchemaExtender.php
+++ b/src/Utils/SchemaExtender.php
@@ -506,8 +506,12 @@ class SchemaExtender
     /**
      * @param array<string, bool> $options
      */
-    public static function extend(Schema $schema, DocumentNode $documentAST, array $options = []): Schema
-    {
+    public static function extend(
+        Schema $schema,
+        DocumentNode $documentAST,
+        array $options = [],
+        ?callable $typeConfigDecorator = null
+    ): Schema {
         if (! (isset($options['assumeValid']) || isset($options['assumeValidSDL']))) {
             DocumentValidator::assertValidSDLExtension($documentAST, $schema);
         }
@@ -588,7 +592,8 @@ class SchemaExtender
                 }
 
                 throw new Error('Unknown type: "' . $typeName . '". Ensure that this type exists either in the original schema, or is added in a type definition.', [$typeName]);
-            }
+            },
+            $typeConfigDecorator
         );
 
         static::$extendTypeCache = [];
--- a/tests/Utils/SchemaExtenderTest.php
+++ b/tests/Utils/SchemaExtenderTest.php
@@ -2039,4 +2039,55 @@ extend type Query {
 
         static::assertEquals($this->dedent($expected), SchemaPrinter::doPrint($extendedSchema));
     }
+
+    public function testSupportsTypeConfigDecorator()
+    {
+        $queryType = new ObjectType([
+            'name' => 'Query',
+            'fields' => [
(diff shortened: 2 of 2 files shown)


vhenzl:
Hey @AdrienPoupa, thanks for this! I just want to check that this is the intended behaviour:

The type config decorator in `SchemaExtender::extend()` is applied only to the extending types. If I want to decorate also types from the base schema, I have to do that in `BuildSchema::build()`:

```php
$baseDocument = Parser::parse($baseGQL);
$baseSchema = BuildSchema::build($baseDocument, $baseTypeConfigDecorator); // <-- one decorator here

$extensionDocument = Parser::parse($extensionGQL);
$schema = SchemaExtender::extend($baseSchema, $extensionDocument, [], $extensionTypeConfigDecorator); // <-- another decorator here
```

Is it intentional? Would it be (technically) possible to update the extender to decorate the whole resulting schema so I don't have to keep track of from which particular document each type comes?

```php
$schema = BuildSchema::build(Parser::parse($gql1));
$schema = SchemaExtender::extend($schema, $gql2); 
$schema = SchemaExtender::extend($schema, $gql3);
// ...
$schema = SchemaExtender::extend($baseSchema, $gqlN, [], $typeConfigDecorator); // <-- only decorator here for everything
```

vhenzl:
I believe the current implementation isn't correct, I wrote more about it in https://github.com/webonyx/graphql-php/discussions/921.

AdrienPoupa:
Indeed I did not test this case so it may not work.

However I think it makes sense to see the type config you inject as tied to the schema extension, at least that was my use case.

Given a new part of the schema, inject this additional type config linked to this schema. Feel free to modify the current mechanism if it does not suit you :)

Issue #552, opened 2019-10-07
Custom directives get lost during schema printing


franzwilding:
Hey!

I'm currently playing around with a custom schema builder, that allows registering classes that extend the schema and eventually create a cacheable DocumentNode. It is working pretty well, my current solution looks like this: 

```php
    /**
     * @return DocumentNode
     * @throws \GraphQL\Error\Error
     * @throws \GraphQL\Error\SyntaxError
     */
    public function buildCacheableSchema() : DocumentNode {

        $schema = BuildSchema::build(self::BASE_SCHEMA);

        foreach($this->extenders as $extender) {
            if($extension = $extender->extend($schema)) {
                $schema = SchemaExtender::extend($schema, Parser::parse($extender->extend($schema)));
            }
        }

        return Parser::parse(SchemaPrinter::doPrint($schema));
    }
```

However, I am facing the problem, that custom directives get lost during doPrint, because of the implementation of [SchemaPrinter::printField](https://github.com/webonyx/graphql-php/blob/master/src/Utils/SchemaPrinter.php#L375) which only includes deprecated directive .

Now my question consists of two parts: 

1. Is there a better solution to get the DocumentNode from a Schema, besides doing parse(doPrint(schema))? 

2. And if this actually is the best solution to do this, how do I get my custom directives into the final schema?

Thank you!

vladar:
This topic is a bit controversial. See https://github.com/graphql/graphql-js/issues/2020 for some details. We will adopt the same solution as graphql-js eventually. So I will keep this issue open until it is resolved in the reference implementation.

renepardon:
Hello,

I'm building the federation implementation into lighthouse right now. https://github.com/nuwave/lighthouse/pull/1051

The problem I'm facing here is that I also need custom directives printed into the Schema (For the `_service { sdl }` request and we wouldn't like to implemented this through a custom printer into lighthouse, because for this to work we would need to overwrite/duplicate a lot of code. from `webonyx/graphql-php`. Reference request is #515 which is already closed but this issue (#552) looks  promising.

What do you suggest on how to continue?


@spawnia 
@vladar

obi1kenobi:
I believe https://github.com/graphql/graphql-js/issues/2389 is the directly-equivalent issue of this one. Hoping GraphQL.js can address it soon so all the ports can follow suit!

spawnia:
@vladar I agree to wait what the reference implementation does for an official solution.

[...]

PR #875, opened 2021-06-23
Support custom directives in SchemaPrinter

jbtbnl:


spawnia:
See https://github.com/webonyx/graphql-php/issues/552 and https://github.com/graphql/graphql-js/issues/2020 for discussion.

jbtbnl:
@spawnia thanks for pointing out related discussions. Whether this PR is the way to go or not, it hopefully at least contributes to keeping this to keep this to the attention.

PR #730, opened 2020-10-22
propagate PHP errors

s3tezsky:
When I use `serialize` on custom scalar type with an type error inside (wrong number of arguments or wrong argument type - `\ArgumentCountError` or `\TypeError`) the original error message is supressed and changed to hardcoded one (`Expected a value of type * but received: instance of *`). Much better for developing or debugging would be to know the real error.

e.g.:

```php
// in custom scalar type
public static serialize(): string
{
    return static::customScalarTypeSerialize(/*any args*/);
}

public static function customScalarTypeSerialize(/*any args*/): string
{
    return number_format('string value'); // first argument must be float when defined declare(strict_types=1), otherwise \TypeError is thrown
}
```

Maybe this is the only one of many similar troubles.

---

I am not really sure whenever my solutions is OK but this issue really borthers me. If you give me some hints about better solution or how this change should be tested I will try to give it more time to get this done.

PR #876, opened 2021-06-23
Propagate error message for why leaf value serialization failed

spawnia:
Supersedes https://github.com/webonyx/graphql-php/pull/730

### DIFF
--- a/src/Type/Definition/EnumType.php
+++ b/src/Type/Definition/EnumType.php
@@ -147,7 +147,7 @@ class EnumType extends Type implements InputType, OutputType, LeafType, Nullable
             return $lookup[$value]->name;
         }
 
-        throw new Error('Cannot serialize value as enum: ' . Utils::printSafe($value));
+        throw new InvariantViolation('Cannot serialize value as enum: ' . Utils::printSafe($value));
     }
 
     /**
--- a/src/Utils/AST.php
+++ b/src/Utils/AST.php
@@ -245,7 +245,7 @@ class AST
             try {
                 $serialized = $type->serialize($value);
             } catch (Throwable $error) {
-                if ($error instanceof Error && $type instanceof EnumType) {
+                if ($error instanceof InvariantViolation && $type instanceof EnumType) {
                     return null;
                 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 #### Unreleased
 Changed:
 - PHP version required: 7.4+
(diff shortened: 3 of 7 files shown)


spawnia:
The stack trace of the original error should be preserved if possible, but simply adding https://github.com/webonyx/graphql-php/pull/730#issuecomment-724230511 seems to prevent the `debugMessage` from being displayed at all.

simPod (commented):
I don't feel competent to review this but https://github.com/webonyx/graphql-php/pull/730#issuecomment-724230511 seems like a more valid approach.

> seems to prevent the debugMessage from being displayed at all.

did you have a chance to investigate a reason?

spawnia:
This PR lacks an improved stacktrace, but I am fairly confident that combining the error messages is a great idea. The wrapping InvariantViolation provides additional context from a client perspective, but does not aid the server developer that much in finding the underlying cause.

spawnia:
> did you have a chance to investigate a reason?

Got it.

src/Utils/AST.php
```diff
@@ -245,7 +245,7 @@ public static function astFromValue($value, InputType $type)
             try {
                 $serialized = $type->serialize($value);
             } catch (Throwable $error) {
-                if ($error instanceof Error && $type instanceof EnumType) {
+                if ($error instanceof InvariantViolation && $type instanceof EnumType) {
```
spawnia, line 248: Why is this supposed to be ignored?

src/Type/Definition/EnumType.php
```diff
@@ -147,7 +147,7 @@ public function serialize($value)
             return $lookup[$value]->name;
         }
 
-        throw new Error('Cannot serialize value as enum: ' . Utils::printSafe($value));
+        throw new InvariantViolation('Cannot serialize value as enum: ' . Utils::printSafe($value));
```
spawnia, line 150: Just like in scalar types, when serializing a faulty internal value, it can be unsafe to show it in the result. The client won't be able to do much about it anyways.

simPod:
I'm not sure about joining those messages as the original one should be available in $previous in any stacktrace explorer, right? But I fancy the rest 👍🏾.

spawnia:
> I'm not sure about joining those messages as the original one should be available in $previous in any stacktrace explorer, right?

The client would have no way to see them if they are not in `debugMessage`. The trace within the response has no such data.

simPod (approved):
I just tried it and it makes sense I guess

PR #880, opened 2021-07-02
Fix examples/01-blog - fix wrong type for ID at findStories

urbanc:
This is just a quick fix(to have a working example) with a cast into int type at findStories() function like is already at https://github.com/webonyx/graphql-php/blob/bf110702e4e74ef0c7afbf255c914f22cc0ed199/examples/01-blog/Blog/Type/QueryType.php#L78

```
"debugMessage": "Argument 2 passed to GraphQL\\Examples\\Blog\\Data\\DataSource::findStories() must be of the type int or null, string given, called in .../graphql-php/graphql-php/examples/01-blog/Blog/Type/QueryType.php on line 98",
```

To reproduce this error run:
```
{
  viewer {
    id
    email
  }
  user(id: 2) {
    id
    email
  }
  stories(after: 1) {
    id
    body
    comments {
      ...CommentView
    }
  }
  lastStoryPosted {
    id
    hasViewerLiked
    author {
      id
      photo(size: ICON) {
        id
        url
        type
        size
        width
        height
        # Uncomment following line to see validation error:
        # nonExistingField
        
        # Uncomment to see error reporting for fields with exceptions thrown in resolvers
        # fieldWithError
        # nonNullFieldWithError
      }
      lastStoryPosted {
        id
      }
    }
    body(format: HTML, maxLength: 10)
  }
}

fragment CommentView on Comment {
  id
  body
  totalReplyCount
  replies {
    id
    body
  }
}
```

### DIFF
--- a/examples/01-blog/Blog/Data/DataSource.php
+++ b/examples/01-blog/Blog/Data/DataSource.php
@@ -181,7 +181,7 @@ class DataSource
     /**
      * @return array<int, Story>
      */
-    public static function findStories(int $limit, ?int $afterId = null): array
+    public static function findStories(int $limit, $afterId = null): array
     {
         $start = $afterId !== null
             ? (int) array_search($afterId, array_keys(self::$stories), true) + 1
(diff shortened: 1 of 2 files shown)


spawnia:
I would rather keep it strictly typed so I prepared another fix.

PR #882, opened 2021-07-03
Fix examples/03-standard-server - Notice: Trying to access array offset on value of type null

urbanc:
Removed $rootValue['prefix'] variable from field "echo" resolve function because $rootValue is always null.

```
"text": "<br />\n<b>Notice</b>:  Trying to access array offset on value of type null in <b>.../examples/03-standard-server/graphql.php</b> on line <b>30</b><br />\n{\"data\":{\"echo\":\"Hello world\"}}"
```

To reproduce this notice message run:
```
curl -d '{"query": "query { echo(message: \"Hello World\") }" }' -H "Content-Type: application/json" http://localhost:8080
```

### DIFF
--- a/examples/03-standard-server/graphql.php
+++ b/examples/03-standard-server/graphql.php
@@ -11,6 +11,7 @@ declare(strict_types=1);
 
 require_once __DIR__ . '/../../vendor/autoload.php';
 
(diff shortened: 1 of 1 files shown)


spawnia:
I will include the `$rootValue` in this example.

PR #883, opened 2021-07-03
Fix examples/02-schema - schema is not configured for mutations

urbanc:
Fixed wrong schema definition for mutations.

To reproduce this error run:
```
curl -d '{"query": "mutation { sum(x: 2, y: 2) }" }' -H "Content-Type: application/json" http://localhost:8080
```


spawnia (approved):

spawnia:
Thank you

Issue #888, opened 2021-07-12
Return error on PHP 8.0.8 but false on 7.4 in src/utils


romainpltr:
Hi, 

When i use bundle on PHP 8.0.8, i've an error because the new array_key_exist function return an error and no a boolean like in PHP 7.4

I've found the issue, the problem are on the file [src/utils/Value.php](https://github.com/webonyx/graphql-php/blob/master/src/Utils/Value.php) line 170 but some other line have also the same problem, you need to check if is an array or not, because if not, the new array key exist function return an error.

![image-1](https://user-images.githubusercontent.com/14335973/125297440-86d0e980-e327-11eb-8a10-c9da7d700cd9.png)
![image](https://user-images.githubusercontent.com/14335973/125298941-ed0a3c00-e328-11eb-988b-db6954c5f2cf.png)

spawnia:
Thanks for the report, we can definitely make the library safer here and produce a better error.

What you are doing in your specific use case seems like it would not work anyways. Deducting from your screenshot and the code path you hit, you seem to be passing an `UploadedFile` instance in a place where an input object is expected. I have only ever seen uploaded files represented by scalar types in the schema. That way, you can control how the value is coerced/parsed yourself.

Simply casting an `UploadedFile` to an array seems to lead to a nonsensical result which would throw further down the line:

```
>>> $f = new \Symfony\Component\HttpFoundation\File\UploadedFile('.env', 'bar')
=> Symfony\Component\HttpFoundation\File\UploadedFile {#4348
     path: "",
     filename: ".env",
     basename: ".env",
     pathname: ".env",
     extension: "env",
     realPath: "/var/www/.env",
     aTime: 2021-07-13 11:49:42,
     mTime: 2021-07-12 11:10:20,
     cTime: 2021-07-12 11:10:20,
     inode: 433132,
     size: 1414,
     perms: 0100644,
     owner: 1000,
     group: 1000,
     type: "file",
     writable: true,
     readable: true,
     executable: false,
     file: true,
     dir: false,
     link: false,
   }
>>> (array) $f
=> [
     "\0Symfony\Component\HttpFoundation\File\UploadedFile\0test" => false,
     "\0Symfony\Component\HttpFoundation\File\UploadedFile\0originalName" => "bar",
     "\0Symfony\Component\HttpFoundation\File\UploadedFile\0mimeType" => "application/octet-stream",
     "\0Symfony\Component\HttpFoundation\File\UploadedFile\0error" => 0,
[...]

PR #889, opened 2021-07-13
Fix coercion of arbitrary objects to input objects

spawnia:
Resolves https://github.com/webonyx/graphql-php/issues/888

@romainpltr please validate this fixes your issue.

### DIFF
--- a/src/Utils/Value.php
+++ b/src/Utils/Value.php
@@ -140,7 +140,11 @@ class Value
         }
 
         if ($type instanceof InputObjectType) {
-            if (! is_array($value) && ! $value instanceof stdClass) {
+            // Cast objects to associative array before checking the fields.
+            // Note that the coerced value will be an array.
+            if ($value instanceof stdClass) {
+                $value = (array) $value;
+            } elseif (! is_array($value)) {
                 return self::ofErrors([
                     self::coercionError(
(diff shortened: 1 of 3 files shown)


tests/Utils/CoerceValueTest.php
```diff
@@ -396,12 +396,29 @@ public function testReturnsErrorForNonObjectType(): void
...
 
+    /**
+     * @return iterable<int, array{object}>
+     */
+    public function coercibleObjects(): iterable
+    {
+        yield [(object) ['foo' => 123]];
+
+        yield [
+            new class {
```
simPod, line 416: do we want to support casting classes to array as well? 🤔

simPod: Maybe just keep supporting only stdClass. Casting other classes to array (or even iterating over it) is rather an abomination I'd say.

simPod (approved):

src/Utils/Value.php
```diff
@@ -140,7 +140,11 @@ static function ($enumValue): string {
         }
 
         if ($type instanceof InputObjectType) {
-            if (! is_array($value) && ! $value instanceof stdClass) {
+            // Cast objects to associative array before checking the fields.
```
simPod, line 143: I'd put this comment inside the first if branch

Issue #781, opened 2021-02-25
Broken Query causes Exception


hooby404:
Not sure if this is intended behavior (garbage in - garbage out), but I definitely did not expect it and went searching for a non-existent mistake in my schema definition.

If this is working as intended, just close the issue.

### Steps to reproduce:

Take any, existing query that works, for example:
```
user {
    id
}
```
And then add an argument that is not defined in the schema, and put in an empty array like this:
```
user(
    doesnotexist: []
) {
    id
}
```
### Expected result:

Error Response with message:
```
Unknown argument "doesnotexist" on field "user" of type "Query".
```

### Observed result:

Exception:
```
GraphQL\\Type\\Definition\\Type::getNullableType(): Argument #1 ($type) must be of type GraphQL\\Type\\Definition\\Type, null given, called in /var/www/html/vendor/webonyx/graphql-php/src/Validator/Rules/ValuesOfCorrectType.php on line 75
```

spawnia:
I believe we have a validation rule with a nice error message in place, but tightening the type hints might have lead to premature failure. I think we might have to relax the type constraint or wrap the call in a try/catch to fail gracefully on garbage inputs.

PR #895, opened 2021-07-23
Handle `null` parent of list in `ValuesOfCorrectType::getVisitor`

jdfadams:
This fixes https://github.com/webonyx/graphql-php/issues/781.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 - Avoid QueryPlan crash when multiple $fieldNodes are present
 - Clarify error when attempting to coerce anything but `array` or `stdClass` to an input object
-- Handle `null` in `Type::getNullableType`
+- Handle `null` parent of list in `ValuesOfCorrectType::getVisitor`
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 ### Added
 
 - Improve extendability of validator rules
+- Add tests for errors that occur when undeclared fields are passed in input
 
 ### Optimized
 
@@ -27,6 +28,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 - Avoid QueryPlan crash when multiple $fieldNodes are present
 - Clarify error when attempting to coerce anything but `array` or `stdClass` to an input object
+- Handle `null` in `Type::getNullableType`
 
 ### Removed
 
--- a/src/Type/Definition/Type.php
+++ b/src/Type/Definition/Type.php
@@ -267,7 +267,7 @@ abstract class Type implements JsonSerializable
     /**
      * @api
      */
-    public static function getNullableType(?Type $type): ?Type
+    public static function getNullableType(Type $type): Type
     {
         return $type instanceof NonNull
             ? $type->getWrappedType()
--- a/src/Type/Definition/Type.php
+++ b/src/Type/Definition/Type.php
@@ -267,7 +267,7 @@ abstract class Type implements JsonSerializable
     /**
      * @api
      */
-    public static function getNullableType(Type $type): Type
+    public static function getNullableType(?Type $type): ?Type
     {
         return $type instanceof NonNull
             ? $type->getWrappedType()
--- a/src/Validator/Rules/ValuesOfCorrectType.php
+++ b/src/Validator/Rules/ValuesOfCorrectType.php
@@ -73,7 +73,8 @@ class ValuesOfCorrectType extends ValidationRule
             NodeKind::LST          => function (ListValueNode $node) use ($context, &$fieldName): ?VisitorOperation {
                 // Note: TypeInfo will traverse into a list's item type, so look to the
                 // parent input type to check if it is a list.
-                $type = Type::getNullableType($context->getParentInputType());
+                $parentType = $context->getParentInputType();
+                $type = $parentType === null ? null : Type::getNullableType($parentType);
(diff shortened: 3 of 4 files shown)


tests/UserArgsTest.php
```diff
@@ -0,0 +1,71 @@
...
+use GraphQL\Type\Definition\InputObjectType;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Schema;
+use PHPUnit\Framework\TestCase;
+
+class UserArgsTest extends TestCase
+{
+    /**
+     * @see it('disallows non-existent scalar field on input')
```
spawnia, line 17: Are those test cases from graphql-js?

src/Type/Definition/Type.php
```diff
@@ -267,7 +267,7 @@ public static function assertType($type): Type
     /**
      * @api
      */
-    public static function getNullableType(Type $type): Type
+    public static function getNullableType(?Type $type): ?Type
```
spawnia, line 270: Rather than poisoning this method with nullability, I would rather clean up the problematic call site that does not check for null.

tests/UserArgsTest.php
jdfadams-msp: No, they aren't. They're just tests of behavior one expects. Is the `@see` wrong to use here? 

The first test, with a scalar, just covers behavior wasn't tested elsewhere.

The second test, with the array, resulted in an error:
```TypeError: Argument 1 passed to GraphQL\Type\Definition\Type::getNullableType() must be an instance of GraphQL\Type\Definition\Type, null given, called in /editflow/graphql-php/src/Validator/Rules/ValuesOfCorrectType.php on line 77```
With the code in this pull request, the second test passes.

File: src/Type/Definition/Type.php (hunk 2 above)
Comment (jdfadams-msp) [reply src/Type/Definition/Type.php:270]: I understand. I can clean up the problematic call site instead.

File: tests/UserArgsTest.php (hunk 1 above)
Comment (jdfadams) [reply tests/UserArgsTest.php:17]: No, they aren't. They're just tests of behavior one expects. Is the `@see` wrong to use here? 

The first test, with a scalar, just covers behavior wasn't tested elsewhere.

The second test, with the array, resulted in an error:
```TypeError: Argument 1 passed to GraphQL\Type\Definition\Type::getNullableType() must be an instance of GraphQL\Type\Definition\Type, null given, called in /editflow/graphql-php/src/Validator/Rules/ValuesOfCorrectType.php on line 77```
With the code in this pull request, the second test passes.

File: src/Type/Definition/Type.php (hunk 2 above)
Comment (jdfadams) [reply src/Type/Definition/Type.php:270]: I understand. I'll do that instead.

Comment (jdfadams) [reply src/Type/Definition/Type.php]: Here's the new commit: https://github.com/webonyx/graphql-php/pull/895/commits/c8420c88cc0088755bf4f120f0d7e24df71eae52

File: CHANGELOG.md [hunk 3]
```diff
@@ -27,6 +28,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 - Avoid QueryPlan crash when multiple $fieldNodes are present
 - Clarify error when attempting to coerce anything but `array` or `stdClass` to an input object
+- Handle `null` in `Type::getNullableType`
```
spawnia, line 31: No longer accurate

tests/UserArgsTest.php
spawnia: Content looks good, but as you noted the annotations should go.

CHANGELOG.md
jdfadams-msp: I updated this.

tests/UserArgsTest.php
jdfadams: I removed the annotations and updated the comments and test names.

CHANGELOG.md
jdfadams: I updated this.

src/Validator/Rules/ValuesOfCorrectType.php
```diff
@@ -73,7 +73,8 @@ public function getVisitor(ValidationContext $context)
             NodeKind::LST          => function (ListValueNode $node) use ($context, &$fieldName): ?VisitorOperation {
                 // Note: TypeInfo will traverse into a list's item type, so look to the
                 // parent input type to check if it is a list.
-                $type = Type::getNullableType($context->getParentInputType());
+                $parentType = $context->getParentInputType();
+                $type       = $parentType === null ? null : Type::getNullableType($parentType);
```
spawnia, line 77:
```suggestion
                $type       = $parentType === null
                    ? null
                    : Type::getNullableType($parentType);
```

tests/UserArgsTest.php
```diff
@@ -0,0 +1,71 @@
...
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Schema;
+use PHPUnit\Framework\TestCase;
+
+class UserArgsTest extends TestCase
+{
+    /**
+     * Returns error when a scalar value is provided for a non-existent input field.
+     */
```
spawnia, line 18:
```suggestion
```

tests/UserArgsTest.php
```diff
@@ -0,0 +1,71 @@
...
+            })
+        }
+        ';
+        $result = GraphQL::executeQuery($this->schema(), $query)->toArray();
+        self::assertEquals('Field "scalar" is not defined by type InputType.', $result['errors'][0]['message']);
+    }
+
+    /**
+     * Returns error when an array value is provided for a non-existent input field.
+     */
```
spawnia, line 35:
```suggestion
```

PR #896, opened 2021-07-25
Drop @inheritDoc phpdoc

simPod:
Where not required by `SlevomatCodingStandard.TypeHints.XXXTypeHint.MissingTraversableTypeHintSpecification`


spawnia:
Can we just get rid of those? To me it seems like pure boilerplate with no benefit at all, is there any?

Issue #894, opened 2021-07-20
Fully comply with the spec for the error result format


spawnia:
http://spec.graphql.org/draft/#sec-Errors.Error-result-format

> GraphQL services should not provide any additional entries to the error format since they could conflict with additional entries that may be added in future versions of this specification.

If individual services decide to add such entries, that is not even as bad. Given this library is supposed to be a non-opinionated base implemenation of the spec, we absolutely should not.

I already started a pull request that removes the `category` entry. I think we should do the same for other non-spec entries such as `debugMessage` and `trace` (are there more?).

simPod:
hmm though I must admit I use those 😅 (`debugMessage` and `trace`). Have not checked yet how hard it would be to implement it in user land.

spawnia:
To clarify: I would drop `category`, but move the other entries under `extensions`.

PR #898, opened 2021-07-29
Move debugMessage and trace into error extensions

spawnia:
Resolves https://github.com/webonyx/graphql-php/issues/894


spawnia:
I will finish the tests once we agree this is the way to go forward.

simPod (commented):
LGTM
_disclosure: I'm not a graphql expert_ 🙃

Issue #570, opened 2019-11-02
Scalar types should throw exception different from the base `Exception`


vladar:
Follow up after this note: https://github.com/webonyx/graphql-php/issues/356#issuecomment-443249654

PR #903, opened 2021-07-29
Throw `SerializationError` over client safe `Error` when failing to serialize leaf types

spawnia:
Resolves https://github.com/webonyx/graphql-php/issues/570


src/Error/SerializationError.php
```diff
@@ -0,0 +1,17 @@
...
+
+use Exception;
+
+/**
+ * Thrown when failing to serialize a scalar value.
+ *
+ * Not generally safe for clients, as the wrong given value could
+ * be something not intended to ever be seen by clients.
+ */
+class SerializationError extends Exception
```
simPod, line 15: Shouldn't we name it ScalarSerializationError?

or even `ScalarSerializationFailed extends Exception`?

phpcs.xml.dist
```diff
@@ -21,6 +21,8 @@
         <exclude name="SlevomatCodingStandard.PHP.RequireExplicitAssertion.RequiredExplicitAssertion" />
         <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingAnyTypeHint" />
         <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint" />
+        <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingTraversableTypeHintSpecification" />
+        <exclude name="SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingAnyTypeHint" />
```
simPod, line 25: I'd keep those. If there's any issue with them, we should address it in the CS lib

src/Type/Definition/BooleanType.php
```diff
@@ -14,33 +13,22 @@
...
 {
-    /** @var string */
     public $name = Type::BOOLEAN;
 
-    /** @var string */
     public $description = 'The `Boolean` scalar type represents `true` or `false`.';
 
     /**
-     * Serialize the given value to a boolean.
+     * Serialize the given value to a bool.
```
simPod, line 21: I'd use `value of bool type` or kept `boolean`

src/Type/Definition/BooleanType.php
```diff
@@ -14,33 +13,22 @@
 
 class BooleanType extends ScalarType
 {
-    /** @var string */
```
simPod, line 17: we're missing type now

simPod (changes requested):

phpcs.xml.dist
spawnia: The issue is that it does not respect inheritance, without using the lovely `@inheritDoc`.

simPod: So we can keep inheritdoc for now, there's one or two. Better than omitting types everywhere else.

src/Error/SerializationError.php
spawnia: I chose this name because serialization is not limited to scalars, but rather general to leaf types. Your comment did remind me to use it in `EnumType` as well, thank you.

src/Type/Definition/BooleanType.php
spawnia: Capitalized it, since it refers to the GraphQL type.

src/Type/Definition/BooleanType.php
spawnia: Nope, it is defined in `Type`.

spawnia:
I let the `SerializationError` thrown in Enum serialization bubble up, given what the spec says at http://spec.graphql.org/draft/#sec-Enums.Result-Coercion

> GraphQL services must return one of the defined set of possible values. If a reasonable coercion is not possible they must raise a field error.

phpcs.xml.dist
spawnia: We are not omitting types anywhere though. PHPStan has still got our back, PHP_CodeSniffer is just flawed in that regard and more trouble than its worth.

simPod: hm, that's true. Is level 3 enough to check that? I'll do some tests eventually.

src/Type/Definition/StringType.php
```diff
@@ -17,45 +17,29 @@
...
+            throw new SerializationError(
                 'String cannot represent value: ' . Utils::printSafe($value)
             );
         }
 
         return (string) $value;
     }
 
-    /**
-     * @param mixed $value
```
simPod, line 52: I'd keep those mixed so they can be converted to `(mixed $value)` on php 8 automatically.

With this change we wouldn't be able to distinguish intentional mixed versus forgotten type. Also, consumers can `@inheritDoc`

src/Type/Definition/BooleanType.php
```diff
@@ -14,33 +13,22 @@
...
     public $name = Type::BOOLEAN;
```
simPod, line 16: why not do this so it's type safe?

```suggestion
    public string $name = Type::BOOLEAN;
```

src/Error/SerializationError.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Error;
+
+use Exception;
+
+/**
+ * Thrown when failing to serialize a scalar value.
```
simPod, line 10: https://github.com/webonyx/graphql-php/pull/903/files#r679674725

Ah, then this needs to be modified

src/Type/Definition/BooleanType.php
simPod: but when the type of Type changes, it won't be detected here by SA

phpcs.xml.dist
simPod: When I remove `@param Promise[]|mixed[] $promisesOrValues Promises or values.` from `PromiseAdapter::all()` it is not detected unfortunatelly

https://github.com/webonyx/graphql-php/blob/c886d2bf0679f6dc010209c2e64961776c74bd27/src/Executor/Promise/PromiseAdapter.php#L85

spawnia: It comes in at level 6. We'll get there eventually 😉

src/Type/Definition/StringType.php
spawnia: The parent method still has the explicit `mixed` annotation.

src/Type/Definition/BooleanType.php
spawnia: See https://github.com/webonyx/graphql-php/pull/906

src/Type/Definition/BooleanType.php
spawnia: https://github.com/webonyx/graphql-php/pull/906

phpcs.xml.dist
simPod: Eventually, but until then we're not covered by checks. Still disagree on disabling these until there's alternative in place.

src/Type/Definition/StringType.php
simPod: But the thing with php is to specify types even in implementations but that's related to previous discussion on disabling missing type checks.  
We can have @inheritDoc here so it explicitly says we want `mixed` here. 

With this change it's ambiguous: is it forgotten or inherited? What if parent type narrows, should it be narrowed here as well?

spawnia: Good tooling I use assumes the parent PHPDoc is inherited, unless specifically overwritten. Perhaps we can document and reify that assumption as part of the contribution guidelines?

simPod: Do as you think is the best but personally I don't like trading automated checks for a piece of text few people read 🤷🏾‍♀️

simPod (approved):

src/Type/Definition/StringType.php
spawnia: I am fairly confident we will keep this project moving towards stronger types without duplicating code.

phpcs.xml.dist
simPod: https://github.com/doctrine/coding-standard/pull/258

PR #906, opened 2021-08-02
Use native PHP types for properties of `Type` and its subclasses

spawnia:


simPod (approved):

sebastienbarre:
I love native types, but this is possibly going to trigger fatal errors and break a lot of people's code.

For example:
```
class SearchTermType extends ScalarType
{
  /**
   * Scalar type description.
   */
  public $description = 'The `SearchTerm` scalar type represents textual data [blahblah].';
...
```
It is now breaking:

```
Fatal error: Type of SearchTermType::$description must be ?string (as in class GraphQL\Type\Definition\Type) inSearchTermType.php on line 0
```

This is also breaking external dependencies that provide custom scalars. For example, [mll-lab/graphql-php-scalars: A collection of custom scalar types for usage with https://github.com/webonyx/graphql-php](https://github.com/mll-lab/graphql-php-scalars):
```
Fatal error: Type of MLL\GraphQLScalars\Email::$description must be ?string (as in class GraphQL\Type\Definition\Type) in Vendor/mll-lab/graphql-php-scalars/src/Email.php on line 10
```
and these I can't fix without manually patching.

This is for a major release I assume, version 15?

spawnia:
I am maintaining v15 compatible branches of my packages depending on this library:
- https://github.com/mll-lab/graphql-php-scalars/tree/graphql-php-15
- https://github.com/nuwave/lighthouse/tree/graphql-php-15

PR #909, opened 2021-08-03
Clean up types for src/Server

spawnia:


src/Server/Helper.php
```diff
@@ -159,7 +158,7 @@ public function parseRequestParams($method, array $bodyParams, array $queryParam
      *
      * @api
```
simPod, line 159: do we know the array k,v type?

src/Server/ServerConfig.php
```diff
@@ -57,45 +57,39 @@ public static function create(array $config = [])
         return $instance;
     }
 
-    /** @var Schema|null */
-    private $schema;
+    private ?Schema $schema = null;
 
     /** @var mixed|callable */
-    private $context;
+    private $context = null;
```
simPod, line 63: null is missing in `@var`

simPod (approved):

src/Server/Helper.php
spawnia: It is specified.

Generally, I did specify the type in return values when known and left it open in params where it does not matter.

src/Server/ServerConfig.php
spawnia: Thanks, added.

PR #916, opened 2021-08-17
Always throw `RequestError` with useful message when clients provide an invalid JSON body

FritsvanCampen:
I'm seeing this error on production. When a client makes an invalid or empty request to the server it fatals. The correct behavior is that it should reply with a 400 Bad Request.

I don't understand why this was considered an InvariantViolation or why it is specifically tested against. It's a LogicException so it should not be part of the interface.

### DIFF
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -533,7 +533,7 @@ class Helper
                     : json_decode((string) $request->getBody(), true);
 
                 if ($bodyParams === null) {
-                    throw new InvariantViolation(
+                    throw new RequestError(
                         $request instanceof ServerRequestInterface
                          ? 'Expected to receive a parsed body for "application/json" PSR-7 request but got null'
                          : 'Expected to receive a JSON array in body for "application/json" PSR-7 request'
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -535,8 +535,8 @@ class Helper
                 if ($bodyParams === null) {
                     throw new RequestError(
                         $request instanceof ServerRequestInterface
-                         ? 'Expected to receive a parsed body for "application/json" PSR-7 request but got null'
-                         : 'Expected to receive a JSON array in body for "application/json" PSR-7 request'
+                         ? 'Expected to receive a parsed body for "application/json" PSR-7 request, got: null'
+                         : 'Expected to receive a JSON body for "application/json" PSR-7 request, got: null'
                     );
                 }
 
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -528,16 +528,16 @@ class Helper
             if (stripos($contentType[0], 'application/graphql') !== false) {
                 $bodyParams = ['query' => (string) $request->getBody()];
             } elseif (stripos($contentType[0], 'application/json') !== false) {
-                $bodyParams = $request instanceof ServerRequestInterface
-                    ? $request->getParsedBody()
-                    : json_decode((string) $request->getBody(), true);
-
-                if ($bodyParams === null) {
-                    throw new RequestError(
-                        $request instanceof ServerRequestInterface
-                         ? 'Expected to receive a parsed body for "application/json" PSR-7 request, got: null'
-                         : 'Expected to receive a JSON body for "application/json" PSR-7 request, got: null'
-                    );
+                if ($request instanceof ServerRequestInterface) {
+                    $bodyParams = $request->getParsedBody();
+                    if ($bodyParams === null) {
+                        throw new InvariantViolation('Expected to receive a parsed body for "application/json" PSR-7 request, got: null');
+                    }
+                } else {
+                    $bodyParams = json_decode((string) $request->getBody(), true);
+                    if ($bodyParams === null) {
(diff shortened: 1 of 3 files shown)


src/Server/Helper.php
```diff
@@ -533,7 +533,7 @@ public function parsePsrRequest(RequestInterface $request)
                     : json_decode((string) $request->getBody(), true);
 
                 if ($bodyParams === null) {
-                    throw new InvariantViolation(
+                    throw new RequestError(
                         $request instanceof ServerRequestInterface
                          ? 'Expected to receive a parsed body for "application/json" PSR-7 request but got null'
                          : 'Expected to receive a JSON array in body for "application/json" PSR-7 request'
```
spawnia, line 539:
```suggestion
                         : 'Expected to receive a JSON body for "application/json" PSR-7 request, got: null'
```

src/Server/Helper.php
```diff
@@ -533,7 +533,7 @@ public function parsePsrRequest(RequestInterface $request)
                     : json_decode((string) $request->getBody(), true);
 
                 if ($bodyParams === null) {
-                    throw new InvariantViolation(
+                    throw new RequestError(
                         $request instanceof ServerRequestInterface
                          ? 'Expected to receive a parsed body for "application/json" PSR-7 request but got null'
```
spawnia, line 538:
```suggestion
                         ? 'Expected to receive a parsed body for "application/json" PSR-7 request, got: null'
```

spawnia (commented):
Generally agree with the change, error messages can be cleaned up a bit.

FritsvanCampen:
> Generally agree with the change, error messages can be cleaned up a bit.

I don't really have an opinion on the contents of the error message. I've included your suggestion.

src/Server/Helper.php
```diff
@@ -533,10 +533,10 @@ public function parsePsrRequest(RequestInterface $request)
                     : json_decode((string) $request->getBody(), true);
 
                 if ($bodyParams === null) {
-                    throw new InvariantViolation(
+                    throw new RequestError(
                         $request instanceof ServerRequestInterface
-                         ? 'Expected to receive a parsed body for "application/json" PSR-7 request but got null'
-                         : 'Expected to receive a JSON array in body for "application/json" PSR-7 request'
+                         ? 'Expected to receive a parsed body for "application/json" PSR-7 request, got: null'
```
spawnia, line 538: Codecov complains about missing coverage for this line. Now that I think about it more, we can just delete the conditional. Given this is a client facing error, we can simplify to just always say:

```
Expected to receive a JSON body for "application/json" request, got: null
```

More useful for clients, since they do not care about PSR-7 or any other server-internal mechanics.

simPod: @spawnia IIRC the same payload is represented differently in `ServerRequestInterface`, that's why there's "parsed" body. `ServerRequestInterface` has `getParsedBody()` method.

So in once case it's a json body and in second case it's parsed payload

spawnia: At this point we know the client sent `content-type: application/json`. Thus, I think we should inform them that they should also send a non-null JSON body in their request. The internal method name of the PSR interface seems irrelevant for them?

FritsvanCampen:
I've made a change to the code. I think it makes sense to throw an InvariantViolation in the case of a `ServerRequestInterface`, but not in the case of a client request. This should resolve the warning about code coverage (because the code coverage doesn't care about LogicExceptions?)

spawnia:
A `ServerRequestInterface` represents an incoming HTTP request, so it *is* a request sent by a client.

From the docs of `ServerRequestInterface::getParsedBody()` at https://www.php-fig.org/psr/psr-7:

> A null value indicates the absence of body content.

It is quite possible for clients to send a request with no body. A proper implementation of PSR-7 does not require a body to be present. Thus, no invariant was violated - it is simply that the client sent a bad request. That's a `RequestError`.

src/Server/Helper.php
simPod: Since we support `ServerRequestInterface` I can't imagine every possible way to construct it so I could tell there was a json at the beginning in all cases. I'd prefer to keep those two branches separated for clarity, though we can definitely reword it, e.g.

```diff
- Expected to receive a parsed body for "application/json" PSR-7 request but got null
+ Expected to receive a body for "application/json" PSR-7 request but the body is missing
```

spawnia: Please read the surrounding code. It must have been JSON, otherwise this execution branch is never entered.

spawnia:
I unified all errors related to incoming JSON requests.

Once it is known a client should send JSON via `Content-Type: application/json`, the expectation is that the body will indeed contain JSON; and that JSON will either be an object (single request) or and array (batched request).

Thus, all errors will start out by telling the client: `Expected JSON object or array for "application/json" request`.

Subsequently, one of the following can happen:
- We are unable to parse the JSON. We should tell the client about it, and tell them why (most likely a syntax error).
- The provided JSON evaluates to something other then an object or array. We should tell the client about it, and tell them what their payload was.
- The provided JSON is valid and holds appropriate data. No error occurs.

FritsvanCampen:
@spawnia thank you for the merge!

PR #917, opened 2021-08-17
Add strict native type hints for Schema and SchemaConfig

spawnia:


src/Type/Schema.php
```diff
@@ -51,38 +51,35 @@
...
 {
-    /** @var SchemaConfig */
-    private $config;
+    private SchemaConfig $config;
 
     /**
      * Contains currently resolved schema types
      *
-     * @var Type[]
+     * @var array<Type>
```
simPod, line 59: I think the key here is a `string`

src/Type/Schema.php
```diff
@@ -51,38 +51,35 @@
...
-     *
-     * @var bool
+     * True when $resolvedTypes contains all possible schema types.
      */
-    private $fullyLoaded = false;
+    private bool $fullyLoaded = false;
 
-    /** @var Error[] */
-    private $validationErrors;
+    /** @var array<Error> */
```
simPod, line 75:
```suggestion
    /** @var array<int, InvariantViolation|Error> */
```

I took the type from the method below

simPod (approved):

src/Type/Schema.php
spawnia: I double checked and found it is always `array<int, Error>`.

simPod (approved):

spawnia:
I am not 100% sure if native type errors are preferrable to the schema validation errors.

Downsides:
- early return before accumulated schema errors
- no way to provide guidance or additional context

Upsides:
- validation is cheap and always caught early
- standardized error that most developers know how to read and diagnose
- code is clear and safe

Since we have the luxury of native types that JS does not have, I think it is generally positive for us to take advantage of it. In this case, the native error messages are pretty good still.

simPod:
Yup, was thinking in a same direction 👍🏿

PR #918, opened 2021-08-20
Expose structured enumeration of directive locations

spawnia:


src/Language/DirectiveLocation.php
```diff
@@ -55,8 +55,10 @@ class DirectiveLocation
         self::INPUT_FIELD_DEFINITION => self::INPUT_FIELD_DEFINITION,
     ];
 
+    const LOCATIONS = self::EXECUTABLE_LOCATIONS + self::TYPE_SYSTEM_LOCATIONS;
```
simPod, line 58: Can we, also add access modifier?

simPod (approved):

src/Language/DirectiveLocation.php
spawnia: I know `php-cs-fixer` can, how about `PHP_CodeSniffer`?

simPod: It's disabled rn https://github.com/webonyx/graphql-php/blob/f7c99d24fe6f65d287e007b5284e8e71e973fc88/phpcs.xml.dist#L20 but we should not introduce more violations.

PR #928, opened 2021-09-02
Align `BuildSchemaTest` with `graphql-js`

vhenzl:
While working on tests for #921, I noticed that tests for schema building and extending are not in pair with the reference implementation.

This PR aligns `BuildSchemaTest.php` with `buildASTSchema-test.js` (up to v15.5.2) and reveals where `graphql-php` lags behind. 

The first commit (v14.0) is the messiest one (a lot of various changes), but the only one where tests are passing. From the v14.7 commit up, the behaviour between `graphql-php` and `graphql-js` starts to differ and tests to fail.

The v14.0 commit could be merged (I can create a separated PR for it), the other commits are currently more for reference only and can be cherry-picked when a related problem is resolved.

I'm planning to do the same for `SchemaExtenderTest`.

---
Tests no longer present in `buildASTSchema-test.js` have been moved to `BuildSchemaLegacyTest`. These are the corresponding commits in` graphql-js`:
- https://github.com/graphql/graphql-js/commit/64a5c3448a201737f9218856786c51d66f2deabd
- https://github.com/graphql/graphql-js/commit/257797a0ebdddd3da6e75b7c237fdc12a1a7c75a
- https://github.com/graphql/graphql-js/commit/9b7a8af43fc0865a01df5b5a084f37bbb8680ef8
- https://github.com/graphql/graphql-js/commit/3b9ea61f2348215dee755f779caef83df749d2bb
---

Tests fail because:

#### v14.7 commit
- [x] a problem with the printing of empty types (e.g. `type Query` or `union EmptyUnion`) (fixed in #940)
- [ ] scalar types are added to schema automatically even if they are not used in SDL (#964)

#### v15.0 commit
- [x] `extensionASTNodes` is not set
- [x] missing `@` in directive name in error messages
- [ ] build doesn't throw on unknown types

#### v15.5 commit
- [x] different order of default directives


spawnia:
> The v14.0 commit could be merged (I can create a separated PR for it)

That would be great!

spawnia:
@vhenzl do you have an idea when you would pick this up again? I went through the open issues and pull requests, as far as I can tell this is the last one that should be in v15.

spawnia:
Thanks for your feedback. I might pick this merge request up and finish it, I will tell you when I start working on it.

The missing features seem like something that can be added without breaking changes, so we do not necessarily need to wait with v15.

vhenzl:
I'm just working on it.

spawnia:
@vhenzl I made the necessary changes to make the tests pass. Let me know if you would like to make other changes or I should merge this as is. We can fix the partial/skipped tests when implementing the missing features.

vhenzl:
You can merge it.

PR #929, opened 2021-09-04
Align `SchemaExtenderTest` with `graphql-js`

vhenzl:
Follow-up to #928. Aligns `SchemaExtenderTest.php` with `extendSchema-test.js` (up to `graphql-js` v15.5.2).

### v14.7 commit
Tests no longer present in `extendSchema-test.js` have been moved to `SchemaExtenderLegacyTest`. These are the corresponding commits in `graphql-js` (relevant also for #928):
- https://github.com/graphql/graphql-js/commit/c1745228b2ae5ec89b8de36ea766d544607e21ea
- https://github.com/graphql/graphql-js/commit/257797a0ebdddd3da6e75b7c237fdc12a1a7c75a
- https://github.com/graphql/graphql-js/commit/3b9ea61f2348215dee755f779caef83df749d2bb
- https://github.com/graphql/graphql-js/commit/e6a3f08cc92594f68a6e61d3d4b46a6d279f845e
- https://github.com/graphql/graphql-js/commit/9b7a8af43fc0865a01df5b5a084f37bbb8680ef8

Tests in this commit fail because:
- [x] `repeatable` in directives is lost when calling `extend()` (fixed in #931)
- [x] `Schema::getAstNode()` returns `null` for an extended schema
- [ ] scalar types are added to schema automatically even if they are not used in SDL (#964)

### v15.0 commit
Reasons, why tests fail (in addition to those mentioned above):
- [x] a problem with parsing/printing of types without fields (e.g. `type Query`) (fixed in #940)
- [x] string comments become multiline (e.g. `"""Comment"""` becomes `"""\nComment\n"""` (fixed in #938)
- [x] different order of definitions affects `printSchemaChanges()` (#954)
- [x] missing `@` in directive name in error messages


spawnia:
Awesome work so far. What is your plan for integrating those changes?

vhenzl:
> What is your plan for integrating those changes?

I'm not quite sure yet. I want to have a look at `print()` and `printSchema()` first. If I can get these to handle empty types and string blocks correctly, that would resolve a significant number of test failures. The other thing that accounts for a big amount of failing tests is `astNode` and `extensionASTNodes` not being properly set/copied, which shouldn't be hard to fix. With these two problems resolved, I believe most of this and also #928 should be possible to merge up to v15.5.

spawnia:
If small pull requests emerge, that would be great. If not, I am also fine with a large pull request that fixes it all.

spawnia:
Thank you for this. Apart from https://github.com/webonyx/graphql-php/issues/1140, we are now in sync.

PR #930, opened 2021-09-04
Add `concatAST` utility function

vhenzl:
Cherry-picked from #929.

### DIFF
--- a/tests/Utils/ConcatASTTest.php
+++ b/tests/Utils/ConcatASTTest.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Tests\Utils;
+
+use GraphQL\Language\Parser;
+use GraphQL\Language\Printer;
+use GraphQL\Utils\AST;
+use PHPUnit\Framework\TestCase;
+
+use function preg_match;
+use function preg_replace;
+use function trim;
+
+class ConcatASTTest extends TestCase
+{
+    private function dedent(string $str): string
+    {
+        $trimmedStr = trim($str, "\n");
+        $trimmedStr = preg_replace('/[ \t]*$/', '', $trimmedStr);
+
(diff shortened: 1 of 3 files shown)


tests/Utils/ConcatASTTest.php
```diff
@@ -0,0 +1,63 @@
...
+use GraphQL\Utils\AST;
+use PHPUnit\Framework\TestCase;
+
+use function preg_match;
+use function preg_replace;
+use function trim;
+
+class ConcatASTTest extends TestCase
+{
+    private function dedent(string $str): string
```
spawnia, line 18: Is this going to be needed for multiple tests?

vhenzl: Yeah, definitely. I thought I would later extract it somewhere. But I'm not quite sure where to put it. `graphql-js` has [`__testUtils__`](https://github.com/graphql/graphql-js/tree/main/src/__testUtils__) in `src`. Maybe we could colocate testing-related stuff to `tests/TestUtils`?

I'm also wondering if make it a static method like `Dedent::dedent()` (unnecessary duplication) or `Utils::dedent()` (yet another `Utils` class), or if wouldn't make more sense to have it just like a function (in a namespace).

spawnia: `Tests\TestUtils::dedent()` seems good.

vhenzl: I was thinking more about `TestUtils` directory and putting there also the "kitchen sink" and other things scattered elsewhere. But we can look at that later…

And I have just noticed, there is `TestUtils` class in `test/Language`. Should I move it to the root, or create another class?

spawnia: Do what you think is right, I will review it.

spawnia: Can you update this with nowdoc syntax?

vhenzl:
This should be good to go now.

PR #934, opened 2021-09-08
Preserve extended methods from class-based types in `SchemaExtender::extend()`

### LINKED ISSUE #933: SchemaExtender Not Loading Custom Scalar Types Correctly
The `SchemaExtender` doesn't appear to load custom scalar types that extend `ScalarType` correctly.  You get the following error: `"Expected a value of type Email but received: test@test.com. Value of type null is not callable"`.  Inline definitions using `CustomScalarType` work. `error_log`ing [here ](https://github.com/webonyx/graphql-php/blob/master/src/Executor/ReferenceExecutor.php#L952) shows that `$returnType` is an instance of `CustomScalarType` rather than `EmailType` (when using `SchemaExtender`).

Test case: https://gist.github.com/abdullahseba/0e2cb2a8598eba0f44b736806d599e68

abdullahseba:
Resolves #933


spawnia:
It seems like extending class-based scalar types does not work at all as of now, given the implementation of `SchemaExtender`: https://github.com/webonyx/graphql-php/blob/2abac8cdd9cdbc892dd09d05eae15b1431eb3b7c/src/Utils/SchemaExtender.php#L132-L143

The `$config` values do not exist in a class. How else could we extend it then? Perhaps using closures to reference the orginal scalar's methods?

tests/Utils/SchemaExtenderTest.php
```diff
@@ -2106,4 +2108,29 @@ public function testSupportsTypeConfigDecorator(): void
...
+
+        $someScalar              = $this->testSchema->getType('SomeScalar');
+        $someScalarClass         = $this->testSchema->getType('SomeScalarClass');
+        $extendedSomeScalar      = $extendedSchema->getType('SomeScalar');
+        $extendedSomeScalarClass = $extendedSchema->getType('SomeScalarClass');
+
+        self::assertInstanceOf(CustomScalarType::class, $someScalar);
+        self::assertInstanceOf(SomeScalarClassType::class, $someScalarClass);
+        self::assertInstanceOf(CustomScalarType::class, $extendedSomeScalar);
+        self::assertInstanceOf(SomeScalarClassType::class, $extendedSomeScalarClass);
```
spawnia, line 2134: This test pointed me in the right direction, but I do not think it is correct. The `SchemaExtender` consistently produces new types, which is why `$someScalar !== $extendedSomeScalar`.

What should happen though is that the behaviour of the type does not get erased. Thus, the test should ensure the method of the original scalar class is called when using `$extendedSomeScalarClass`.

abdullahseba: Makes sense, but should that be the case if the scalar type isn't actually extended in the new Schema?  In my case, I was just using the scalar type that had been declared in the original schema rather than extending it.

abdullahseba: Also, can scalars types actually be extended?  I don't think you can do much with scalars in GQL and I didn't think you can extend a native schema with another in PHP.

spawnia: Certainly possible, see http://spec.graphql.org/draft/#sec-Scalar-Extensions. It is useful for adding directives.

There is consistency in always producing a new type.

abdullahseba:
I've implemented a fix using closures as you suggested and it doesn't appear to break any tests.  But at the same time, the testing for scalars is probably not compleate?  Only `serialize()` is really tested.

spawnia:
> I've implemented a fix using closures as you suggested and it doesn't appear to break any tests. But at the same time, the testing for scalars is probably not compleate? Only `serialize()` is really tested.

I cut down the test case to be as minimal as possible and cover all three methods.

src/Utils/SchemaExtender.php
```diff
@@ -203,6 +198,7 @@ protected static function extendInputFieldMap(InputObjectType $type): array
 
         $extensions = static::$typeExtensionsMap[$type->name] ?? null;
         if ($extensions !== null) {
+            /** @var InputObjectTypeExtensionNode $extension */
```
simPod, line 201: maybe we could use `assert()`s instead of `@var`

src/Utils/SchemaExtender.php
```diff
@@ -296,15 +294,22 @@ protected static function extendImplementedInterfaces(ImplementingType $type): a
...
                 }
             }
         }
 
         return $interfaces;
     }
 
-    protected static function extendType($typeDef)
+    /**
+     * @param ListOfType|NonNull|(Type &NamedType) $typeDef
```
simPod, line 308:
```suggestion
     * @param ListOfType|NonNull|(Type&NamedType) $typeDef
```

src/Utils/SchemaExtender.php
```diff
@@ -314,45 +319,43 @@ protected static function extendType($typeDef)
             return Type::nonNull(static::extendType($typeDef->getWrappedType()));
         }
 
+        // @phpstan-ignore-next-line generic is not properly caught
```
simPod, line 322: if this is not intentionally caught, we should update baseline instead. Phpstan will then throw error when it's caught and the baseline entry is redundant

src/Utils/SchemaExtender.php
```diff
@@ -296,15 +294,22 @@ protected static function extendImplementedInterfaces(ImplementingType $type): a
             /** @var ObjectTypeExtensionNode|InterfaceTypeExtensionNode $extension */
             foreach ($extensions as $extension) {
                 foreach ($extension->interfaces as $namedType) {
-                    $interfaces[] = static::$astBuilder->buildType($namedType);
+                    /** @var InterfaceType $interface */
```
simPod, line 297: maybe we could use `assert()`s instead of `@var`

src/Utils/SchemaExtender.php
```diff
@@ -366,13 +369,14 @@ protected static function extendFieldMap($type): array
...
                 'args' => static::extendArgs($field->args),
-                'astNode' => $field->astNode,
                 'resolve' => $field->resolveFn,
+                'astNode' => $field->astNode,
             ];
         }
 
         $extensions = static::$typeExtensionsMap[$type->name] ?? null;
         if ($extensions !== null) {
+            /** @var ObjectTypeExtensionNode|InputObjectTypeExtensionNode $extension */
```
simPod, line 379: maybe we could use `assert()`s instead of `@var`

tests/Utils/SomeInterfaceClassType.php
```diff
@@ -0,0 +1,22 @@
...
+namespace GraphQL\Tests\Utils;
+
+use GraphQL\Type\Definition\InterfaceType;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\ResolveInfo;
+
+/**
+ * Custom class-based interface type for testing.
+ */
+final class SomeInterfaceClassType extends InterfaceType
```
simPod, line 14: are these classes test specific? I'd keep them under `\Fixture`; namespace at least so things don't get messy soon

simPod (approved):
no objections, some nits

👍🏾

src/Utils/SchemaExtender.php
spawnia: I added comments to explain why this must hold true.

src/Utils/SchemaExtender.php
spawnia: I would prefer that, PHP_CS keeps messing it up. I don't like that tool, php-cs-fixer does a much better job in my experience.

PR #936, opened 2021-09-12
Introduce reusable test utility `dedent()`

vhenzl:
Equivalent of [`dedent()`](https://github.com/graphql/graphql-js/blob/99d6079434/src/__testUtils__/dedent.js) test utility function from `graphql-js`.

Its behaviour matches `graphql-js` v15 (v16 version trims trailing newlines).

Introduced in reaction to https://github.com/webonyx/graphql-php/pull/930#pullrequestreview-747366447

It lives in `tests/TestUtils` directory, where I later intend to move other testing utilities and fixtures (similarly to `graphql-js`'s [`__testUtils__`](https://github.com/graphql/graphql-js/tree/99d6079434e2353ce042fd6df93bb8076c59d47f/src/__testUtils__)).

Personally, I would prefer to have the utility as a function, rather than a static method. But that wouldn't quite fit into the current codebase. 

Compare these two snippets:
```php
$sdl = TestUtils::dedent('
    type Query {
      field(in: Input): String
    }
');
```
```php
$sdl = dedent('
    type Query {
      field(in: Input): String
    }
');
```

`TestUtils::` is completely unnecessary noise that doesn't add any additional information about `dedent()`. And it also further complicates already difficult diffing with `graphql-js`.

To keep the code consistent with the current "static methods for everything" approach, I went with `Str::dedent()`. It's at least short and has the meaning of "dedent is a function operating on a string".

However, if you would like to give it a go, I'm happy to refactor the method to a function. There are more static methods that could be changed to functions and I believe the codebase could benefit from it.

### DIFF
--- a/tests/TestUtils/Str.php
+++ b/tests/TestUtils/Str.php
@@ -14,7 +14,7 @@ class Str
      * Fixes indentation. Also removes leading newlines
      * and trailing spaces and tabs, but keeps trailing newlines.
      *
-     * Example usage:
+     * @example
      * $str = Str::dedent('
      *   {
      *     test
--- a/tests/TestUtils/Str.php
+++ b/tests/TestUtils/Str.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Tests\TestUtils;
+
+use function preg_match;
+use function preg_replace;
+use function trim;
+
+class Str
+{
+    /**
+     * Fixes indentation. Also removes leading newlines
+     * and trailing spaces and tabs, but keeps trailing newlines.
+     *
+     * Example usage:
+     * $str = Str::dedent('
+     *   {
+     *     test
+     *   }
+     * ');
+     * $str === "{\n  test\n}\n";
+     */
+    public static function dedent(string $string): string
+    {
+        $trimmedString = trim($string, "\n");
+        $trimmedString = preg_replace('/[ \t]*$/', '', $trimmedString);
+        preg_match('/^[ \t]*/', $trimmedString, $indentMatch);
+        $indent = $indentMatch[0];
+
+        return preg_replace('/^' . $indent . '/m', '', $trimmedString);
+    }
+}
--- a/tests/TestUtils/StrTest.php
+++ b/tests/TestUtils/StrTest.php
@@ -1,0 +1,154 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Tests\TestUtils;
+
+use PHPUnit\Framework\TestCase;
+
+use function implode;
+
+class StrTest extends TestCase
+{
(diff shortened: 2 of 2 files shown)


spawnia:
We can have a discussion about using functions over static methods in an issue, separate from this pull request. Last time I checked, they were suboptimal for performance when compared with autoloading.

tests/TestUtils/Str.php
```diff
@@ -0,0 +1,34 @@
...
+use function preg_replace;
+use function trim;
+
+class Str
+{
+    /**
+     * Fixes indentation. Also removes leading newlines
+     * and trailing spaces and tabs, but keeps trailing newlines.
+     *
+     * Example usage:
```
spawnia, line 17:
```suggestion
     * @example
```

tests/TestUtils/StrTest.php
```diff
@@ -0,0 +1,154 @@
...
+            me: User
+          }
+
+          type User {
+            id: ID
+            name: String
+          }
+        ');
+        self::assertEquals(
+            implode(
```
spawnia, line 31: Wouldn't it be easier to use a NOWDOC?

vhenzl: I was thinking about it for a short moment myself, but I dismissed the idea. I'm trying to keep the code generally and tests particularly as much close to the reference implementation as possible to make it easy to reflect future changes.

`dedent()` deals with whitespaces and empty lines, so it's important to be clear about what the testing string is. I believe using `implode()` is more explicit than NOWDOC there is easier to get lost in the indentation etc. It also makes changes like this one super obvious: https://github.com/graphql/graphql-js/commit/1e46389a1970bb3609f33029cfcf0a41f65ad999#diff-6738ab51d9c26ea2d4b271f22bb521b267a1e290755dc4cf5da78fcf86facb6c

vhenzl:
@spawnia, your comment (https://github.com/webonyx/graphql-php/pull/936#discussion_r707634424) made me think if we actually need `dedent()` at all and if nowdoc/heredoc with indentation wouldn't be enough. Probably would. In some way, it's actually closer to the tagged template ` dedent`` `.

One disadvantage I can see is that nowdoc ignores the last newline, so it needs an extra one to get the same behaviour:

```php
$sdl = <<<'GQL'
    type Query {
      field(in: Input): String
    }

    GQL;
```

Another disadvantage is that nowdoc is a language construct and we can't control its behaviour as we could with `dedent()`. So the change like https://github.com/graphql/graphql-js/commit/1e46389a1970bb3609f33029cfcf0a41f65ad999 would require us to modify all nowdoc occurrences in all tests instead of changing only `dedent()`'s implementation.

vhenzl:
@spawnia, have a look at #941. I have used nowdoc there and it looks fine. If you are happy with it, we can close this PR.

PR #938, opened 2021-09-14
Unify `Printer` and `SchemaPrinter` output for block strings

vhenzl:
Adds `StringBlock::print()` method, an equivalent of [`printBlockString()`](https://github.com/graphql/graphql-js/blob/v15.5.3/src/language/blockString.js#L94) function from `graphql-js`. It's used to print string blocks in `Printer` and `SchemaPrinter`.

Relevant changes in the reference implementation:
- https://github.com/graphql/graphql-js/pull/1751 
  (commits https://github.com/graphql/graphql-js/commit/ded32618cc17c3a42a79da09dd1f0516ceb85103 and https://github.com/graphql/graphql-js/commit/961dab320ba579b8639d218c041b6bd6b7f6664d)
- https://github.com/graphql/graphql-js/pull/2223

The behaviour matches `graphql-js` v15 (v16 removes the indentation inside of block strings: https://github.com/graphql/graphql-js/commit/45e33cefae831a9df2b1a9edebd03a85cdfb33ff)

### DIFF
--- a/src/Language/BlockString.php
+++ b/src/Language/BlockString.php
@@ -122,22 +122,29 @@ class BlockString
         $hasTrailingQuote     = $value[$valueLength - 1] === '"';
         $hasTrailingSlash     = $value[$valueLength - 1] === '\\';
         $printAsMultipleLines =
-            ! $isSingleLine ||
-            $hasTrailingQuote ||
-            $hasTrailingSlash ||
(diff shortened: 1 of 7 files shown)


vhenzl:
Ah, I didn't realise `symfony/polyfill-php80` is just a transitive dev dependency :/

PR #943, opened 2021-09-18
Fix website to display table of contents correctly

vhenzl:
`mkdocs-material` has support for table of contents enabled by default but right now it's not working properly (majority of pages doesn't show ToC at all, only [Concepts](https://webonyx.github.io/graphql-php/concepts/) page shows something but it's broken anyway).

The problem is, for ToC to work, every page must contain only one heading of level 1. So what this PR essentially does is to bump titles one level up. Plus it adds a few new headings or changes levels as appropriate. It also enables permalinks to easily link to a page section.

If a Markdown file doesn't have a level 1 heading at all, it's inherited from the page name in the navigation. I'm a bit undecided if to add level one everywhere to be explicit about page title or not.

Before:

webonyx github io_graphql-php_schema-definition

After:

localhost_8000_schema-definition

### DIFF
--- a/docs/class-reference.md
+++ b/docs/class-reference.md
@@ -1,4 +1,4 @@
-# GraphQL\GraphQL
+## GraphQL\GraphQL
 
 This is the primary facade for fulfilling GraphQL operations.
 See [related documentation](executing-queries.md).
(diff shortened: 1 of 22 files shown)


simPod (approved):
As for me, inheriting the title from nav is just fine. At least we avoid inconsistencies.

👍🏾

Issue #945, opened 2021-09-25
Field name as part of the field config


simPod:
Currently, we expect the object receives a field definitions as `key -> value` where `key` is a field name and `value` is there rest of field configuration. I propose to allow a name to be a part of field configuration.

Now, it is not possible to create a field factory as: 

```php
function createField(): array {
  // no way to pass a name here
  return [
    'description' => 'desc',
    ...
  ];
}
```

I'd love to have this for more flexibility:

```php
function createField(): array {
  return [
    'name' => 'Field',
    'description' => 'desc',
    ...
  ];
}
```

simPod:
Though, it is a bit in conflict with https://github.com/webonyx/graphql-php/pull/914 I'll try to find out what can be done.

simPod:
Hm, what if we drop dependency on an array-key and rely on a field config?

simPod:
For the sake of consistency it makes sense to me. E.g. objects have the `name` entry as well.

simPod:
Hm, there's a lot of complexity around resolving field name from keys.

simPod:
FieldDefinition expects name as a part of config

https://github.com/webonyx/graphql-php/blob/88f9f4913a8ebcafa43c61fee1761fcf26e3f63d/src/Type/Definition/FieldDefinition.php#L73-L75

vhenzl:
> Now, it is not possible to create a field factory […]

Have you tried it? What exactly is the problem? AFAIK, it should work. 

The complexity is to support ["shorthand definitions"](https://webonyx.github.io/graphql-php/type-definitions/object-types/#shorthand-field-definitions). The field name is required in the config passed to `FieldDefinition` constructor but optional in object config. If it's not provided in object config, it's inferred from the fields array key. The same applies to interface fields, [input object fields](https://webonyx.github.io/graphql-php/type-definitions/inputs/#:~:text=When%20not%20set%20-%20inferred%20from%20fields%20array%20key), [enum values](https://webonyx.github.io/graphql-php/type-definitions/enums/#shorthand-definitions) and even [field args](https://webonyx.github.io/graphql-php/type-definitions/object-types/#:~:text=inferred%20from%20args%20array%20key).

simPod:
This https://github.com/webonyx/graphql-php/pull/914 enforced input fields to be `array<string, ...>`

```
[...]

PR #946, opened 2021-09-27
Ensure all InputObjectType definitions styles work

simPod:
Related to https://github.com/webonyx/graphql-php/issues/945


simPod:
Ok, it passed here but fails on my localhost 🤔

spawnia:
Test cases were still lacking quite a bit. Do you think this covers every case now?

src/Type/Definition/InputObjectType.php
```diff
@@ -91,10 +93,6 @@ public function getFields(): array
 
     protected function initializeFields(): void
     {
-        if (isset($this->fields)) {
```
simPod: I intended to reduce complexity by putting the check here so it did not have to be repeated in other methods. Thought the method call was cheap and on some php optimization level this would have been inlined anyway I believe.

tests/Type/DefinitionTest.php
```diff
@@ -752,18 +752,69 @@ public function testInputObjectTypeAllowsRecursiveDefinitions(): void
...
             ],
         ]);
 
         $objType->assertValid();
 
-        self::assertSame(Type::string(), $objType->getField('f')->getType());
+        self::assertSame(Type::string(), $objType->getField($fieldName)->getType());
+    }
+
+    public function testAllowsInputTypeWhichDefinesFieldsWithLazyTypes(): void
```
vhenzl, line 770: Most input object tests are around line 1600, maybe to move these down there as well?

src/Type/Definition/InputObjectType.php
spawnia: In terms of opcodes, a method call is vastly more expensive than the single opcode of `isset()`.

The amount of code is reduced by putting the check inside the method, but I don't necessarily think the semantics are better. I generally prefer *does the thing need doing? -> do the thing* to *do the thing -> thing does nothing if it does not need doing*. It is somewhat comparable to having a method accept `null` as an argument and bail if it is passed - I would rather have the method be pure and check for `null` before.

simPod: Yup, this is where we differ and we keep stumbling upon this over and over (😄) as I prefer to encapsulate checks within the method so I don't have to worry about them every time I call it. Though I get the approach and accent on performance in terms of this library.

tests/Type/DefinitionTest.php
simPod: I just checked and they are also here :/ starting with L716. I'd postpone it into another PR.

simPod (commented):
🚢 ?

tests/Type/DefinitionTest.php
spawnia: I think I was able to put them in some kind of coherent order now 😅

vhenzl: Right, sure. I got some ideas about testing, how to improve it. I'll write it down.

PR #947, opened 2021-09-27
Restore support for Non-null variable types

simPod:
There was a huge change in #697 and tests did not cover it.

### DIFF
--- a/tests/Language/PrinterTest.php
+++ b/tests/Language/PrinterTest.php
@@ -88,6 +88,22 @@ class PrinterTest extends TestCase
         self::assertEquals($expected, Printer::doPrint($mutationAstWithArtifacts));
     }
 
+    public function testCorrectlyPrintsOpsWithNonNullableVariableArgument(): void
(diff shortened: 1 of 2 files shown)


tests/Language/PrinterTest.php
```diff
@@ -86,6 +86,18 @@ public function testCorrectlyPrintsOpsWithoutName(): void
...
+            'query ($var: ID!) { a(arg: $var) { __typename } }
+'
+        );
+        $expected                 = 'query ($var: ID!) {
+  a(arg: $var) {
+    __typename
+  }
+}
+';
+        self::assertEquals($expected, Printer::doPrint($queryWithNonNullVariable));
```
vhenzl, line 100: Could you move it to a separated test method without `@see it` annotation so that's clear it's an extra test not originating from `graphql-js`? It's to make sure it's not removed later when `PrinterTest.php` synced with `printer-test.ts`. Thanks.

PR #941, opened 2021-09-17
Align `SchemaPrinterTest` with `graphql-js` v15.5

vhenzl:
Aligns `SchemaPrinterTest.php` with `printSchema-test.js` from the reference implementation v15.5 as much as possible.

Relevant changes in `graphql-js`:
- https://github.com/graphql/graphql-js/commit/34b7ced883ff813a1d2d13c5fd58ba8158d69b60
- https://github.com/graphql/graphql-js/commit/4fcef75248a3db75da6080ed889ea2ce26a29928
- https://github.com/graphql/graphql-js/commit/ea089e016f1526fb5b5137b5f8a8739ec17b322c

TBD:
- [ ] different order of printed types (interfaces, unions)
- [ ] outdated introspection
- [ ] schema with description

Depends on #940, conflicts with #938 (shouldn't matter much but probably better merge #938 first)

PR #949, opened 2021-09-30
Document usage of nowdoc over dedent() in CONTRIBUTING.md

spawnia:
@vhenzl nowdoc looks great in #941


vhenzl:
Cool. Maybe also mention what nowdoc "tag" should be used for consistency? I see you changed `'GQL'` to `'GRAPHQL'` in a6985e43d913edb39d9c5a1efa3fb55d73cf9427. Any particular reason? Seems to me like unnecessary extra characters to type…

spawnia:
My impression is that `GQL` is not universally used as an abbreviation for GraphQL, and is also easily conflated with https://www.gqlstandards.org.

vhenzl:
Seem common to me, at least in JS worlds (for example [`gql` template literal tag](https://www.npmjs.com/package/graphql-tag)). But I'm easy…

PR #956, opened 2021-10-05
Use PHPStan type alias for options in `SchemaPrinter`

vhenzl:
This is a cherry-pick of @spawnia's changes from #941 that utilises `@phpstan-type` for options in  `SchemaPrinter`.

This PR is also to document the official intention of using PHPStan's [local type aliases](https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases) (`@phpstan-type` and `@phpstan-import-type`) and inherently [array shapes](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes) in the `graphql-php`'s codebase.

Both `@phpstan-type` and `@phpstan-import-type` are [supported](https://github.com/vimeo/psalm/commit/39e61ae942be9937bc877d2c2c69ba905507613f) by Psalm but there is no support yet in PHPStorm (vote in [WI-55929](https://youtrack.jetbrains.com/issue/WI-55929) and [WI-55928](https://youtrack.jetbrains.com/issue/WI-55928)).

PHPStorm's support for array shapes is currently [limited to one-line definitions only](https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-release/#array_shapes_phpdoc_syntax_support) (vote in [WI-56038](https://youtrack.jetbrains.com/issue/WI-56038)).

It's also important to note that array shapes in PHPStan are rather primitive; they can't be used for more advanced typing. Probably the biggest limitation is that two [array shapes can't be merged](https://github.com/phpstan/phpstan/issues/3931#issuecomment-921752417). Another important missing feature is support for additional arbitrary keys (an equivalent of TypeScript's `type X = {[k: string]: any}`). Both problems are related to phpstan/phpstan#4703.

Because of these limitations, and to support other tools and IDEs that don't understand `@phpstan-type`, it might be good to use `@phpstan-return` and `@phpstan-param` when referencing `@phpstan-type` (or using array shapes?) and `@return` and `@param` as a fallback. At least in the public API. For example:

```php
/**
 * @phpstan-type Options array{commentDescriptions?: bool}
 */
class SchemaPrinter
{
    /**
     * @phpstan-param Options $options
     * @param array<string, bool> $options
     *
     * @api
     */
    public static function doPrint(Schema $schema, array $options = []): string
    {
        // ...
    }
}
```

What do you think @spawnia and @simPod?

### DIFF
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -37,16 +37,18 @@ use function str_replace;
 use function strlen;
 
 /**
- * Given an instance of Schema, prints it in schema definition language.
+ * Prints the contents of a Schema in schema definition language.
+ *
+ * @phpstan-type Options array{commentDescriptions?: bool}
+ *    Available options:
+ *    - commentDescriptions:
+ *        Provide true to use preceding comments as the description.
+ *        This option is provided to ease adoption and will be removed in v16.
  */
 class SchemaPrinter
 {
     /**
-     * @param array<string, bool> $options
-     *    Available options:
-     *    - commentDescriptions:
-     *        Provide true to use preceding comments as the description.
-     *        This option is provided to ease adoption and will be removed in v16.
+     * @param Options $options
      *
      * @api
      */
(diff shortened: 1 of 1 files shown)


spawnia:
I agree with using `@phpstan-param` when using type aliases.

PR #971, opened 2021-10-10
Extract common functionality for types with interfaces

spawnia:


simPod:
This now gives with psalm

> ERROR: PropertyNotSetInConstructor - src/MyType.php:11:13 - Property My\Ns\MyType::$interfaces is not defined in constructor of My\Ns\MyType or in any private or final methods called in the constructor (see https://psalm.dev/074)
final class MyType extends ObjectType

This is my current solution which narrows ignoring this error to a directory. Though it ignores also other files in that directory, I don't think there's much more that can be done about it in user land code.

```xml
        <!-- graphql-php uses property lazy-loading -->
        <PropertyNotSetInConstructor>
            <errorLevel type="suppress">
                <directory name="src/somepath/Schema/Type" />
            </errorLevel>
        </PropertyNotSetInConstructor>
```

Issue #964, opened 2021-10-10
Exclude unused standard types from the schema


spawnia:
The current implementation of `Schema` always adds standard scalar types to the schema, even if they are not used. This differs from the reference implementation, which has added quite a few test cases that rely on missing standard scalars.

We should align our behaviour with the specification, see http://spec.graphql.org/draft/#sec-Scalars.Built-in-Scalars:

> If a built-in scalar type is not referenced anywhere in a schema (there is no field, argument, or input field of that type) then it must not be included.

PR #974, opened 2021-10-15
Exclude unused standard types from the schema

spawnia:
Improves our behaviour in regards to https://github.com/webonyx/graphql-php/issues/964, except `Schema::getType()`


spawnia:
Thanks @vhenzl for bringing that up. I noticed that difference and changed the implementation of a test to use `getTypeMap()` instead.

I documented the difference for now (see https://github.com/webonyx/graphql-php/issues/964#issuecomment-945969162) and plan to leave the issue open until we can resolve it fully. Unless you have an idea how to fix our implementation, I would like to go ahead with this PR. It does at least fix the externally observable behaviour in regards to introspection.

PR #975, opened 2021-10-15
IncludeDirective `include(if: $bool)` should not include when `$bool = null`

ruudk:
I noticed something weird.

Given you have the following query:

```graphql
query MyQuery($extra: Boolean) {
  a @include(if: $extra)
  b
}
```

`a` should only be returned when `{ extra: true }` but it turns out that it's also returned when `{ extra: null }`.

### DIFF
--- a/tests/Executor/DirectivesTest.php
+++ b/tests/Executor/DirectivesTest.php
@@ -303,4 +303,55 @@ class DirectivesTest extends TestCase
             ],
         ];
     }
+
+    /**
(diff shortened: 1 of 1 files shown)


ruudk:
Added a test case for the Skip directive, there it works as expected. It skips only when `extra = true`, and not when `extra = null` or `extra = false`.

ruudk:
Checked with V14 and it doesn't allow using `Boolean` indeed. Great.

PR #995, opened 2021-10-30
Extract check for unique operation types into separate rule

vhenzl:
Introduces `UniqueOperationTypes` validation rule.

Corresponding change in the reference implementation: https://github.com/graphql/graphql-js/commit/9b7a8af43fc0865a01df5b5a084f37bbb8680ef8

### DIFF
--- a/src/Validator/Rules/UniqueOperationTypes.php
+++ b/src/Validator/Rules/UniqueOperationTypes.php
@@ -5,14 +5,12 @@ declare(strict_types=1);
 namespace GraphQL\Validator\Rules;
 
 use GraphQL\Error\Error;
-use GraphQL\Language\AST\Node;
 use GraphQL\Language\AST\NodeKind;
 use GraphQL\Language\AST\SchemaDefinitionNode;
 use GraphQL\Language\AST\SchemaTypeExtensionNode;
 use GraphQL\Language\Visitor;
 use GraphQL\Language\VisitorOperation;
 use GraphQL\Validator\SDLValidationContext;
-use TypeError;
 
 /**
  * Unique operation types
@@ -36,11 +34,7 @@ class UniqueOperationTypes extends ValidationRule
         /**
          * @param SchemaDefinitionNode|SchemaTypeExtensionNode $node
          */
-        $checkOperationTypes = static function (Node $node) use ($context, &$definedOperationTypes, $existingOperationTypes): VisitorOperation {
-            if (! $node instanceof SchemaDefinitionNode && ! $node instanceof SchemaTypeExtensionNode) {
-                throw new TypeError('Expected $node to be ' . SchemaDefinitionNode::class . ' or ' . SchemaTypeExtensionNode::class . '.');
-            }
-
+        $checkOperationTypes = static function ($node) use ($context, &$definedOperationTypes, $existingOperationTypes): VisitorOperation {
             foreach ($node->operationTypes as $operationType) {
                 $operation                   = $operationType->operation;
                 $alreadyDefinedOperationType = $definedOperationTypes[$operation] ?? null;
--- a/src/Validator/Rules/UniqueOperationTypes.php
+++ b/src/Validator/Rules/UniqueOperationTypes.php
@@ -5,12 +5,14 @@ declare(strict_types=1);
 namespace GraphQL\Validator\Rules;
 
 use GraphQL\Error\Error;
+use GraphQL\Language\AST\Node;
 use GraphQL\Language\AST\NodeKind;
 use GraphQL\Language\AST\SchemaDefinitionNode;
 use GraphQL\Language\AST\SchemaTypeExtensionNode;
 use GraphQL\Language\Visitor;
 use GraphQL\Language\VisitorOperation;
 use GraphQL\Validator\SDLValidationContext;
+use TypeError;
 
 /**
  * Unique operation types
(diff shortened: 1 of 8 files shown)


src/Validator/Rules/UniqueOperationTypes.php
```diff
@@ -0,0 +1,71 @@
...
+                'query' => $schema->getQueryType(),
+                'mutation' => $schema->getMutationType(),
+                'subscription' => $schema->getSubscriptionType(),
+            ]
+            : [];
+
+        /**
+         * @param SchemaDefinitionNode|SchemaTypeExtensionNode $node
+         */
+        $checkOperationTypes = static function ($node) use ($context, &$definedOperationTypes, $existingOperationTypes): VisitorOperation {
```
spawnia, line 37:
```suggestion
        $checkOperationTypes = static function (Node $node) use ($context, &$definedOperationTypes, $existingOperationTypes): VisitorOperation {
```

Probably requires the hint above to use intersection types with `&Node`

src/Validator/Rules/UniqueOperationTypes.php
```diff
@@ -0,0 +1,71 @@
...
+            $operationTypesNodes = $node->operationTypes;
+            foreach ($operationTypesNodes as $operationType) {
+                $operation                    = $operationType->operation;
+                $alreadyDefinedOperationType  = $definedOperationTypes[$operation] ?? null;
+                $alreadyExistingOperationType = $existingOperationTypes[$operation] ?? null;
+
+                if ($alreadyExistingOperationType !== null) {
```
spawnia, line 44:
```suggestion
                if (isset($existingOperationTypes[$operation])) {
```

src/Validator/Rules/UniqueOperationTypes.php
```diff
@@ -0,0 +1,71 @@
...
+                'subscription' => $schema->getSubscriptionType(),
+            ]
+            : [];
+
+        /**
+         * @param SchemaDefinitionNode|SchemaTypeExtensionNode $node
+         */
+        $checkOperationTypes = static function ($node) use ($context, &$definedOperationTypes, $existingOperationTypes): VisitorOperation {
+            $operationTypesNodes = $node->operationTypes;
+            foreach ($operationTypesNodes as $operationType) {
```
spawnia, line 39:
```suggestion
            foreach ($node->operationTypes as $operationType) {
```

tests/Validator/UniqueOperationTypesTest.php
```diff
@@ -0,0 +1,463 @@
...
+use GraphQL\Validator\Rules\UniqueOperationTypes;
+
+class UniqueOperationTypesTest extends ValidatorTestCase
+{
+    private function expectSDLErrors(string $sdlString, ?Schema $schema, array $errors): void
+    {
+        $this->expectSDLErrorsFromRule(new UniqueOperationTypes(), $sdlString, $schema, $errors);
+    }
+
+    // Validate: Unique operation types
```
spawnia, line 18: @simPod recently brought it up, and I am inclined to agree that we should put this in the PHPDoc of the first test of the group in order to have it be sticky in a structural sense.

spawnia (commented):
Looks solid overall, can be micro-optimized a bit

tests/Validator/UniqueOperationTypesTest.php
vhenzl: Something like this?
```php
/**
 * @see describe('Validate: Unique operation types')
 * @see it('no schema definition')
 */
```

spawnia: Looks good to me!

src/Validator/Rules/UniqueOperationTypes.php
vhenzl: No `&Node` needed as `SchemaDefinitionNode` and `SchemaTypeExtensionNode` extends `Node`.

But the problem is, PHPStan doesn't understand `@params` on closures (https://github.com/phpstan/phpstan/issues/3770), therefore, without the `Node` type-hint, `$node` is `mixed`, with the type-hint, it's only `Node` resulting in `Access to an undefined property GraphQL\Language\AST\Node::$operationTypes.` error.

A possible solution is to check types explicitly:

```php
/**
 * @param SchemaDefinitionNode|SchemaTypeExtensionNode $node
 */
$checkOperationTypes = static function (Node $node) use ($context, &$definedOperationTypes, $existingOperationTypes): VisitorOperation {
    if (! $node instanceof SchemaDefinitionNode && ! $node instanceof SchemaTypeExtensionNode) {
        throw new TypeError('Expected $node to be ' . SchemaDefinitionNode::class . ' or ' . SchemaTypeExtensionNode::class . '.');
    }

    // ...
}
```

Once PHPStan starts understanding PhpDoc above closures, it can point this `if` as extraneous and we can remove it.

spawnia: You are right, `Node` is not needed. Thanks for researching that thoroughly.

I would rather forgo strict type safety if the alternative means worsening runtime performance. We have tests that ensure correctness, and its not like the code is going to run into a catastrophic issue like a memory leak if the types don't line up.

Let's just keep the code as is, that way we are ready for when PHPStan understands closure param hints. When we switch to PHP 8.1, we can just move to an actual type hint.

spawnia:
Can you add a CHANGELOG entry to mention the introduction of a new rule?

vhenzl:
@spawnia It's ready.

spawnia (approved):

Issue #954, opened 2021-10-05
Printed schema should preserve original order of types


vhenzl:
The reference implementation v15 changed the order in which types are printed. Types are no longer ordered alphabetically; instead, the printed schema preserves the original user-provided order of types. It's a step towards reversible schema build (`sdl === printSchema(buildSchema(sdl))`).

The relevant commits in `graphql-js` are graphql/graphql-js@68a0818b44f96e2ff18eb188eab815084ad8f211 and graphql/graphql-js@057510bb76b70f0d2ba4487add54f5d53f4c7d8e.

I'm currently not confident enough to refactor `Schema` which differs significantly from the JS version. Nor I'm sure it's possible to implement it at all because of type lazy loading. So this issue is to note the difference between `graphql-php` and `graphql-js` for now.

PR #996, opened 2021-10-31
SchemaPrinter: Directives support 🤩

LastDragon-ru:
## Features: 

* Seems no API breaking changes (but output for long lines may be a bit different, see below);
* By default it has the same behavior as before; directives printing should be enabled explicitly via `printDirectives` option;
* Supported location: `ARGUMENT_DEFINITION`, `ENUM`, `ENUM_VALUE`, `FIELD_DEFINITION`, `INPUT_FIELD_DEFINITION`, `INPUT_OBJECT`, `INTERFACE`, `OBJECT`, `SCALAR`, `UNION`;
* Long directives (= with a lot of arguments) will be split into multiple lines
* Long list of arguments will be split into multiple lines 🆕
* Long List values (`[1,2,3]`) will be split into multiple lines 🆕
* Better separation for children with long-line/description 🆕

```graphql
# Example
input InputB
@test(
  value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
) {
  a: [String] = [
    "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
    "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
    "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
  ]
}

# Before
enum Color {
  RED

  """Not a creative color"""
  GREEN
  BLUE
}

#After
enum Color {
  RED

  """Not a creative color"""
  GREEN

  BLUE
}
```

Closes: #552, #875

### DIFF
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -46,7 +46,6 @@ use function ksort;
 use function ltrim;
 use function mb_strlen;
 use function mb_strpos;
-use function self;
 use function sprintf;
 use function str_replace;
 use function strlen;
@@ -531,7 +530,7 @@ class SchemaPrinter
         // Enabled?
         $filter = $options['printDirectives'] ?? null;
 
-        if (!$filter || !is_callable($filter)) {
+        if (!is_callable($filter)) {
             if ($type instanceof EnumValueDefinition || $type instanceof FieldDefinition) {
                 return static::printDeprecated($type);
             }
@@ -542,12 +541,11 @@ class SchemaPrinter
         // AST Node available and has directives?
         $node = $type->astNode;
 
-        if (!$node) {
+        if ($node === null) {
             return '';
         }
 
         // Print
-        /** @var array<string> $directives */
         $length = 0;
         $directives = [];
 
@@ -562,7 +560,7 @@ class SchemaPrinter
                 // empty
             }
 
-            if ($string) {
+            if ($string !== null) {
                 $length = $length + mb_strlen($string);
                 $directives[] = $string;
             }
@@ -571,7 +569,7 @@ class SchemaPrinter
         // Multiline?
         $serialized = '';
 
-        if ($directives) {
+        if (count($directives) > 0) {
             $delimiter  = static::isLineTooLong($length) ? "\n{$indentation}" : ' ';
             $serialized = $delimiter.implode($delimiter, $directives);
         }
@@ -632,7 +630,7 @@ class SchemaPrinter
     {
         $block = '';
 
-        if ($lines) {
+        if (count($lines) > 0) {
             $forceMultiline = $forceMultiline || static::isLineTooLong(array_sum(array_map('mb_strlen', $lines)));
 
             if ($forceMultiline) {
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -334,6 +334,7 @@ class SchemaPrinter
         return static::printDescription($options, $type) .
             sprintf('type %s', $type->name) .
             self::printImplementedInterfaces($type) .
+            static::printTypeDirectives($type, $options) .
             static::printFields($options, $type);
     }
 
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -494,8 +494,8 @@ class SchemaPrinter
     {
         $fields = array_values($type->getFields());
         $fields = array_map(
-            static function ($f, $i) use ($options): string {
-                return static::printDescription($options, $f, '  ', ! $i) . '  ' . static::printInputValue($f);
+            static function (InputObjectField $f, $i) use ($options): string {
+                return static::printInputObjectField($f, $options, '  ', ! $i);
             },
             $fields,
             array_keys($fields)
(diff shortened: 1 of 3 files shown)


vhenzl:
Some of my thoughts from the first impression:

- This PR seems to be solving more unrelated problems – directive printing and long lines/args/lists printing. Breaking it into more scoped PRs would make it easier to understand.
- As for the main topic, directive printing, I believe arguments from #552 still holds. It's a complex problem and this library should wait until it is resolved in the reference implementation.
- Regarding long lines/args/lists wrapping, what problem is it trying to solve? How does it match the reference implementation and GraphQL spec? Ideally, schema building and printing should be reversible, i.e. `sdl === printSchema(buildSchema(sdl))`. The library doesn't work like that on 100% yet (see #954), but there have been some other changes done in this direction (e.g. #938 removes long description wrapping). These changes seem to go against this effort.

LastDragon-ru:
> This PR seems to be solving more unrelated problems – directive printing and long lines/args/lists printing. 

Nope, it is a free bonus 😁 But you are welcome to split it if you want.

> Regarding long lines/args/lists wrapping, what problem is it trying to solve?

The same problem as the multiline description - readability. This is impossible to read:

```
input InputB
@test(value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", a: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa") @deprecated(reason: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa") {
  a: [String] = ["aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",  "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",  "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"]
}
```

> As for the main topic, directive printing, I believe arguments from Include directives in SchemaPrinter output #552 still holds. It's a complex problem and this library should wait until it is resolved in the reference implementation.

I highly disagree:

1) I don't see any arguments why directives should not be printed in related issues. Yes, some of the directives probably should not be included in the output, and this PR allows you to control which directives should be printed;
2) Directives are very important: for example, we use directives to check permissions and they are required for UI team, right now we are using descriptions but it sucks.
4) They are extremely useful to check schema after automatic transformation (more related to Lighthouse), right now there is no way to check added directives and this is a big problem.
3) Moreover, they are already used by many others: GitHub GraphQL API, Lighthouse Federated printer (and Apollo), etc;
5) Also, as I can see the upstream issue is dead.

> Ideally, schema building and printing should be reversible, i.e. sdl === printSchema(buildSchema(sdl))

Parser strips whitespace while parsing so I guess that is not possible... But you are right, my second goal - generate GraphlQL that can be read and compared by humans even if the schema was generated by robots (= without formatting). Because from my point of view if readability doesn't matter there are no reasons to add LFs and paddings at all.

Also, all original tests except two related to enum members are passed, the difference will be only in (these cases are not covered by original tests):

```graphql
test(aaaaa: String, bbbb: String, cccc: String, dddddddddcccc: String, fsdfsdfsdfcccc: String, dsafsdcccc: String, sdfdfsdfcccc: String, sdfsdfsdcccc: String, sdfsdfsdfsfsdfcccc: String, sdfsfsdfdfsdfcccc: String): String;
```

That will be split into much more readable form:
```graphql
test(
  aaaaa: String
  bbbb: String
  cccc: String
  dddddddddcccc: String
  fsdfsdfsdfcccc: String
  dsafsdcccc: String
  sdfdfsdfcccc: String
  sdfsdfsdcccc: String
  sdfsdfsdfsfsdfcccc: String
  sdfsfsdfdfsdfcccc: String
): String;
```

vhenzl:
Readability can be an issue, the last example is a good one. There actually has been some work done on that topic in the reference implementation (https://github.com/graphql/graphql-js/pull/2797). But in general, I would argue that the library shouldn't do any sophisticated formatting. Instead, tools like [Prettier](https://prettier.io/) can be used to format printed GraphQL if you need it. Anyway, it's unrelated to directive printing.

spawnia:
See https://github.com/graphql/graphql-js/pull/3362 for an attempt to get support into the reference implementation first. Let's see what they say before continuing here.

LastDragon-ru:
PR is outdated and I have no time to update it. If someone need directives maybe you can try to use [`SchemaPrinter`](https://github.com/LastDragon-ru/lara-asp/tree/master/packages/graphql#schemaprinter--testing-assertions) from `lara-asp` package.

Issue #997, opened 2021-10-31
`Schema::getType()` shoud always return `null` for non-existent type


vhenzl:
There is an inconsistency in how `Schema::getType()` behaves depending on how the schema is created.

If the schema is created using type definitions, it correctly returns `null` for a type that doesn't exist in the schema:

```php
$schema = new Schema([
    'types' => [
        new ObjectType([
            'name' => 'Foo',
            'fields' => ['id' => Type::id()],
        ]),
    ],
]);

$foo = $schema->getType('Foo');
$bar = $schema->getType('Bar'); // ✅ returns null as expected
```

If the schema is built from SDL, it throws `Unknown type: "Bar".` error instead or returning `null`:

```php
$schema = BuildSchema::build('type Foo { id: ID }');

$foo = $schema->getType('Foo');
$bar = $schema->getType('Bar'); // 💥 throws error
```

This has a negative impact for example on SDL validation rules.

For reference, `graphql-js` always returns `undefined` for a non-existent type, regardless of how the schema was created.

I'm trying to wrap my head around how the lazy type loading works (it seems related, at least partially). Any suggestions on how to fix this without a significant rewrite of `BuildSchema`, `SchemaExtender` and `ASTDefinitionBuilder` are welcome. (However, I'm hitting various problems with schema building & extending from different directions, so I'm more and more convinced that rewriting them in accordance with the reference implementation is actually the only reasonable solution.)

spawnia:
Relates to https://github.com/webonyx/graphql-php/issues/964#issuecomment-945969162

> I'm trying to wrap my head around how the lazy type loading works

It all happens in `Schema`. One important part is where the actual lazy loading happens: https://github.com/webonyx/graphql-php/blob/647e2cf5ea7bf83a5efc3eb6bdcc040d6c524428/src/Type/Schema.php#L295-L306

However, loading types simply on demand would be problematic. Suppose a schema is constructed from SDL and a lazy loader is provided, how would validation/introspection be able to load all existing types? That's why there is an additional method that allows loading all types, by providing a callable to the `types` option: https://github.com/webonyx/graphql-php/blob/647e2cf5ea7bf83a5efc3eb6bdcc040d6c524428/src/Type/Schema.php#L150-L168

vhenzl:
Sorry for the long silence; haven't got much time recently…

#1038 that allowed type loaders to return null was a step in the correct direction.

[...]

PR #998, opened 2021-10-31
Extract check for unique type names into separate rule

vhenzl:
Introduces `UniqueTypeNames` validation rule.

Corresponding change in the reference implementation: https://github.com/graphql/graphql-js/commit/257797a0ebdddd3da6e75b7c237fdc12a1a7c75a

Tests are failing because of #997.


src/Utils/SchemaExtender.php
```diff
@@ -576,16 +576,7 @@ public static function extend(
...
-                    : null;
-
-                $type = $schema->getType($typeName);
-
-                if (null !== $type) {
-                    throw new Error('Type "' . $typeName . '" already exists in the schema. It cannot also be defined in this type definition.', [$def]);
-                }
-
+                $typeName = $def->name->value;
                 $typeDefinitionMap[$typeName] = $def;
```
spawnia, line 580:
```suggestion
                $typeDefinitionMap[$def->name->value] = $def;
```

tests/Validator/UniqueTypeNamesTest.php
```diff
@@ -0,0 +1,210 @@
...
+{
+    /**
+     * @param array<int, array<string, mixed>> $errors
+     */
+    private function expectSDLErrors(string $sdlString, ?Schema $schema, array $errors): void
+    {
+        $this->expectSDLErrorsFromRule(new UniqueTypeNames(), $sdlString, $schema, $errors);
+    }
+
+    // Validate: Unique type names
```
spawnia, line 21: This should rather be attached to the first test case of the group, otherwise it can get lost during structural editing

spawnia:
Nicely done, thank you

PR #1000, opened 2021-11-02
Add validation rule UniqueEnumValueNames

spawnia:
https://github.com/graphql/graphql-js/commit/0737212567a394caccc92f26a0705869f9e4ee3b

### DIFF
--- a/src/Validator/Rules/UniqueEnumValueNames.php
+++ b/src/Validator/Rules/UniqueEnumValueNames.php
@@ -8,7 +8,6 @@ use GraphQL\Error\Error;
 use GraphQL\Language\AST\EnumTypeDefinitionNode;
 use GraphQL\Language\AST\EnumTypeExtensionNode;
 use GraphQL\Language\AST\EnumValueNode;
-use GraphQL\Language\AST\Node;
 use GraphQL\Language\AST\NodeKind;
 use GraphQL\Language\Visitor;
 use GraphQL\Language\VisitorOperation;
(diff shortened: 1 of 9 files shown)


spawnia:
@vhenzl working on this, just need to finish tests

src/Validator/Rules/UniqueEnumValueNames.php
```diff
@@ -0,0 +1,71 @@
...
+{
+    public function getSDLVisitor(SDLValidationContext $context): array
+    {
+        /** @var array<string, array<string, EnumValueNode>> $knownValueNames */
+        $knownValueNames = [];
+
+        /**
+         * @param (Node&EnumTypeDefinitionNode)|(Node&EnumTypeExtensionNode) $enum
+         */
+        $checkValueUniqueness = static function ($enum) use ($context, &$knownValueNames): VisitorOperation {
```
vhenzl, line 28:
```suggestion
        $checkValueUniqueness = static function (Node $enum) use ($context, &$knownValueNames): VisitorOperation {
```
Native type here?

tests/Validator/UniqueEnumValueNamesTest.php
```diff
@@ -0,0 +1,274 @@
...
+            new UniqueEnumValueNames(),
+            '
+      extend enum SomeEnum {
+        FOO
+      }
+      extend enum SomeEnum {
+        FOO
+      }
+        ',
+            BuildSchema::build('
```
vhenzl, line 260: I would prefer to build the schema before calling expect and pass `$schema` here. That way the test structure would be more similar to `graphql-js` and the code easier to compare. 

Applies also to the tests above.

vhenzl (commented):
Two suggestions. Looks good to me in general.

src/Validator/Rules/UniqueEnumValueNames.php
vhenzl: See https://github.com/webonyx/graphql-php/pull/995#discussion_r741482567

spawnia: I would feel safer with a native type, but am fine with relying on tests proving the correctness of this piece of code. I would rather not worsen performance.

simPod:
💯 0

PR #1002, opened 2021-11-03
Fix `Typed static property GraphQL\Executor\Promise\Adapter\SyncPromise::$queue must not be accessed before initialization`

simPod:


tests/Executor/Promise/SyncPromiseTest.php
```diff
@@ -425,4 +425,11 @@ public function testPendingPromiseThen(): void
         self::assertValidPromise($nextPromise3, null, 2, SyncPromise::FULFILLED);
         self::assertValidPromise($nextPromise4, null, 3, SyncPromise::FULFILLED);
     }
+
+    public function testRunEmptyQueue(): void
+    {
+        SyncPromise::runQueue();
+
+        self::assertTrue(true);
```
spawnia, line 433: This will fail in PHPStan Level 4.

simPod: Hm, I use to ignore this. There's no other way to count it towards coverage.

spawnia: See my edit. I can see how there is really nothing else to do here, but let's make it explicit and have a single place to `@phpstan-ignore`.

simPod: can do. I explicitly put this into phpstan config:

```neon
        # There's no other way to test-pass without assertions while counting it towards coverage https://github.com/sebastianbergmann/phpunit/issues/3016
        - "~Call to static method PHPUnit\\\\Framework\\\\Assert::assertTrue\\(\\) with true will always evaluate to true~"
```

spawnia: I would rather have a method with a name that makes it explicit what we are doing here. Otherwise someone else comes along with `assertNull(null)` and we need to ignore another thing.

Also, I think that ignore might be overly eager in instances where PHPStan has determined some dynamic result is always `true`. An explicit method let's us ignore that single line specifically.

simPod: > Also, I think that ignore might be overly eager in instances where PHPStan has determined some dynamic result is always true


The ignore message specifically constrains it to `assertTrue(true)`, nothing else.

Anyway I have added the method.

PR #1003, opened 2021-11-03
Use singletons for visitor operations

spawnia:


simPod (approved):
Since there's no state in those I guess it's fine.

PR #1004, opened 2021-11-04
Support non-JSON ServerRequestInterface

dsavina:
In #598, use of `ServerRequestInterface::getParsedBody()` has been replaced with inlined parsing of `RequestInterface::getBody()`.

Later, it has been reintroduced (cf. #715), but for JSON requests only.

Therefore, this Pull Request aims to restore use of `getParsedBody()` when available, in fallback case as well (commonly `application/x-www-form-urlencoded`).

### DIFF
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -525,11 +525,9 @@ class Helper
 
                 $this->assertJsonObjectOrArray($bodyParams);
             } else {
-                parse_str((string) $request->getBody(), $bodyParams);
-
-                if (! is_array($bodyParams)) {
-                    throw new RequestError('Unexpected content type: ' . Utils::printSafeJson($contentType[0]));
-                }
+                $bodyParams = $request instanceof ServerRequestInterface
+                    ? $request->getParsedBody()
+                    : $this->decodeContent((string) $request->getBody(), $contentType[0]);
             }
         }
 
@@ -553,6 +551,22 @@ class Helper
 
         if (json_last_error() !== JSON_ERROR_NONE) {
             throw new RequestError('Expected JSON object or array for "application/json" request, but failed to parse because: ' . json_last_error_msg());
+        }
+
+        return $bodyParams;
+    }
+
+    /**
+     * @return array<string, mixed>
+     *
+     * @throws RequestError
+     */
+    protected function decodeContent(string $rawBody, string $contentType): array
+    {
+        parse_str($rawBody, $bodyParams);
+
+        if (! is_array($bodyParams)) {
+            throw new RequestError('Unexpected content type: ' . Utils::printSafeJson($contentType));
         }
 
         return $bodyParams;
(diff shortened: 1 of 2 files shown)


spawnia:
Please add a test to ensure this actually fixes a reproducible issue and we do not cause a regression later.

src/Server/Helper.php
```diff
@@ -558,6 +556,22 @@ protected function decodeJson(string $rawBody)
         return $bodyParams;
     }
 
+    /**
+     * @return array
+     *
+     * @throws RequestError
+     */
+    protected function decodeContent(string $rawBody, string $contentType)
```
spawnia, line 564:
```suggestion
    protected function decodeContent(string $rawBody, string $contentType): array
```

src/Server/Helper.php
```diff
@@ -558,6 +556,22 @@ protected function decodeJson(string $rawBody)
         return $bodyParams;
     }
 
+    /**
+     * @return array
```
spawnia, line 560:
```suggestion
     * @return array<string, mixed>
```

dsavina:
Thanks for the feedback, I updated the code as requested.

I actually had an issue using Symfony's [`PsrHttpFactory`](https://github.com/symfony/psr-http-message-bridge/blob/main/Factory/PsrHttpFactory.php), which, for some reason, ends up providing a `ServerRequest` with an empty `body` but a valid `parsedBody`.

Although this could be perceived as an issue on Symfony's side, considering that, in the case of an URL-encoded body:
- Method `Helper::parseHttpRequest()` relies on global variable `$_POST` (i.e. already parsed)
- Test method `RequestParsingTest::parseRawFormUrlencodedRequest()` indicates that reading from `php://input` should be prohibited

, I guess this behavior is nevertheless considered pretty standard, isn't it?

Lastly, in order to support these inconsistencies between `body` and `parsedBody`, I could propose this instead:
```php
                if ($request instanceof ServerRequestInterface) {
                    $bodyParams = $request->getParsedBody();
                }

                $bodyParams ??= $this->decodeContent((string) $request->getBody(), $contentType[0]);
```
Would that be preferable?

spawnia:
> Would that be preferable?

If you can create a test case that fails with the current implementation and passes with that one, sure.

spawnia:
Thanks for the PR and the thorough tests

dsavina:
Hello @spawnia, do you have some insight regarding when this could be part of a new release, by any chance?

We will need this fix for our package [thecodingmachine/graphqlite-bundle](https://github.com/thecodingmachine/graphqlite-bundle). If it's not something we can expect in the near future (which I would totally understand, I don't mean to push my personal agenda), I guess we could implement some temporary fix on our side. I'd still prefer not to if it's something we can avoid, though.

spawnia:
Backported and releaed with https://github.com/webonyx/graphql-php/releases/tag/v14.11.2

dsavina:
Nice, thank you very much!

PR #1005, opened 2021-11-05
Make server related code fit for PHPStan level 4

spawnia:


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Move complexity related code from `FieldDefinition` to `QueryComplexity`
 - Exclude unused standard types from the schema
 - Require lazy type loader to return `Type` directly without an intermediary callable
+- Rename `ServerConfig` option `persistentQueryLoader` to `persistedQueryLoader`
```
simPod, line 27: Must admit I don't know where this is coming from and maybe is not really related to phpstan?

Is it this https://www.apollographql.com/docs/apollo-server/performance/apq/?

src/Server/OperationParams.php
```diff
@@ -84,57 +106,69 @@ public static function create(array $params, bool $readonly = false): OperationP
...
-        // Some parameters could be provided as serialized JSON.
-        foreach (['extensions', 'variables'] as $param) {
-            if (! is_string($params[$param])) {
-                continue;
-            }
-
-            $tmp = json_decode($params[$param], true);
-            if (json_last_error() !== JSON_ERROR_NONE) {
-                continue;
+        array_walk($params, static function (&$value): void {
```
simPod, line 109: array_map is more appropriate here

src/Server/OperationParams.php
```diff
@@ -84,57 +106,69 @@ public static function create(array $params, bool $readonly = false): OperationP
...
+        ) {
             $instance->queryId = $instance->extensions['persistedQuery']['sha256Hash'];
         }
 
         return $instance;
     }
 
     /**
-     * @return mixed
+     * @deprecated
```
simPod, line 137: This was not deprecated so we might need to add it to changelog? I'd also add some note on why deprecation happens.

src/Server/OperationParams.php
spawnia: I explicitly want to mutate the array, so `array_map` is not an option. I just noticed a good old `foreach` works as well though.

src/Server/StandardServer.php
```diff
@@ -75,6 +75,7 @@ public function __construct($config)
             $config = ServerConfig::create($config);
         }
 
+        // @phpstan-ignore-next-line necessary until we can use proper union types
```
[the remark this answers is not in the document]
spawnia: I don't like adding to the baseline, because that does not allow adding a comment. Moving it over to the config file is pretty cumbersome and still does not allow pinpointing the exact location of the error.

There are cases for it, such as errors we can never fix (some variable property accesses), or bugs in PHPStan. In this case, I prefer the annotation. I think the risk of introducing a bug here and accidentally not noticing is very slim.

CHANGELOG.md
spawnia: Not related to PHPStan at all, just felt like making a breaking change 😉  The previously chosen name is inconsistent with the usual terminology, and I think this library is better off in the long run with having the "correct" name in place.

src/Server/OperationParams.php
spawnia: Given we are working towards a new major version anyways, I just removed them and put that into the changelog. I don't like getters, especially ones that prevent direct access to an array. It prevents so many useful things that can be done with arrays.

simPod (approved):

PR #1007, opened 2021-11-12
Call `EnumType::parseLiteral` from `AST` to support PHP enums/objects

ruudk:
Somehow, this call was missing. It is defined in `EnumType` but was never used.

To make it possible to convert enum string values to PHP objects (or enums in PHP 8.1) we need
to call it.

This allows developers to override the `parseLiteral` method and do the conversion there.

See the example in the test.

### DIFF
--- a/tests/Type/EnumTypeTest.php
+++ b/tests/Type/EnumTypeTest.php
@@ -9,12 +9,15 @@ use GraphQL\Error\DebugFlag;
 use GraphQL\GraphQL;
 use GraphQL\Language\SourceLocation;
 use GraphQL\Tests\PHPUnit\ArraySubsetAsserts;
+use GraphQL\Tests\Type\TestClasses\OtherEnum;
+use GraphQL\Tests\Type\TestClasses\OtherEnumType;
 use GraphQL\Type\Definition\EnumType;
 use GraphQL\Type\Definition\ObjectType;
 use GraphQL\Type\Definition\Type;
 use GraphQL\Type\Introspection;
 use GraphQL\Type\Schema;
 use PHPUnit\Framework\TestCase;
+use function assert;
 use function count;
 use function is_array;
 
@@ -53,6 +56,8 @@ class EnumTypeTest extends TestCase
                 'THREE',
             ],
         ]);
+
+        $otherEnum = new OtherEnumType();
 
         $Complex1 = [
             'someRandomFunction' => static function () : void {
@@ -112,6 +117,28 @@ class EnumTypeTest extends TestCase
                         if (isset($args['fromValue'])) {
                             return $args['fromValue'];
                         }
+                    },
+                ],
+                'otherEnum'  => [
+                    'type'    => Type::string(),
+                    'args'    => [
+                        'from'  => ['type' => $otherEnum],
+                    ],
+                    'resolve' => static function ($rootValue, $args) : string {
+                        assert($args['from'] instanceof OtherEnum);
+
+                        return 'Hello enum ' . $args['from']->getValue();
+                    },
+                ],
+                'returnOtherEnum'  => [
+                    'type'    => $otherEnum,
+                    'args'    => [
+                        'from'  => ['type' => $otherEnum],
+                    ],
+                    'resolve' => static function ($rootValue, $args) : ?OtherEnum {
+                        assert($args['from'] instanceof OtherEnum);
+
+                        return $args['from'];
                     },
                 ],
                 'colorInt'    => [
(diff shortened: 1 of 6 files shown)


tests/Type/EnumTypeTest.php
```diff
@@ -114,6 +118,32 @@ public function setUp() : void
...
                     },
                 ],
+                'otherEnum'  => [
+                    'type'    => Type::string(),
+                    'args'    => [
+                        'fromEnum'  => ['type' => $otherEnum],
+                    ],
+                    'resolve' => static function ($rootValue, $args) : string {
+                        if (isset($args['fromEnum']) && $args['fromEnum'] instanceof OtherEnum) {
+                            return $args['fromEnum']->__toString();
```
spawnia, line 128: `$args['fromEnum']` is duplicated thrice

tests/Type/EnumTypeTest.php
```diff
@@ -114,6 +118,32 @@ public function setUp() : void
...
+                        return 'fromEnum is not provided';
+                    },
+                ],
+                'returnOtherEnum'  => [
+                    'type'    => $otherEnum,
+                    'args'    => [
+                        'fromEnum'  => ['type' => $otherEnum],
+                    ],
+                    'resolve' => static function ($rootValue, $args) : ?OtherEnum {
+                        if (isset($args['fromEnum']) && $args['fromEnum'] instanceof OtherEnum) {
```
spawnia, line 140: `$args['fromEnum']` is duplicated thrice

tests/Type/EnumTypeTest.php
```diff
@@ -606,4 +636,22 @@ public function testAllowsSimpleArrayAsValues() : void
...
+    {
+        $q = '{
+            otherEnum(fromEnum: ONE)
+            returnOtherEnum(fromEnum: TWO)
+        }';
+
+        self::assertArraySubset(
+            [
+                'data'   => [
+                    'otherEnum' => 'StringRepresentationOfOtherEnumWithValueONE',
```
spawnia, line 650: Relation to enum class should be clarified through constant

tests/Type/TestClasses/OtherEnumType.php
```diff
@@ -0,0 +1,43 @@
...
+                'ONE',
+                'TWO',
+                'THREE',
+            ],
+        ];
+        parent::__construct($config);
+    }
+
+    /**
+     * @param mixed[]|null $variables
```
spawnia, line 26:
```suggestion
     * @param array<string, mixed>|null $variables
```

spawnia (changes requested):

spawnia:
I have been tying `EnumType` to programmatically defined PHP-Enums for a while now through https://github.com/nuwave/lighthouse/blob/master/src/Schema/Types/LaravelEnumType.php.

ruudk:
> I have been tying `EnumType` to programmatically defined PHP-Enums for a while now through https://github.com/nuwave/lighthouse/blob/master/src/Schema/Types/LaravelEnumType.php.

I saw it, but that only works for serializing Enum objects to string values. This PR introduces the possibility to also have those Enum objects as input arguments. Instead of getting the string value of the enum, you now can get the Enum object. 

Thanks for the review, I applied your feedback and updated the PR.

tests/Type/TestClasses/OtherEnumType.php
ruudk: Not possible because of 
```
  28     Parameter #2 $variables (array<string, mixed>|null) of method GraphQL\Tests\Type\TestClasses\OtherEnumType::parseLiteral() should be contravariant with parameter
         $variables (array|null) of method GraphQL\Type\Definition\EnumType::parseLiteral()
  28     Parameter #2 $variables (array<string, mixed>|null) of method GraphQL\Tests\Type\TestClasses\OtherEnumType::parseLiteral() should be contravariant with parameter
         $variables (array|null) of method GraphQL\Type\Definition\LeafType::parseLiteral()
```

spawnia:
> but that only works for serializing Enum objects to string values

@ruudk nope, it actually works for input and output.

ruudk:
Maybe I confuse input with arguments. I tried to use them as arguments and there it doesn't work. The args are passed as raw string values. With this PR I get the actual objects.

spawnia:
There is no difference between field arguments and input fields here, it works for both. We even have the test case to prove it: https://github.com/nuwave/lighthouse/blob/master/tests/Unit/Schema/Types/LaravelEnumTypeTest.php

spawnia:
That said, I am open to merging this PR. I just want to make extra sure it does not break something, since I think it does not really enable new use cases.

ruudk:
Interesting. I'll try and debug some more.

Do you know what the purpose of the `EnumType::parseLiteral` function is then?

spawnia:
Well, should be the same as in scalar types. It is not really getting called though, so it seems pretty useless.

I simplified the test case to omit the somewhat realistic enum class use case. Instead, I went with the most basic test that ensures the proper methods are getting called: https://github.com/webonyx/graphql-php/pull/1008

Looks like your implementation change fixes `parseLiteral`, but `parseValue` is still not getting called. This means that your implementation would not work with variable values getting passed into the query.

spawnia:
There is some risk of breakage and slight behavioral change through this PR that I don't want to risk in v14. https://github.com/webonyx/graphql-php/pull/1008 will be included in the upcoming v15 though.

PR #1011, opened 2021-11-15
Fix compatiblity of methods with native return type in PHP 8.1

SilinMykola:
This pull request (PR) provides fixes for the extension to make it compatible with PHP 8.1.
The fixes provides backward compatibility with previous versions PHP.

Return types have been fixed using attribute #[\ReturnTypeWillChange], please see this [link](https://php.watch/versions/8.1/ReturnTypeWillChange)


spawnia:
Thanks, fixed and backported to v14

PR #1012, opened 2021-11-18
Fix compatibility with PHP 8.1

janlanger:
These changes were made in master - https://github.com/webonyx/graphql-php/commit/451834341593bef6b82d080e2428e19edd59c4f5#diff-5f4e849a2f284951ac6488a3b3b386a9b0778daed44f61397dd89e398e3c53af but not ported to 14.x

### DIFF
--- a/src/Utils/MixedStore.php
+++ b/src/Utils/MixedStore.php
@@ -7,6 +7,7 @@ namespace GraphQL\Utils;
 use ArrayAccess;
 use GraphQL\Type\Definition\EnumValueDefinition;
 use InvalidArgumentException;
+use ReturnTypeWillChange;
 use SplObjectStorage;
 use function array_key_exists;
 use function array_search;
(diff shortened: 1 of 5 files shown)


spawnia:
I am up for backporting this change, given CI passes.

janlanger:
@spawnia CS and phpstan fail should be fixed

src/Utils/MixedStore.php
```diff
@@ -184,7 +186,7 @@ public function offsetGet($offset)
      *
      * @return void
```
ruudk, line 187:
```suggestion
```

src/Utils/MixedStore.php
```diff
@@ -220,7 +222,7 @@ public function offsetSet($offset, $value)
...
```
ruudk, line 223:
```suggestion
```

ruudk (commented):
To fixes needed to make the PR green.

PR #1013, opened 2021-11-19
Switch to PHPStan level 4

spawnia:


simPod (approved):

tests/BaseTestCase.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Tests;
+
+use PHPUnit\Framework\TestCase;
+
+abstract class BaseTestCase extends TestCase
```
simPod, line 9: I prefered the previous name as it implies it's not a standalone testcase but only the base for testcase

PR #1014, opened 2021-11-19
Fix PHPStan level 5 errors for tests

spawnia:


tests/Server/ServerConfigTest.php
```diff
@@ -155,14 +167,18 @@ public function testAllowsSettingPersistedQueryLoader(): void
...
+        $config->setPersistedQueryLoader($closure);
+        self::assertSame($closure, $config->getPersistedQueryLoader());
+    }
 
-        $loader = 'date'; // test for callable
-        $config->setPersistedQueryLoader($loader);
-        self::assertSame($loader, $config->getPersistedQueryLoader());
+    public static function loadPersistedQuery(): ?string
+    {
+        return rand() === 0 ? '{ foo }' : null;
```
simPod, line 181: I'd avoid randomness in tests

simPod (approved):

tests/Server/ServerConfigTest.php
spawnia: Right. I was trying to match the signature and figured that was an easy way to make PHPStan not complain about useless code, but now that you mentioned it it is unnecessary.

PR #1015, opened 2021-11-20
Progress towards PHPStan level 5

spawnia:


phpcs.xml.dist
```diff
@@ -18,6 +18,9 @@
     <file>tests</file>
 
     <rule ref="Doctrine">
+        <!-- Prevents useful type assertions -->
+        <exclude name="SlevomatCodingStandard.Commenting.InlineDocCommentDeclaration.MissingVariable" />
```
simPod, line 22: I wonder it prevents exactly 🤔

simPod (approved):

phpcs.xml.dist
spawnia: For example an `@var` after an if condition that proves the type of a variable.

PR #1016, opened 2021-11-21
PHPStan level 5

spawnia:


spawnia:
@simPod I would appreciate a review so I can go ahead with level 6 😉

src/Language/AST/NodeList.php
```diff
@@ -133,15 +134,14 @@ public function splice(int $offset, int $length, $replacement = null): NodeList
...
      *
      * @phpstan-return NodeList<T>
      */
-    public function merge($list): NodeList
+    public function merge(iterable $list): NodeList
     {
-        if ($list instanceof self) {
-            $list = $list->nodes;
+        if (! is_array($list)) {
+            $list = iterator_to_array($list);
```
simPod, line 144: this will fail when key is not array-key, can we type it as `iterable<array-key, T>`?

src/Utils/Value.php
```diff
@@ -36,7 +37,7 @@ class Value
     /**
      * Given a type and any value, return a runtime value coerced to match the type.
      *
-     * @param ScalarType|EnumType|InputObjectType|ListOfType|NonNull                              $type
+     * @param ScalarType|EnumType|InputObjectType|ListOfType<Type                                 &InputType>|NonNull $type
```
simPod, line 40: the space is weird

simPod (approved):
sorry, missed the notification

src/Utils/Value.php
spawnia: https://github.com/webonyx/graphql-php/issues/1020

src/Language/AST/NodeList.php
spawnia: Good catch, never thought about this.

PR #1017, opened 2021-11-24
Change return type of `SyncPromiseAdapter::wait()` to `mixed`

ruudk:
The value can be anything, depending on what the SyncPromise holds.


ruudk:
@spawnia Is it really needed that `codecov` bot replies as a comment? I'm always happy when I see a notification with a reply on my PR, then it turns out it's a bot 🙈

spawnia:
@ruudk I considered getting rid of them before since they have been annoying me as well. You know what - your comment is the straw that broke the camel's back. I will turn it off.

PR #1022, opened 2021-12-01
Upgrade to PHPStan 1

spawnia:


composer.json
```diff
@@ -9,21 +9,21 @@
   "homepage": "https://github.com/webonyx/graphql-php",
   "license": "MIT",
   "require": {
-    "php": "^7.4 || ^8.0",
+    "php": "^7.4 || ^8",
```
simPod, line 12: Is there any difference btw?

composer.json
```diff
@@ -9,21 +9,21 @@
...
-    "phpstan/phpstan-phpunit": "0.12.22",
-    "phpstan/phpstan-strict-rules": "0.12.11",
+    "amphp/amp": "^2.6",
+    "dms/phpunit-arraysubset-asserts": "^0.3",
+    "doctrine/coding-standard": "^9",
+    "ergebnis/composer-normalize": "^2.16",
+    "nyholm/psr7": "^1.4",
+    "phpbench/phpbench": "^1.2",
+    "phpstan/extension-installer": "^1.1",
+    "phpstan/phpstan": "1.2.0",
```
simPod, line 24: for v1.x.y we can use 

```suggestion
    "phpstan/phpstan": "^1.2.0",
```

phpstan.neon.dist
```diff
@@ -21,6 +21,12 @@ parameters:
         # allow partial parsing of language fragments.
         - "~Variable method call on GraphQL\\\\Language\\\\Parser\\.~"
 
+        # We utilize lazy initialization of non-nullable properties
+        - "~Property .* in isset\\(\\) is not nullable~"
```
simPod, line 25:
```suggestion
        - '~Property .+? in isset\(\) is not nullable~'
```

simPod (approved):

composer.json
spawnia: Semantically, no. I like the simplicity and aesthetics.

phpstan.neon.dist
spawnia: Single quotes reducing the amount of backslashes is nice, but double quotes are used in the rest of the file and in the generated baseline.

composer.json
spawnia: Let's try it. We might have to revert to a fixed version regardless of the 1.x status, because PHPStan might find new errors in newer versions which would then randomly fail CI in completely unrelated PRs.

PR #1023, opened 2021-12-02
Clean up separation of Type vs. NamedType

spawnia:


simPod (changes requested):

src/Type/Definition/FieldDefinition.php
```diff
@@ -196,38 +198,24 @@ public function assertValid(Type $parentType): void
...
-            $this->resolveFn === null || is_callable($this->resolveFn),
-            sprintf(
-                '%s.%s field resolver must be a function if provided, but got: %s',
-                $parentType->name,
-                $this->name,
-                Utils::printSafe($this->resolveFn)
-            )
-        );
+        /** @var Type&NamedType $type */
+
```
simPod, line 207:
```suggestion
```

src/Type/Definition/NamedTypeTrait.php
```diff
@@ -0,0 +1,52 @@
...
+
+use ReflectionClass;
+
+use function array_key_exists;
+use function preg_replace;
+
+/**
+ * @see NamedType
+ */
+trait NamedTypeTrait
```
simPod, line 15: `use NameType<how to name all these contents?>`. It reads well then.

src/Type/Definition/FieldDefinition.php
spawnia: I actually wanted this right under the previous `if` block, as that is the code that proves this type. PHPCS screwed that up.

src/Type/Definition/NamedTypeTrait.php
spawnia: I definitely want to keep `NamedType` in there, as the trait is directly tied to the interface. This pattern approximates multiple inheritance, which is unfortunately not available in PHP. Perhaps `NamedTypeImplementation` will make it more clear?

simPod: yup 👌🏾

simPod (approved):

PR #1026, opened 2021-12-06
Improve static validation for type config

spawnia:


src/Type/Definition/InputObjectField.php
```diff
@@ -51,8 +60,9 @@ public function getType(): Type
             /**
              * @see it('rejects an Input Object type with incorrectly typed fields')
              */
+
              // @phpstan-ignore-next-line schema validation will catch a Type that is not an InputType
-            $this->type = Schema::resolveType($this->config['type']);
+            return $this->type = Schema::resolveType($this->config['type']);
```
simPod, line 65: maybe the return here is a bit confusing as it's used with assignment and the following line seems to handle it.

src/Type/Definition/EnumType.php
```diff
@@ -159,18 +182,14 @@ public function assertValid(): void
...
 
         $values = $this->config['values'] ?? null;
-        if (! is_array($values)) {
-            $safeValues = Utils::printSafe($values);
+        // @phpstan-ignore-next-line should not happen if used correctly
+        if (! is_iterable($values)) {
+            $notIterable = Utils::printSafe($values);
 
-            throw new InvariantViolation("{$this->name} values must be an array, got: {$safeValues}");
+            throw new InvariantViolation("{$this->name} values must be iterable, got: {$notIterable}");
```
simPod, line 189:
```suggestion
            throw new InvariantViolation("{$this->name} values must be an iterable, got: {$notIterable}");
```

simPod (approved):

src/Type/Definition/InputObjectField.php
spawnia: I started introducing this early return where the code is:

```diff
if (! isset($this->lookup[$key]) {
-   $this->lookup[$key] = $this->produceValue($key);
+   return $this->lookup[$key] = $this->produceValue($key);
}

return $this->lookup[$key];
```

In the previous example, it should obviously improve performance - but even in this case I think it does. Property lookups are not free.

PR #1029, opened 2021-12-08
CustomScalarType > Make `serialize` optional

ruudk:
Just like all the other keys.


ruudk:
I don't understand why this change breaks here:
https://github.com/webonyx/graphql-php/blob/6fed3469a76761add9c727c35e3c5dc4578ab739/src/Utils/SchemaExtender.php#L124-L135

I only made it optional, while here it's still passed.

spawnia:
It is required for a reason, without it `CustomScalarType` will not work.

PR #1030, opened 2021-12-08
ObjectType > Make `fields` optional

ruudk:
Just like everything else. It can default to an empty array.


spawnia:
I don't think this change is right. See http://spec.graphql.org/draft/#sec-Objects.Type-Validation

> An Object type must define one or more fields.

ruudk:
Good point, thanks.

PR #1032, opened 2021-12-08
Use return instead of phpstan-return

ruudk:
This stops the Deprecation Helper from Symfony PHPUnit Bridge to stop complaining:
```
  1x: Method "ArrayAccess::offsetGet()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "GraphQL\Language\AST\NodeList" now to avoid errors or add an explicit @return annotation to suppress this message.
```

### DIFF
--- a/src/Language/AST/NodeList.php
+++ b/src/Language/AST/NodeList.php
@@ -70,7 +70,7 @@ class NodeList implements ArrayAccess, IteratorAggregate, Countable
      *
      * @param int|string $offset
      *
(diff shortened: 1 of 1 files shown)


spawnia:
Does https://github.com/webonyx/graphql-php/pull/1033 also resolve your issue?

ruudk:
I'm sure that will fix it, I'll wait for that to be merged in main.

PR #1038, opened 2021-12-17
Allow lazy type loader to return null

spawnia:
Fixes https://github.com/nuwave/lighthouse/issues/1936

### DIFF
--- a/src/Type/Schema.php
+++ b/src/Type/Schema.php
@@ -228,6 +228,22 @@ class Schema
         return $this->config->directives ?? GraphQL::getStandardDirectives();
     }
 
+    /**
+     * @param mixed $typeLoaderReturn could be anything
+     */
+    public static function typeLoaderNotType($typeLoaderReturn): string
+    {
+        $typeClass = Type::class;
+        $notType   = Utils::printSafe($typeLoaderReturn);
+
+        return "Type loader is expected to return an instanceof {$typeClass}, but it returned {$notType}";
+    }
+
+    public static function typeLoaderWrongTypeName(string $expectedTypeName, string $actualTypeName): string
+    {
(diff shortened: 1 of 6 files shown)


src/Type/Schema.php
```diff
@@ -228,6 +228,22 @@ public function getDirectives(): array
...
+        $notType   = Utils::printSafe($typeLoaderReturn);
+
+        return "Type loader is expected to return an instanceof {$typeClass}, but it returned {$notType}";
+    }
+
+    public static function typeLoaderWrongTypeName(string $expectedTypeName, string $actualTypeName): string
+    {
+        return "Type loader is expected to return type {$expectedTypeName}, but it returned type {$actualTypeName}.";
+    }
+
```
simPod, line 246: I'd extract this to standalone InvariantViolation factory class so Schema does not grow larger

simPod (approved):

src/Type/Schema.php
spawnia: How about dedicated exception classes, extending from InvariantViolation?

simPod: Also possible.

I'd prefer to have `InvariantViolation` interface and concrete final classes implementing it. But everything similar works too I guess.

spawnia: Do you actually mean an interface, or rather an abstract class?

simPod: @spawnia interface, I'm not really fan of extending when it's not necessary

spawnia: Generally agree, but in this particular case it seems very inconvenient/useless? The interface would have no methods then, and the implementing classes all have to extend a base exception class?

I think the main idea behind dedicated exception classes *is* indeed to have a hierarchy of exceptions. This of course is object-orientation and inheritance hell galore, but given the options of what PHP offers, it may be the best course of action?

Issue #1020, opened 2021-11-25
Switch to php-cs-fixer


spawnia:
We currently use https://github.com/squizlabs/PHP_CodeSniffer for formatting. I have been annoyed with it because:
1. it keeps messing up complex types in PHPDocs (e.g. https://github.com/webonyx/graphql-php/pull/1016#discussion_r756812057)
2. the way it aligns some code is nearly impossible to do right manually
3. alignment leads to ugly diffs
4. configuration options are very hard to discover
5. XML-based configuration is clunky

I propose we switch to https://github.com/FriendsOfPHP/PHP-CS-Fixer instead. I am using it in a lot of projects and have not had much issues with it at all. It addresses the above mentioned pain points:
1. it deals fine with complex types (e.g. https://github.com/nuwave/lighthouse/blob/04a74cc6bbc6e592de988096a890958ebe112a4e/src/Schema/AST/DocumentAST.php#L40)
2. my code is mostly fine the way i type it
3. it is easily configured to left-align everything
4. https://mlocati.github.io/php-cs-fixer-configurator
5. PHP-based configuration is nice and composable

mfn:
# 👍🏼 

from my side.

Whilst I'm happy both projects are maintained, my impression is that with PHP adaptions, adjustments, release cadence and whatnot makes a healthy project, next to the technical things pointed out, php-cs-fixer is in a better state.

My experience is similar and I propose it also in many projects working on.

This combined with a CI webhook to auto-add a commit reduces developer friction a lot.

ruudk:
FWIW: It would be great to move forward with this, as currently phpcs/phpcbf don't work at all:
```
vendor/bin/phpcbf
ERROR: Referenced sniff "Doctrine" does not exist

Run "phpcbf --help" for usage information
```

PR #1039, opened 2021-12-18
PHPStan level 6

spawnia:


spawnia:
@simPod I am having a hard time with some of the template/generics stuff, see `phpstan-baseline.neon`.

Having to specify a generic type for `NodeList` is tricky, because type hinting it as the most open `NodeList<Node>` causes errors when trying to pass in a more specific instance, e.g. `NodeList<VariableNode>`. I was able to solve a similar case in `ListOfType` by marking its generic parameter as `@template-covariant`, but that is not possible in `NodeList` due to it accepting writes. Do you have an idea how to solve this?

If you have any ideas on how to fix the other errors I put in the baseline, that would also be great. Some of them just seem like we are hitting the limits of PHPStan's capabilities?

simPod:
I'm currently doing some winter sports, will check it out ASAP when I'm
back (few days)

spawnia:
Thank you, have fun in the mountains 🏂

simPod:
> due to it accepting writes. Do you have an idea how to solve this?

this is actually correct as `NodeList<VariableNode>` has the same contract as `NodeList<Node>`

1. With current state we could accept `mixed` (`template T`) and throw something similar like java does `java.lang.ArrayStoreException` if the node value is not of the same type as others nor subtype of `Node`
2. Also, we can consider dropping array access and making the NodeList immutable. I tried to search for usages and it seems we don't need `accepting writes`

src/Executor/ReferenceExecutor.php
```diff
@@ -926,6 +927,7 @@ function ($previous, $value) use ($callback) {
     /**
      * Complete a list value by completing each item in the list with the inner type.
      *
+     * @param ListOfType<Type             &OutputType> $returnType
```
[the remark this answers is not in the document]
spawnia: I will try https://github.com/webonyx/graphql-php/issues/1020 after this PR is merged.

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -95,60 +90,14 @@ public function buildDirective(DirectiveDefinitionNode $directiveNode): Directiv
...
-    /**
-     * Given an ast node, returns its string description.
-     */
-    private function getDescription(Node $node): ?string
-    {
-        if (isset($node->description)) {
-            return $node->description->value;
-        }
-
-        if (isset($this->options['commentDescriptions'])) {
```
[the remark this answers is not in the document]
spawnia: Following the reference implementation, I removed this deprecated option.

tests/Executor/Promise/SyncPromiseTest.php
```diff
@@ -418,16 +433,16 @@ public function testPendingPromiseThen(): void
...
+        self::assertValidPromise($nextPromise, 1, null, SyncPromise::FULFILLED);
+        self::assertValidPromise($nextPromise2, 1, null, SyncPromise::FULFILLED);
+        self::assertValidPromise($nextPromise3, 2, null, SyncPromise::FULFILLED);
+        self::assertValidPromise($nextPromise4, 3, null, SyncPromise::FULFILLED);
     }
 
     public function testRunEmptyQueue(): void
     {
-        SyncPromise::runQueue();
+        $this->expectNotToPerformAssertions();
```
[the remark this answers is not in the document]
spawnia: Reintroduced `assertDidNotCrash()`

spawnia:
1. We should avoid runtime checks, they are bad for performance. I think it is acceptable to have code that *might* be misused, as long as we have the tests to ensure we don't.
2. We make heavy use of AST manipulation in `nuwave/lighthouse`, including appending to `NodeList` via `[]=`.

simPod:
It we want to keep it this way we should maybe document the NodeList is not designed well as consumers will get the same errors as we do now from static analysis.

> The following example shows the disparity between the validation rules for array and generic classes

The example is about something else than we're dealing with though. And in out situations arrays are not relevant. The problem is LSP that requires method parameters to accept the same as parent does. So since `NodeList<Node>::add(Node $node)` accepts any Node, `NodeList<SubNode>::add(Node $node)` must accept it as well. This works the same with inheritance as generics.


> I think of the type safety afforded to us by PHPStan as a tool to help us write better, more robust code. If it fails to achieve that in some cases, we can choose to ignore it.

We can allow it because php does not care, at least yet. E.g. in scala this would no compile as it's against LSP.

spawnia:
I have been pondering the comparison with arrays some more. In contrast to classes whose type parameters are set at construction, an array (and the current implementation of `NodeList`) allows dynamically changing it. When pushing a `FieldNode` to a `NodeList<ArgumentNode>`, it becomes a `NodeList<FieldNode|ArgumentNode>`. I don't think this is ever sensible, but it would work without crashing. Anyways, that nuance is irrelevant for our case, `NodeList` always contains a single type - and the type does not change at runtime.

I desire the following two properties from `NodeList`, and they seem to be at odds with each other:
1. Allow specifying the contained type in a structured way that is accessible to static analysis. Very useful when dealing the AST. I don't want to erase the genericness.
2. Allow code to accept a `NodeList` with any generic type as long as it is only read.

I dug around for ways to resolve 2., and found something that somewhat works. However, it surfaces what I believe to be a bug in PHPStan - see https://github.com/phpstan/phpstan/issues/6237. We can ignore that in the baseline for now though, we will have a robust solution in place for when it is fixed.

src/Language/AST/Node.php
```diff
@@ -92,36 +93,29 @@ protected static function cloneValue($value)
...
-        $tmp = (array) $this;
-
-        if ($this->loc !== null) {
-            $tmp['loc'] = [
-                'start' => $this->loc->start,
-                'end'   => $this->loc->end,
-            ];
-        }
+        return $this->toArray();
+    }
```
simPod, line 105: Is this actually needed?

simPod (approved):
This is already too huge so we can 🚢 

_About the NodeList, I wonder if I'm right (generic NodeList abuses LSP and should not compile in decent language) or you're right (phpstan needs some feature in order to define that)_ 😜

CHANGELOG.md
```diff
@@ -42,6 +45,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Add SDL validation rule `UniqueOperationTypes` (#995)
 - Add ability to remove custom validation rules after adding them via `DocumentValidator::removeRule()`
 - Allow lazy enum values
+- Make `Node` implement `JsonSerializable`
```
simPod, line 48: can we specify why eventually? If I was to chose I'd not add the interface

spawnia:
I think the definition of `NodeList` is fine, it is a mutable container around a generic type. Thus, its template type is neither covariant nor covariant. PHPStan lacks the ability for a method to express that it uses `NodeList` exclusively as if its template type were covariant. Java actually has such a feature, which is why I opened https://github.com/phpstan/phpstan/issues/6240.

src/Language/AST/Node.php
spawnia: Yeah, it makes the JSON representation more compact. In contrast to the default serialization, null values and some runtime only values (such as pointers to previous and next tokens in `Location`) are excluded.

CHANGELOG.md
spawnia: Added a docblock to explain why it is useful

PR #1040, opened 2021-12-22
Allow passing enum values via callable

ruudk:
This makes it possible to return the enum values via a closure. It also makes it possible to use generators. 

It's in line with the rest of the library.

### DIFF
--- a/src/Type/Definition/EnumType.php
+++ b/src/Type/Definition/EnumType.php
@@ -15,6 +15,7 @@ use GraphQL\Utils\MixedStore;
 use GraphQL\Utils\Utils;
 
 use function is_array;
+use function is_callable;
 use function is_iterable;
 use function is_string;
 
@@ -26,10 +27,11 @@ use function is_string;
  *   description?: string|null,
  *   astNode?: EnumValueDefinitionNode|null,
  * }
+ * @phpstan-type EnumValues iterable<string, PartialEnumValueConfig>|iterable<string, mixed>|iterable<int, string>
  * @phpstan-type EnumTypeConfig array{
  *   name?: string|null,
  *   description?: string|null,
- *   values: iterable<string, PartialEnumValueConfig>|iterable<string, mixed>|iterable<int, string>,
+ *   values: EnumValues|callable(): EnumValues,
  *   astNode?: EnumTypeDefinitionNode|null,
  *   extensionASTNodes?: array<int, EnumTypeExtensionNode>|null,
  * }
@@ -96,8 +98,13 @@ class EnumType extends Type implements InputType, OutputType, LeafType, Nullable
         if (! isset($this->values)) {
             $this->values = [];
 
+            $values = $this->config['values'];
+            if (is_callable($values)) {
+                $values = $values();
+            }
+
             // We are just assuming the config option is set correctly here, validation happens in assertValid()
-            foreach ($this->config['values'] as $name => $value) {
+            foreach ($values as $name => $value) {
                 if (is_string($name)) {
                     if (is_array($value)) {
                         $value += ['name' => $name, 'value' => $name];
@@ -183,10 +190,10 @@ class EnumType extends Type implements InputType, OutputType, LeafType, Nullable
 
         $values = $this->config['values'] ?? null;
         // @phpstan-ignore-next-line should not happen if used correctly
-        if (! is_iterable($values)) {
+        if (! is_iterable($values) && ! is_callable($values)) {
             $notIterable = Utils::printSafe($values);
 
-            throw new InvariantViolation("{$this->name} values must be an iterable, got: {$notIterable}");
+            throw new InvariantViolation("{$this->name} values must be an iterable or callable, got: {$notIterable}");
         }
 
         $this->getValues();
--- a/tests/Type/DefinitionTest.php
+++ b/tests/Type/DefinitionTest.php
@@ -1684,6 +1684,19 @@ final class DefinitionTest extends TestCase
         self::assertEquals(20, $enumType->getValue('BAR')->value);
     }
 
+    public function testAcceptsACallableAsEnumValues(): void
+    {
+        $enumType = new EnumType([
+            'name'   => 'SomeEnum',
+            'values' => static function (): iterable {
+                yield 'FOO' => ['value' => 10];
+                yield 'BAR' => ['value' => 20];
+            },
+        ]);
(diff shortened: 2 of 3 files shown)


ruudk:
@spawnia Ready for review 🎉

tests/Type/EnumTypeTest.php
```diff
@@ -64,10 +64,10 @@ public function setUp(): void
 
         $ComplexEnum = new EnumType([
```
spawnia, line 65: This test case is supposed to mirror the reference implementation. Please revert this part.

Instead, add a dedicated test method such as `testLazilyDefineValuesAsCallable()`. In addition to `getValues()` working, the test should also ensure lazyness - the callable should not be evaluated upon construction.

src/Type/Definition/EnumType.php
```diff
@@ -183,10 +190,10 @@ public function assertValid(): void
 
         $values = $this->config['values'] ?? null;
         // @phpstan-ignore-next-line should not happen if used correctly
-        if (! is_iterable($values)) {
+        if (! is_iterable($values) && ! is_callable($values)) {
             $notIterable = Utils::printSafe($values);
 
-            throw new InvariantViolation("{$this->name} values must be an iterable, got: {$notIterable}");
+            throw new InvariantViolation("{$this->name} values must be an iterable or callable, got: {$notIterable}");
```
spawnia, line 196: Please add a test case to cover this, I think [here](https://github.com/webonyx/graphql-php/blob/b4cbc3ab27d35e18f4fd9318a855952aa8711875/tests/Type/DefinitionTest.php#L1703) would be a good spot.

spawnia (changes requested):

src/Type/Definition/EnumType.php
```diff
@@ -26,10 +27,11 @@
  *   description?: string|null,
  *   astNode?: EnumValueDefinitionNode|null,
  * }
+ * @phpstan-type EnumValuesType iterable<string, PartialEnumValueConfig>|iterable<string, mixed>|iterable<int, string>
```
spawnia, line 30:
```suggestion
 * @phpstan-type EnumValues iterable<string, PartialEnumValueConfig>|iterable<string, mixed>|iterable<int, string>
```

tests/Type/EnumTypeTest.php
ruudk: Ok, can I still add a new enum type to the setUp or should it be isolated in the `testLazilyDefineValuesAsCallable`?

spawnia: Isolated.

src/Type/Definition/EnumType.php
ruudk: Done.

tests/Type/EnumTypeTest.php
ruudk: Done

tests/Type/DefinitionTest.php
```diff
@@ -1684,6 +1684,22 @@ public function testAcceptsAWellDefinedEnumTypeWithInternalValueDefinition(): vo
         self::assertEquals(20, $enumType->getValue('BAR')->value);
     }
 
+    /**
+     * @see it('accepts a callable as Enum values')
+     */
```
spawnia, line 1689: Should be gone, since this test does not exist in the reference implementation.

tests/Type/DefinitionTest.php
```diff
@@ -1701,6 +1717,23 @@ public function testRejectsAnEnumTypeWithIncorrectlyTypedValues(): void
         $enumType->assertValid();
     }
 
+    /**
+     * @see it('rejects an Enum type with incorrectly typed values')
+     */
```
spawnia, line 1722:
```suggestion
```

tests/Type/EnumTypeTest.php
```diff
@@ -655,4 +655,57 @@ public function testCallsOverwrittenEnumTypeMethods(): void
             GraphQL::executeQuery($this->schema, $query, null, null, $variables)->toArray(DebugFlag::INCLUDE_DEBUG_MESSAGE)
         );
     }
+
+    public function testLazilyDefineValuesAsCallable(): void
+    {
+        $called = false;
```
spawnia, line 661:
```suggestion
        $called = 0;
```

Let's have the test case use the enum in both argument and result and assert this is only called once.

tests/Type/EnumTypeTest.php
```diff
@@ -655,4 +655,57 @@ public function testCallsOverwrittenEnumTypeMethods(): void
...
+        $ColorType = new EnumType([
+            'name'   => 'Color',
+            'values' => function() use (&$called) : iterable {
+                $called = true;
+                yield 'RED'   => ['value' => 0];
+                yield 'GREEN' => ['value' => 1];
+                yield 'BLUE'  => ['value' => 2];
```
spawnia, line 668:
```suggestion
                yield 'RED'   => ['value' => 0];
```

Should suffice for this test

ruudk:
Btw, it appears that `phpcbf` is broken, any idea how to solve that?
```
vendor/bin/phpcbf
ERROR: Referenced sniff "Doctrine" does not exist

Run "phpcbf --help" for usage information
```

tests/Type/EnumTypeTest.php
```diff
@@ -655,4 +655,57 @@ public function testCallsOverwrittenEnumTypeMethods(): void
...
+
+        $QueryType = new ObjectType([
+            'name'   => 'Query',
+            'fields' => [
+                'colorEnum'   => [
+                    'type'    => $ColorType,
+                    'args'    => [
+                        'fromEnum'   => ['type' => $ColorType],
+                        'fromInt'    => ['type' => Type::int()],
+                        'fromString' => ['type' => Type::string()],
```
spawnia, line 680: No need
```suggestion
```

spawnia (changes requested):

spawnia:
> Btw, it appears that `phpcbf` is broken, any idea how to solve that?

https://github.com/webonyx/graphql-php/issues/1020 😉 

For now, no idea. Perhaps `composer update`?

ruudk:
Ouch, it seems that PHPStan doesn't understand the `assertSame(1, $called)` trick. That's odd 🤔  Any idea?

ruudk:
It's solved :)

spawnia:
Nicely done, thank you

PR #1058, opened 2022-01-27
Allow constructing `EnumType` from PHP enum

spawnia:
ou can reuse your existing PHP enums in GraphQL.
Leverage PHP attributes to add descriptions and deprecate values:

```php
use GraphQL\Type\Definition\Deprecated;
use GraphQL\Type\Definition\Description;

#[Description(description: 'Sweet and juicy.')]
enum Fruit
{
    #[Description(description: 'Rich in potassium.')]
    case BANANA;

    #[Deprecated(reason: 'Too sour.')]
    case CITRON;
}
```

PhpDocs will be used as descriptions if present, but are overridden by the `Description` attribute.

Wrap them with `GraphQL\Type\Definition\PhpEnumType` to use them in a GraphQL schema:

```php
use GraphQL\Type\Definition\PhpEnumType;

$fruitEnumType = new PhpEnumType(Fruit::class);
```

The following type will be deduced from `Fruit`:

```graphql
"""
Sweet and juicy.
"""
enum Fruit {
  """
  Rich in potassium.
  """
  BANANA

  CITRON @deprecated(reason: "Too sour.")
}
```

Conversion rules:

- Client-given arguments will be converted to enum instances
- Returned output values are expected to be enum instances


src/Type/Definition/PhpEnumType.php
```diff
@@ -0,0 +1,71 @@
...
+
+            $enumDefinitions[] = [
+                'name' => $case->name,
+                'value' => $case,
+                'description' => $this->description($docComment),
+                'deprecationReason' => $this->deprecationReason($docComment),
+            ]
+        }
+
+        parent::__consstruct([
```
leoloso, line 46: Typo: `__consstruct` => `__construct`

src/Type/Definition/PhpEnumType.php
```diff
@@ -0,0 +1,71 @@
...
+        $enumDefinitions = [];
+        foreach ($cases as $case) {
+            $reflectionCase = $reflection->getCase($case->name);
+            $docComment = $reflectionCase->getDocComment();
+
+            $enumDefinitions[] = [
+                'name' => $case->name,
+                'value' => $case,
+                'description' => $this->description($docComment),
+                'deprecationReason' => $this->deprecationReason($docComment),
```
ruudk, line 42: I think good to add test cases for this too...

ruudk:
Is it an idea to also include a `parseValue` method that calls `enum::from` so that enums coming from input are parsed to the actual enum value?

spawnia:
> Is it an idea to also include a `parseValue` method that calls `enum::from` so that enums coming from input are parsed to the actual enum value?

This will be the behaviour, stemming from the implementation of `EnumType`. I covered this in a test case.

spawnia:
@olivernybroe please review.

olivernybroe:
@spawnia Great job!

One suggestion which I think could improve it, in regards to descriptions. The description which you would add via docblock is basically the same you would like to add in the `description` attribute. So what about either by default reading the docblock or by adding description attribute without any value in it?

Issue #811, opened 2021-03-28
Usage of the `list<T>` PHPDoc annotation


spawnia:
Since @simPod has brought this up in some PRs, let's have a proper discussion about annotating sequentially numerically indexed arrays in PHPDocs.

PHPStan added partial support for this type in https://github.com/phpstan/phpstan-src/pull/83/files, but is lacking the full validation capabilities (ensuring gapless indexes) that Psalm provides https://psalm.dev/docs/annotating_code/type_syntax/array_types/#listsa. Thus, they do not even mention their usage in the docs: https://github.com/phpstan/phpstan/pull/4282

I think we should hold off on using `list<T>` until PHPStan fully commits to supporting it properly. Until then, `array<int, T>` does a pretty good impression of it, with the added bonus of IDE compatibility.

spawnia:
Apparantly, those two types are not the same. I think the semantics line up with the new `array_is_list` function in PHP 8.1: https://php.watch/versions/8.1/array_is_list

```php
/** @var list<int> */
$a = [1, 2];
unset($a[0]); // Should not be allowed, or the type must be array<int, int>
```

simPod:
Yes, the question should be whether we want to adapt the `list` type. But since it's indirectly introduced by `array_is_list()` in 8.1 I see no issue here. As you said, `list<T>` is covariant to `array<int, T>` but are not the same.

spawnia:
We could start using `list<T>` through `@phpstan-`-prefixed PHPDocs for now. I am not sure if the juice is worth the squeeze, I can't recall hitting bugs in this library due to sparse arrays.

PR #1059, opened 2022-01-27
Increase PHPStan analysis level to 8

spawnia:


src/Type/SchemaValidationContext.php
```diff
@@ -347,12 +347,14 @@ private function validateDirectivesAtLocation(NodeList $directives, string $loca
                 continue;
             }
 
-            $includes = Utils::some(
-                $schemaDirective->locations,
-                static function ($schemaLocation) use ($location): bool {
-                    return $schemaLocation === $location;
+            $includes = false;
```
simPod, line 350: some was more straightforward on what's going on

src/Validator/DocumentValidator.php
```diff
@@ -260,14 +257,23 @@ public static function validateSDL(
         ?Schema $schemaToExtend = null,
         ?array $rules = null
     ): array {
-        $usedRules = $rules ?? self::sdlRules();
+        $rules ??= self::sdlRules();
+
+        if (0 === count($rules)) {
```
simPod, line 262: btw why are we using yoda conditions? 🤔

src/Validator/SDLValidationContext.php
```diff
@@ -2,6 +2,42 @@
...
+use GraphQL\Language\AST\DocumentNode;
+use GraphQL\Type\Schema;
+
+class SDLValidationContext implements ValidationContext
 {
+    protected DocumentNode $ast;
+
+    protected ?Schema $schema;
+
+    /** @var array<int, Error> */
```
simPod, line 15: list?

spawnia: I would do one full pass over all instances of `array<int, ` and check if they can be `list<` in a separate PR - see https://github.com/webonyx/graphql-php/issues/811

src/Validator/DocumentValidator.php
spawnia: Because of `php-cs-fixer`. I don't mind them - do you have a strong opinion about it?

src/Language/Visitor.php
```diff
@@ -193,8 +202,6 @@ public static function visit(object $root, array $visitor, ?array $keyMap = null
         $ancestors = [];
         $newRoot = $root;
 
-        $UNDEFINED = null;
```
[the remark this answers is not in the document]
spawnia: Right, but it also decreased the efficiency of the code and made it harder to understand. I think the reference implementation shines by providing us with a really solid test suite, which we can easily copy. For the implementation, we can do better by diverging and using idiomatic PHP.

src/Language/Visitor.php
```diff
@@ -183,6 +183,15 @@ public static function visit(object $root, array $visitor, ?array $keyMap = null
     {
         $visitorKeys = $keyMap ?? self::VISITOR_KEYS;
 
+        /**
+         * array{
```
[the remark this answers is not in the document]
spawnia: Good catch. Turns out it is quite impossible to strictly type the recursive implementation. I switched to a non-recursive definition.

src/Validator/DocumentValidator.php
simPod: Hm, it makes reading code hard because that's not how people speak. 

I have never seen anyone to use assign instead comparison in a condition. Also we have phpstan and tests to prevent that. So there's no use for yoda.

simPod (approved):

PR #1060, opened 2022-01-29
Remove trailing comma from array shapes since psalm cannot read it

simPod:
Causes huge amount of type errors with code bases using psalm


spawnia:
Can we put some basic Psalm analysis in to catch this? A low level should suffice, not looking to satisfy two strict static analysers at the moment 😉

simPod:
Will see what can be done (in another PR)

simPod:
There are more issues, created this as a starter. Will try to detect them via psalm CI step now.

PR #1062, opened 2022-01-30
Run basic psalm step in CI

simPod:
It has better detection of broken array shape definitions

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 [![CI](https://github.com/webonyx/graphql-php/workflows/CI/badge.svg)](https://github.com/webonyx/graphql-php/actions?query=workflow:CI+branch:master)
 [![Coverage Status](https://codecov.io/gh/webonyx/graphql-php/branch/master/graph/badge.svg)](https://codecov.io/gh/webonyx/graphql-php/branch/master)
 [![Latest Stable Version](https://poser.pugx.org/webonyx/graphql-php/version)](https://packagist.org/packages/webonyx/graphql-php)
+[![Shepherd Type][shepherd image]][shepherd link]
 
 This is a PHP implementation of the GraphQL [specification](https://github.com/graphql/graphql-spec)
 based on the [reference implementation in JavaScript](https://github.com/graphql/graphql-js).
@@ -61,3 +62,6 @@ Support this project by becoming a sponsor. Your logo will show up here with a l
 ## License
 
 See [LICENSE](LICENSE).
+
+[shepherd image]: https://shepherd.dev/github/webonyx/graphql-php/coverage.svg
+[shepherd link]: https://shepherd.dev/github/webonyx/graphql-php
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![CI](https://github.com/webonyx/graphql-php/workflows/CI/badge.svg)](https://github.com/webonyx/graphql-php/actions?query=workflow:CI+branch:master)
 [![Coverage Status](https://codecov.io/gh/webonyx/graphql-php/branch/master/graph/badge.svg)](https://codecov.io/gh/webonyx/graphql-php/branch/master)
 [![Latest Stable Version](https://poser.pugx.org/webonyx/graphql-php/version)](https://packagist.org/packages/webonyx/graphql-php)
-[![Shepherd Type][shepherd image]][shepherd link]
+[![Shepherd Type](https://shepherd.dev/github/webonyx/graphql-php/coverage.svg)](ttps://shepherd.dev/github/webonyx/graphql-php)
 
 This is a PHP implementation of the GraphQL [specification](https://github.com/graphql/graphql-spec)
 based on the [reference implementation in JavaScript](https://github.com/graphql/graphql-js).
@@ -62,6 +62,3 @@ Support this project by becoming a sponsor. Your logo will show up here with a l
 ## License
 
 See [LICENSE](LICENSE).
-
-[shepherd image]: https://shepherd.dev/github/webonyx/graphql-php/coverage.svg
-[shepherd link]: https://shepherd.dev/github/webonyx/graphql-php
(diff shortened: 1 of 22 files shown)


simPod:
It works as planned

image

src/Type/SchemaConfig.php
```diff
@@ -24,8 +24,10 @@
  *
  *     $schema = new Schema($config);
  *
- * @phpstan-type TypeLoader callable(string $typeName): ((Type&NamedType)|null)
- * @phpstan-type Types array<Type&NamedType>|(callable(): array<Type&NamedType>)
+ * @see Type, NamedType
```
simPod, line 27: I do this to prevent IDE marking those imports as unused to prevent accidental removal and allow using auto format with import optimization

src/Validator/Rules/CustomValidationRule.php
```diff
@@ -7,7 +7,10 @@
 use GraphQL\Validator\ValidationContext;
 
 /**
- * @phpstan-type VisitorFn callable(ValidationContext): (array<string, callable(Node): VisitorOperation|mixed|null>|array<string, array<string, callable(Node): VisitorOperation|mixed|null>>)
```
simPod, line 10: There is like no support for using parentheses for defining precedence in union/intersection definitions

https://github.com/phpstan/phpdoc-parser/issues/74

src/Error/Warning.php
```diff
@@ -14,6 +14,7 @@
  * Also, it is possible to override warning handler (which is **trigger_error()** by default).
  *
  * @phpstan-type WarningHandler callable(string $errorMessage, int $warningId, ?int $messageLevel): void
+ * @psalm-type WarningHandler callable(string, int, int|null): void
```
simPod, line 17: Let's temporarily keep these so we suppress psalm errors in order to find other errors

https://github.com/vimeo/psalm/issues/7527

README.md
```diff
@@ -3,6 +3,7 @@
 [![CI](https://github.com/webonyx/graphql-php/workflows/CI/badge.svg)](https://github.com/webonyx/graphql-php/actions?query=workflow:CI+branch:master)
 [![Coverage Status](https://codecov.io/gh/webonyx/graphql-php/branch/master/graph/badge.svg)](https://codecov.io/gh/webonyx/graphql-php/branch/master)
 [![Latest Stable Version](https://poser.pugx.org/webonyx/graphql-php/version)](https://packagist.org/packages/webonyx/graphql-php)
+[![Shepherd Type][shepherd image]][shepherd link]
```
spawnia, line 6: Let's inline those links

src/Error/Warning.php
spawnia: Please link the relevant issue in a TODO above this line so we know when to remove it again. Same with other such workarounds.

src/Utils/SchemaPrinter.php
```diff
@@ -37,7 +37,7 @@
 /**
  * Prints the contents of a Schema in schema definition language.
  *
- * @phpstan-type Options array{}
+ * @phpstan-type Options array<string, bool>
```
spawnia, line 40: Empty array is more accurate.

src/Validator/Rules/CustomValidationRule.php
```diff
@@ -7,7 +7,10 @@
...
+ * @see Node, VisitorOperation
+ *
+ * @phpstan-type VisitorFnValue array<string, callable(Node): VisitorOperation|mixed|null>|array<string, array<string, callable(Node): VisitorOperation|mixed|null>>
```
spawnia, line 12:
```suggestion
 * @phpstan-type VisitorFnResult array<string, callable(Node): VisitorOperation|mixed|null>|array<string, array<string, callable(Node): VisitorOperation|mixed|null>>
```

spawnia (changes requested):
Can you disable the notices? I think we should only have hard errors or no report at all.

psalm-baseline.xml
```diff
@@ -0,0 +1,54 @@
...
+      <code>ParserOptions</code>
+      <code>array{string, ParserOptions}</code>
+    </UndefinedDocblockClass>
+  </file>
+  <file src="src/Language/Visitor.php">
+    <EmptyArrayAccess occurrences="1">
+      <code>$path[count($path) - 1]</code>
+    </EmptyArrayAccess>
+  </file>
+  <file src="src/Type/Definition/FieldDefinition.php">
```
simPod, line 25: I initially hoped to fix this especially but there's some crazy issue I'm not even able to isolate now.

spawnia: The type declarations in there are a little crazy - the full version was too crazy for PHPStan to handle 😆

spawnia (approved):
Nicely done. Some extra checks seldom hurt, and ensuring compatitibility for Psalm users is great.

psalm-baseline.xml
simPod: I'll definitely revisit after https://github.com/vimeo/psalm/issues/6928 https://github.com/vimeo/psalm/issues/7527 are solved

simPod:
> and ensuring compatitibility for Psalm users is great.

yup and it also helps discover PHPStan's issues as phpstan does not complain about stuff like this

https://github.com/webonyx/graphql-php/pull/1062/files#diff-0986d100050c3646f31bc2b6a974ab36b5932ab57b5f41bb3c120a334f767972R21-R23

psalm-baseline.xml
spawnia: Yeah. I did not feel like actually messing around in there tool beyond the PR I attempted. Using regex for parsing of something as complex as a type definition seems like a slippery slope.

simPod: > Using regex for parsing of something as complex as a type definition seems like a slippery slope.

exactly, already encountered so many issues caused by that :/

Reminds me of https://stackoverflow.com/a/1732454/519333

PR #1067, opened 2022-02-13
Fix problems found by the newest version of PHPStan

vhenzl:
The current master is failing as PHPStan got smarter in the newest versions.

Not 100% sure about the change to `FormattedError.php`. Context in phpstan/phpstan-src#914.


spawnia:
Thank you

PR #1068, opened 2022-02-13
Fix schema built from SDL to return null for unknown types

### LINKED ISSUE #997: `Schema::getType()` shoud always return `null` for non-existent type

vhenzl:
Fixes #997 (and unblocks #998 and #999).

### History background
The implementation of `BuildSchema`, `SchemaExtender` and `ASTDefinitionBuilder` roughly matches their counterparts from `graphql-js` v14.0.

However, during the lifetime of v14, the JS code underwent significant changes. In the final version 14.7, `buildASTSchema()`, `extendSchema()` and `ASTDefinitionBuilder` looks much different then they did in v14.0. For the context of this PR, the most important changes (https://github.com/graphql/graphql-js/commit/afc6b7c04708068e3733d8e85882e4b2e4ef58e5) were that  `ASTDefinitionBuilder`'s constructor didn't accept `typeDefinitionsMap` param any more, and that the original single method `buildType()` was replaced with a new `buildType()` and `getNamedType()`. Each was used in a different context, and only the latter could throw the `"Unknown type"` error.

These changes to the reference implementation have never been reflected in `graphql-php`.

### Implementation notes
The context is everything. Sometimes it's okay for the builder to return `null` instead of throwing `"Unknown type"` error – like in this case when it's requested to build a type that isn't defined in the SDL schema.

To handle this explicitly, I added the new `maybeBuildType(string $name): ?Type`  method to the builder.

`graphql-js` (v14.2+) builds `typeMap` with all types from SDL ahead. Then calling `typeMap[typeName]` returns either a type or `undefined`.

The new `maybeBuildType()` provides similar functionality for the lazy environment where we don't have such a map and where the individual types are built from the SDL schema on demand.

### Additional notes
In v15, `graphql-js` has undergone even more significant changes and `ASTDefinitionBuilder` has been removed completely (https://github.com/graphql/graphql-js/commit/3948aa20d9c38c5d4aff06f1b3af2538b20f909a). 
To ensure future compatibility with the reference implementation, it would be worth considering reimplementing schema builder and extender completely.

### DIFF
--- a/src/Utils/ASTDefinitionBuilder.php
+++ b/src/Utils/ASTDefinitionBuilder.php
@@ -34,7 +34,6 @@ use GraphQL\Type\Definition\OutputType;
 use GraphQL\Type\Definition\Type;
 use GraphQL\Type\Definition\UnionType;
 use function is_array;
-use function is_string;
 use Throwable;
 
 /**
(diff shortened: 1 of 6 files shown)


src/Utils/ASTDefinitionBuilder.php
```diff
@@ -162,6 +161,17 @@ public function buildType($ref): Type
...
 
+    /**
+     * Calling this method is an equivalent of `typeMap[typeName]` in `graphql-js`.
+     * It is legal to access a type from the map of already-built types that doesn't exist in the map.
+     * Since we build types lazily, and we don't have a such map of built types,
+     * this method provides a way to build a type that may not exist in the SDL definitions and returns null instead.
+     */
+    public function maybeBuildType(string $name): ?Type
+    {
+        return isset($this->typeDefinitionsMap[$name]) ? $this->buildType($name) : null;
```
spawnia, line 172:
```suggestion
        return isset($this->typeDefinitionsMap[$name])
            ? $this->buildType($name)
            : null
```

spawnia:
Awesome work! This solution is really simple and powerful.

PR #1069, opened 2022-02-15
Allow iterable in `PromiseAdapter::all()`

simPod:
I need to write my own PromiseAdapter that can index by other types than `int|string` (since array is used to index values in SyncPromiseAdapter).

My `$promisesOrValues` is of type \Ds\Map. Widening the interface method arg would allow me to pass it into my implementation.

### DIFF
--- a/src/Executor/Promise/Adapter/AmpPromiseAdapter.php
+++ b/src/Executor/Promise/Adapter/AmpPromiseAdapter.php
@@ -8,8 +8,10 @@ use function Amp\Promise\all;
 use Amp\Promise as AmpPromise;
 use Amp\Success;
 use function array_replace;
+use function assert;
 use GraphQL\Executor\Promise\Promise;
 use GraphQL\Executor\Promise\PromiseAdapter;
+use function is_array;
 use Throwable;
 
 class AmpPromiseAdapter implements PromiseAdapter
@@ -77,8 +79,13 @@ class AmpPromiseAdapter implements PromiseAdapter
         return new Promise($promise, $this);
     }
 
-    public function all(array $promisesOrValues): Promise
+    public function all(iterable $promisesOrValues): Promise
     {
+        assert(
+            is_array($promisesOrValues),
+            'AmpPromiseAdapter::all(): Argument #1 ($promisesOrValues) must be of type array'
+        );
+
         /** @var array<AmpPromise<mixed>> $promises */
         $promises = [];
         foreach ($promisesOrValues as $key => $item) {
--- a/src/Executor/Promise/Adapter/ReactPromiseAdapter.php
+++ b/src/Executor/Promise/Adapter/ReactPromiseAdapter.php
@@ -52,8 +52,13 @@ class ReactPromiseAdapter implements PromiseAdapter
         return new Promise($promise, $this);
     }
 
-    public function all(array $promisesOrValues): Promise
+    public function all(iterable $promisesOrValues): Promise
     {
+        assert(
+            is_array($promisesOrValues),
+            'ReactPromiseAdapter::all(): Argument #1 ($promisesOrValues) must be of type array'
+        );
+
         // TODO: rework with generators when PHP minimum required version is changed to 5.5+
 
         foreach ($promisesOrValues as &$promiseOrValue) {
--- a/src/Executor/Promise/Adapter/SyncPromiseAdapter.php
+++ b/src/Executor/Promise/Adapter/SyncPromiseAdapter.php
@@ -2,12 +2,14 @@
 
 namespace GraphQL\Executor\Promise\Adapter;
 
+use function assert;
 use function count;
 use GraphQL\Deferred;
 use GraphQL\Error\InvariantViolation;
 use GraphQL\Executor\Promise\Promise;
 use GraphQL\Executor\Promise\PromiseAdapter;
 use GraphQL\Utils\Utils;
+use function is_array;
 use Throwable;
 
 /**
@@ -72,8 +74,13 @@ class SyncPromiseAdapter implements PromiseAdapter
         return new Promise($promise->reject($reason), $this);
     }
 
-    public function all(array $promisesOrValues): Promise
+    public function all(iterable $promisesOrValues): Promise
     {
+        assert(
+            is_array($promisesOrValues),
+            'SyncPromiseAdapter::all(): Argument #1 ($promisesOrValues) must be of type array'
+        );
+
         $all = new SyncPromise();
 
         $total = count($promisesOrValues);
--- a/src/Executor/Promise/PromiseAdapter.php
+++ b/src/Executor/Promise/PromiseAdapter.php
@@ -63,12 +63,12 @@ interface PromiseAdapter
     public function createRejected(Throwable $reason): Promise;
 
     /**
-     * Given an array of promises (or values), returns a promise that is fulfilled when all the
-     * items in the array are fulfilled.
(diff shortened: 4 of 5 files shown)


src/Executor/Promise/Adapter/AmpPromiseAdapter.php
```diff
@@ -77,8 +79,13 @@ public function createRejected(Throwable $reason): Promise
         return new Promise($promise, $this);
     }
 
-    public function all(array $promisesOrValues): Promise
+    public function all(iterable $promisesOrValues): Promise
     {
+        assert(
+            is_array($promisesOrValues),
+            'AmpPromiseAdapter::all() requires arg to of array type'
+        );
```
spawnia, line 87: I think we can avoid this gotcha by replacing the `array_replace()` usage below with an implementation using `foreach`

src/Executor/Promise/Adapter/ReactPromiseAdapter.php
```diff
@@ -52,8 +52,13 @@ public function createRejected(Throwable $reason): Promise
         return new Promise($promise, $this);
     }
 
-    public function all(array $promisesOrValues): Promise
+    public function all(iterable $promisesOrValues): Promise
     {
+        assert(
+            is_array($promisesOrValues),
+            'ReactPromiseAdapter::all() requires arg of array type'
+        );
```
spawnia, line 60: Does not seem necessary

src/Executor/Promise/Adapter/SyncPromiseAdapter.php
```diff
@@ -72,8 +74,13 @@ public function createRejected(Throwable $reason): Promise
         return new Promise($promise->reject($reason), $this);
     }
 
-    public function all(array $promisesOrValues): Promise
+    public function all(iterable $promisesOrValues): Promise
     {
+        assert(
+            is_array($promisesOrValues),
+            'SyncPromiseAdapter::all() requires arg to of array type'
+        );
```
spawnia, line 82: I think `iterable` should work here without modification

spawnia (changes requested):

src/Executor/Promise/Adapter/SyncPromiseAdapter.php
simPod: There's indexing into array below `$result[$index]`

src/Executor/Promise/Adapter/ReactPromiseAdapter.php
simPod: Reactphp uses array currently https://github.com/reactphp/promise/blob/234f8fd1023c9158e2314fa9d7d0e6a83db42910/src/functions.php#L78

src/Executor/Promise/Adapter/SyncPromiseAdapter.php
spawnia: So? `$result` is a newly built `array`

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
simPod: There's indexing into array below: `$promises[$key]` so key must be of `int|string`

src/Executor/Promise/Adapter/ReactPromiseAdapter.php
spawnia: Ok, luckily there is `iterable_to_array()`

src/Executor/Promise/Adapter/SyncPromiseAdapter.php
simPod: editted the comment. There's restriction of key type to int string so array is reasonable.

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
spawnia: Cool, we can document that and allow `iterable<array-key, ?>`-

src/Executor/Promise/Adapter/ReactPromiseAdapter.php
simPod: That fails when key is not `int|string`

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
simPod: Since it breaks covariance i'd rather not put it into method signature.

I'd keep arrays as they're in graphql-php implementations. We can eventually use better data structure in SyncPromise. for Amp and Reactphp we will have to wait for upstreams to accept iterable as well.

simPod: See https://github.com/amphp/amp/blob/591e06ec427c2d2e5018172552b9354c208419e7/lib/functions.php#L387, we cannot help it much here now.

spawnia: When we violate covariance, we can provide at least some chance of catching an error statically in the callsite.

spawnia: For example, if `AmpPromiseAdapter::all()` were annotated to accept `iterable<array-key, ?>` and it was called with an `iterable<Foo, ?>`. Static analysis could catch such an error.

simPod: I don't think so since `AmpPromiseAdapter::all()` is never called I guess, interface `PromiseAdapter::all()` is usually used. Therefore, `assert()` is more suitable here so at least tests can catch it. WDYT?

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
```diff
@@ -77,8 +79,13 @@ public function createRejected(Throwable $reason): Promise
         return new Promise($promise, $this);
     }
 
-    public function all(array $promisesOrValues): Promise
+    public function all(iterable $promisesOrValues): Promise
     {
+        assert(
+            is_array($promisesOrValues),
+            'AmpPromiseAdapter::all() requires arg to of array type'
```
spawnia, line 86:
```suggestion
            'AmpPromiseAdapter::all(): Argument #1 ($promisesOrValues) must be of type array'
```

This is closer to the native PHP error, see https://3v4l.org/dvLAv

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
spawnia: Yeah right, it is not really called directly. The passed in value should always be in control of the developer and not change at runtime?

src/Executor/Promise/PromiseAdapter.php
```diff
@@ -66,9 +66,9 @@ public function createRejected(Throwable $reason): Promise;
      * Given an array of promises (or values), returns a promise that is fulfilled when all the
      * items in the array are fulfilled.
      *
-     * @param array<Promise|mixed> $promisesOrValues
+     * @param iterable<Promise|mixed> $promisesOrValues
```
spawnia, line 69: The description is now out of date

src/Executor/Promise/Adapter/AmpPromiseAdapter.php
simPod: Yup if I understand the question. I e.g. produce it through https://github.com/overblog/dataloader-php

PR #1073, opened 2022-02-18
Reorganize AST interfaces related to schema and type extensions

spawnia:
Fixes https://github.com/webonyx/graphql-php/pull/999#discussion_r809811050


vhenzl (approved):
LGTM 👍

PR #1075, opened 2022-02-19
Align `Utils::suggestionList()` with the reference implementation

vhenzl:
Aligns `Utils::suggestionList()` with the [current version of `suggestionList()`](https://github.com/graphql/graphql-js/blob/v16.3.0/src/jsutils/suggestionList.ts) from the reference implementation.

Adds the `LexicalDistance` class to calculate the distance between strings. The built-in `levenshtein` function can't be used as it doesn't support transpositions. 

Options with the same length are ordered lexicographically.

See https://github.com/webonyx/graphql-php/pull/999#discussion_r809839779 for context.


src/Utils/LexicalDistance.php
```diff
@@ -0,0 +1,133 @@
...
+
+        $aLength = \count($a);
+        $bLength = \count($b);
+
+        if ($aLength - $bLength > $threshold) {
+            return null;
+        }
+
+        $rows = &$this->rows;
+        for ($j = 0; $j <= $bLength; ++$j) {
```
spawnia, line 79: Why not `$i`, given we are not in a nested for loop?

src/Utils/LexicalDistance.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->input = $input;
+        $this->inputLowerCase = \strtolower($input);
+        $this->inputArray = self::stringToArray($this->inputLowerCase);
+
+        $length = \mb_strlen($input);
+        $this->rows = [
+            \array_fill(0, $length, 0),
+            \array_fill(0, $length, 0),
+            \array_fill(0, $length, 0),
+        ];
```
spawnia, line 46: Do we need to do this, given PHP can auto-populate intermediary arrays when populating nested values?

src/Utils/LexicalDistance.php
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+
+        $distance = $rows[$aLength % 3][$bLength];
+
+        return $distance <= $threshold ? $distance : null;
+    }
+
+    /**
+     * @return array<int>
```
spawnia, line 122:
```suggestion
     * Returns a list of char codes in the given string.
     *
     * @return list<int>
```

tests/Type/EnumTypeTest.php
```diff
@@ -348,8 +348,7 @@ public function testDoesNotAcceptValuesWithIncorrectCasing(): void
             '{ colorEnum(fromEnum: green) }',
             null,
             [
-                // Improves upon the reference implementation
```
spawnia, line 351: Too bad we lost that improvement, but it does not matter too much. Ease of maintenance trumps this IMO.

src/Utils/LexicalDistance.php
```diff
@@ -0,0 +1,133 @@
...
+            $tmp = $a;
+            $a = $b;
+            $b = $tmp;
+        }
+
+        $aLength = \count($a);
+        $bLength = \count($b);
+
+        if ($aLength - $bLength > $threshold) {
+            return null;
```
spawnia, line 75: What does this mean?

tests/Utils/SuggestionListTest.php
```diff
@@ -7,38 +7,90 @@
...
+     * @see describe('suggestionList')
      * @see it('Returns results when input is empty')
      */
     public function testResturnsResultsWhenInputIsEmpty(): void
     {
-        self::assertEquals(
-            Utils::suggestionList('', ['a']),
-            ['a']
-        );
+        self::assertEquals(Utils::suggestionList('', ['a']), ['a']);
```
spawnia, line 16: Expected goes first, and I would prefer `assertSame()` here since order matters

spawnia (changes requested):

src/Utils/LexicalDistance.php
vhenzl: Copied from https://github.com/graphql/graphql-js/blob/v16.3.0/src/jsutils/suggestionList.ts#L88-L90, but yeah, `$i` should be fine here.

src/Utils/LexicalDistance.php
vhenzl: My first impression was that it would lead to `??` in the code but no, you are right. It's enough to initialise the `rows` property with an empty array.

I'm now wondering if sharing the `rows` array between multiple `measure()` calls actually makes sense at all and if a local `$rows` variable initialised in the place of `$rows = &$this->rows;` wouldn't be better. But I'm leaving the property there for now in order not to divert too much for the JS code.

src/Utils/LexicalDistance.php
vhenzl: An early return, optimisation: https://github.com/graphql/graphql-js/blob/v16.3.0/src/jsutils/suggestionList.ts#L84

`levenshtein` function always returns `int`, whereas `LexicalDistance::measure()` returns `null|int`. In contrast with `levenshtein()`, `measure()` takes into account the threshold and returns `null` if the distance is bigger (see changes to `Utils::suggestionList()`, this logic was previously there).

tests/Utils/SuggestionListTest.php
vhenzl: This particular test was like that before, I only changed formatting, but you are right nonetheless. I have fixed it everywhere.

The different order of parameters in _assertions_ and _expectations_ is pretty annoying. Have you considered using https://github.com/pestphp/pest for this project? It would allow us to keep the same structure of tests as in the reference implementation.

spawnia: Great idea, happy to try - see https://github.com/webonyx/graphql-php/issues/1086

src/Utils/LexicalDistance.php
spawnia: Thanks for the explanation, I would like to see that in a comment.

src/Utils/LexicalDistance.php
spawnia: I would rater see it simplified. The JS code seems to be optimized for performance in a way that does not apply to PHP - in such cases I opted to go for optimized PHP code instead. The biggest value the reference implementation provides is the test suite and a starting point for the implementation.

spawnia (approved):
Nicely done. Thanks for working with me to adapt this code to idiomatic PHP.

PR #1081, opened 2022-03-03
Add ReturnTypeWillChange attribute to disable deprecation warnings on PHP 8.1

emodric:
Hi,

I realize this is a really old version, but it is one of the dependencies in our stack which we want to run on PHP 8.1 (and which otherwise works fine on 8.1), and the deprecation emmited by PHP 8.1 is breaking some of our GraphQL responses.

There may be more of these needed, but I haven't run into them so far.

Would it be possible to tag a quick patch release with this fix?

Thanks!

### DIFF
--- a/src/Type/Definition/Type.php
+++ b/src/Type/Definition/Type.php
@@ -359,6 +359,7 @@ abstract class Type implements JsonSerializable
     /**
      * @return string
      */
+    #[\ReturnTypeWillChange]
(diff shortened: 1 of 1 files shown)


spawnia:
We no longer support pre-14 versions, see https://github.com/webonyx/graphql-php/commit/b49ffe990f1876e24660c6f2af7296c6dc8ce037

PR #1099, opened 2022-03-31
Add Path type to ResolveInfo

simPod:
I'm starting to work with Path in my projects and don't really want to use `array<int, string|int>` everywhere. E.g. when we start using `list` type, I'd like it to change automatically with graphql-php version bump. Therefore, I'm introducing Path type I can import from this lib.

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -52,6 +52,7 @@ use Throwable;
 
 /**
  * @phpstan-import-type FieldResolver from Executor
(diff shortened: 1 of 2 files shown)


spawnia:
Do we consider type aliases part of the public API and guarantee not breaking them? 🤔

simPod:
Interesting question. I'd wait a few more years until it's stable and fully adopted by the ecosystem and consider it "on your own risk" for now.

spawnia:
Sounds good!

Issue #1093, opened 2022-03-28
Lazy field definitions are eager loaded when schema has no type loader


NikolaiEngel:
Hey, 

i have created the following sample code to prove that the lazy loading feature does not work as desired.

If I query the field `Test2`, the class `TestInvokable` is created despite lazy loading. This has the consequence that a lot of dependencies have to be resolved, which are not necessary for the query. In this example this is illustrated by a sleep of 3 seconds. 

I have also created an alternative TestCase in which it works. When I set up the `Schema` as in the linked page: https://webonyx.github.io/graphql-php/schema-definition-language/, the `TestInvokable` class is not created. Since there are two ways to use the GraphQL, I would like to have this feature also for the scenario described below.

```
// index.php
$rawInput = file_get_contents('php://input');
$input = json_decode($rawInput, true);
$query = $input['query'] ?? '';
$variableValues = $input['variables'] ?? null;
$operationName = $input['operationName'] ?? null;

$queryType = $container->get(QueryType::class);

$schema = new Schema(['query' => $queryType]);

// execute query
$result = GraphQL::executeQuery(
    $schema,
    $query,
    null,
    null,
    $variableValues,
    $operationName
)->toArray();

echo json_encode($result);
```
```
// QueryType.php
class QueryType extends ObjectType
{

    private const TEST1_FIELD = 'Test1';
    private const TEST2_FIELD = 'Test2';

    private ContainerInterface $container;

    public function __construct(ContainerInterface $container)
    {
        $this->container = $container;
        parent::__construct($this->getConfig());
    }

    public function getConfig(): array
    {
        return [
            'fields' => [
                self::TEST1_FIELD => fn() => [
                    'description' => 'My Test Field 1',
                    'type' => self::string(),
                    'resolve' => $this->container->get(TestInvokable::class),
                ],
                self::TEST2_FIELD => fn() => [
                    'description' => 'My Test Field 2',
                    'type' => self::string(),
                    'resolve' => fn() => 'Hello World',
                ]
            ],
        ];
    }
}
```
```
// Test Invokable
class TestInvokable
{

    public function __construct()
    {
        // Heavy dependencies
        sleep(3);
    }

    public function __invoke($source, array $args): string
    {
        return 'FooBar';
    }
}
```

Linked issue is a feature request to create a DI container known to GraphQL to use invokable classes for resolution. https://github.com/webonyx/graphql-php/issues/1088

spawnia:
[...]

PR #1104, opened 2022-04-07
Avoid eager-loading schema unnecessarily

spawnia:
Resolves https://github.com/webonyx/graphql-php/issues/1093

### DIFF
--- a/src/Utils/SchemaExtender.php
+++ b/src/Utils/SchemaExtender.php
@@ -57,6 +57,139 @@ class SchemaExtender
     protected static ASTDefinitionBuilder $astBuilder;
 
     /**
+     * @param array<string, bool> $options
+     * @phpstan-param TypeConfigDecorator|null $typeConfigDecorator
+     *
+     * @api
+     */
+    public static function extend(
+        Schema $schema,
+        DocumentNode $documentAST,
+        array $options = [],
+        ?callable $typeConfigDecorator = null
+    ): Schema {
+        // Ensure the schema is fully loaded
+        $schema->getTypeMap();
+
+        if (
+            ! ($options['assumeValid'] ?? false)
+            && ! ($options['assumeValidSDL'] ?? false)
+        ) {
+            DocumentValidator::assertValidSDLExtension($documentAST, $schema);
+        }
+
+        /** @var array<string, Node&TypeDefinitionNode> $typeDefinitionMap */
+        $typeDefinitionMap = [];
+
+        static::$typeExtensionsMap = [];
+
+        /** @var array<int, DirectiveDefinitionNode> $directiveDefinitions */
+        $directiveDefinitions = [];
+
+        /** @var SchemaDefinitionNode|null $schemaDef */
+        $schemaDef = null;
+
+        /** @var array<int, SchemaExtensionNode> $schemaExtensions */
+        $schemaExtensions = [];
+
+        foreach ($documentAST->definitions as $def) {
+            if ($def instanceof SchemaDefinitionNode) {
+                $schemaDef = $def;
+            } elseif ($def instanceof SchemaExtensionNode) {
+                $schemaExtensions[] = $def;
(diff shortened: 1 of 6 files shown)


src/Utils/SchemaExtender.php
```diff
@@ -56,6 +56,139 @@ class SchemaExtender
 
     protected static ASTDefinitionBuilder $astBuilder;
 
+    /**
+     * @param array<string, bool> $options
+     * @phpstan-param TypeConfigDecorator|null $typeConfigDecorator
+     *
+     * @api
+     */
+    public static function extend(
```
vhenzl, line 65: Is it necessary to move things around like this? It makes it hard to see what has actually changed…

tests/Utils/SchemaExtenderTest.php
```diff
@@ -352,26 +391,32 @@ enum SomeEnum {
...
-            $queryType->getField('someUnion')->getType(),
-            $extendedSchema->getType('SomeUnion')
+            <<<'EOF'
+union SomeUnion = Foo | Biz | Bar
+EOF,
+            self::printSchemaChanges($schema, $extendedSchema),
         );
     }
 
+    // TODO continue converting tests from here
```
spawnia, line 418: TODO finish before merging.

vhenzl: After #1139, we will be much closer to merging #928 and #929; it should help you here (all these tests have already been converted there).

spawnia: Thank you, I included https://github.com/webonyx/graphql-php/pull/929

spawnia:
@vhenzl do you have an idea why the extended schema order differs from graphql-js and how we can fix it?

vhenzl:
@spawnia, not sure what you mean. Different order of types in a printed schema? It would be https://github.com/webonyx/graphql-php/issues/954.

spawnia:
Some test cases are failing due to the order of the printed output, see https://github.com/webonyx/graphql-php/runs/6302953843?check_suite_focus=true

I went ahead and removed sorting from the `SchemaPrinter`, now some tests fail due to the changed output which should be trivial to fix.

tests/Utils/SchemaExtenderTest.php
vhenzl: @spawnia That's definitely not what I meant! It makes this PR huge and impossible to review. Which changes are to solve the eager-loading problem and which are to address the ordering issue? Impossible to say.

`tests/Utils/SchemaExtenderLegacyTest.php` shouldn't be included. It has been merged to master in #977, addressed in consequent PRs, and finally removed in #1111. We can only guess how many other merge conflicts were resolved incorrectly… I have been working for months to get #929 (and #928) merged nicely and clearly and you do what with it? A big ball of mud…

spawnia:
@vhenzl can you review this?

ruudk:
@spawnia After using `dev-master` and dumping the schema, the types are no longer sorted A-Z.

Why are the types no longer sorted? This is super useful for diffing the persisted schema that is stored in git. We do that for example so that other projects can also pull in the schema.

When the types are now random or based on how they are discovered, it's gonna trigger more changes from time to time without no reason.

spawnia:
The definition order of the schema is now kept. This is how the reference implementation does it now. We generally follow them, it makes maintenance much easier.

ruudk:
What if we bring this back as an option to the `SchemaPrinter`?

```php
SchemaPrinter::doPrint($schema, ['sort' => true])
```

It will then [run this again](https://github.com/webonyx/graphql-php/pull/1104/commits/81e956fc26f006d3089c549d1f3bc602d0ffaacd):
```
$types = $schema->getTypeMap();
ksort($types);
$types = array_filter($types, $typeFilter);
```

ruudk:
@spawnia Brought back the sorting of types as an option to the SchemaPrinter: https://github.com/webonyx/graphql-php/pull/1152

PR #1109, opened 2022-04-18
Extract check for unique directive names into separate rule

vhenzl:
Introduces `UniqueDirectiveNames` validation rule.

Corresponding change in the reference implementation: https://github.com/graphql/graphql-js/commit/c1745228b2ae5ec89b8de36ea766d544607e21ea

### DIFF
--- a/src/Validator/Rules/UniqueDirectiveNames.php
+++ b/src/Validator/Rules/UniqueDirectiveNames.php
@@ -1,6 +1,4 @@
-<?php
-
-declare(strict_types=1);
+<?php declare(strict_types=1);
 
 namespace GraphQL\Validator\Rules;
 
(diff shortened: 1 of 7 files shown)


src/Validator/Rules/UniqueDirectiveNames.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
```
spawnia, line 3:
```suggestion
<?php declare(strict_types=1);
```

tests/Validator/UniqueDirectiveNamesTest.php
```diff
@@ -0,0 +1,163 @@
+<?php
+
+declare(strict_types=1);
```
spawnia, line 3:
```suggestion
<?php declare(strict_types=1);
```

spawnia (changes requested):

vhenzl:
This should be good now.

spawnia (approved):

PR #1110, opened 2022-04-18
Extract check for possible extensions into a separate rule

vhenzl:
Introduces `PossibleTypeExtensions` validation rule.

Corresponding change in the reference implementation: https://github.com/graphql/graphql-js/commit/e6a3f08cc92594f68a6e61d3d4b46a6d279f845e


src/Validator/Rules/PossibleTypeExtensions.php
```diff
@@ -0,0 +1,155 @@
...
+ * A type extension is only valid if the type is defined and has the same kind.
+ */
+class PossibleTypeExtensions extends ValidationRule
+{
+    public function getSDLVisitor(SDLValidationContext $context): array
+    {
+        $schema = $context->getSchema();
+        /** @var array<string, TypeDefinitionNode&Node> $definedTypes */
+        $definedTypes = [];
+
```
spawnia, line 36:
```suggestion

        /** @var array<string, TypeDefinitionNode&Node> $definedTypes */
        $definedTypes = [];
```

src/Validator/Rules/PossibleTypeExtensions.php
```diff
@@ -0,0 +1,155 @@
...
+        foreach ($context->getDocument()->definitions as $def) {
+            if ($def instanceof TypeDefinitionNode) {
+                $definedTypes[$def->name->value] = $def;
+            }
+        }
+
+        $checkTypeExtension = static function ($node) use ($context, $schema, &$definedTypes): ?VisitorOperation {
+            $typeName = $node->name->value;
+            $defNode = $definedTypes[$typeName] ?? null;
+            $existingType = $schema !== null ? $schema->getType($typeName) : null;
```
spawnia, line 46: Ternary should always be multiline, I just documented this in `CONTRIBUTING.md`

src/Validator/Rules/PossibleTypeExtensions.php
```diff
@@ -0,0 +1,155 @@
...
+            NodeKind::SCALAR_TYPE_EXTENSION => $checkTypeExtension,
+            NodeKind::OBJECT_TYPE_EXTENSION => $checkTypeExtension,
+            NodeKind::INTERFACE_TYPE_EXTENSION => $checkTypeExtension,
+            NodeKind::UNION_TYPE_EXTENSION => $checkTypeExtension,
+            NodeKind::ENUM_TYPE_EXTENSION => $checkTypeExtension,
+            NodeKind::INPUT_OBJECT_TYPE_EXTENSION => $checkTypeExtension,
+        ];
+    }
+
+    private static function defKindToExtKind(string $kind): string
```
spawnia, line 96: Is this duplicated elsewhere? Same with the other conversion functions in here, they seem like they might be needed in multiple places.

tests/Validator/PossibleTypeExtensionsTest.php
```diff
@@ -0,0 +1,437 @@
+<?php
+
+declare(strict_types=1);
```
spawnia, line 3:
```suggestion
<?php declare(strict_types=1);
```

tests/Validator/PossibleTypeExtensionsTest.php
```diff
@@ -0,0 +1,437 @@
...
+
+namespace GraphQL\Tests\Validator;
+
+use GraphQL\Type\Schema;
+use GraphQL\Utils\BuildSchema;
+use GraphQL\Validator\Rules\PossibleTypeExtensions;
+
+class PossibleTypeExtensionsTest extends ValidatorTestCase
```
spawnia, line 11:
```suggestion
final class PossibleTypeExtensionsTest extends ValidatorTestCase
```

src/Validator/Rules/PossibleTypeExtensions.php
```diff
@@ -0,0 +1,155 @@
+<?php
+
+declare(strict_types=1);
```
spawnia, line 3:
```suggestion
<?php declare(strict_types=1);
```

spawnia (changes requested):

src/Validator/Rules/PossibleTypeExtensions.php
vhenzl: What is the suggestion here? The code is the same.

src/Validator/Rules/PossibleTypeExtensions.php
vhenzl: I honestly can't see how the multiline version will make the code more readable or maintainable in this case. Anyway, is it possible to enforce it for php-cs-fixer as was for php_codesniffer in https://github.com/webonyx/graphql-php/pull/951?

src/Validator/Rules/PossibleTypeExtensions.php
vhenzl: I don't know about any other place. Right now it's needed only in this rule.

tests/Validator/PossibleTypeExtensionsTest.php
vhenzl: Not sure how this happened, my PHPStorm's settings had to get messed up, I normally use a single line for this.  Why php-cs-fixer doesn't report/fix it?

src/Validator/Rules/PossibleTypeExtensions.php
spawnia: Unfortunately not.

In this case, one advantage would be that an error thrown in `$schema->getType()` would have a more specific line number when debugging.

In general, it is a matter of consistency for me.

tests/Validator/PossibleTypeExtensionsTest.php
vhenzl: Again, if this is a rule now,it would be great to enforce it with php-cs-fixer/phpstan.

tests/Validator/PossibleTypeExtensionsTest.php
spawnia: There is no setting to enforce this, it is only possible for now to have php-cs-fixer *not* mess it up.

tests/Validator/PossibleTypeExtensionsTest.php
spawnia: The fixer for this does not allow scoping this rule to directories. I don't want final classes in the implementation.

src/Validator/Rules/PossibleTypeExtensions.php
spawnia: Whitespace. I prefer to have variables that are filled in a loop as close as possible to that loop.

vhenzl:
This should be good now.

spawnia (approved):

PR #1111, opened 2022-04-18
Extract check for unique field names into separate rule

vhenzl:
Introduces `UniqueFieldDefinitionNames` validation rule.

Corresponding change in the reference implementation: https://github.com/graphql/graphql-js/commit/6be63f0fda6947a228d7119bdd17000831d688df


src/Validator/Rules/UniqueFieldDefinitionNames.php
```diff
@@ -0,0 +1,90 @@
...
+            if (! isset($knownFieldNames[$typeName])) {
+                $knownFieldNames[$typeName] = [];
+            }
+
+            $fieldNames = &$knownFieldNames[$typeName];
+
+            foreach ($node->fields as $fieldDef) {
+                $fieldName = $fieldDef->name->value;
+
+                $existingType = $schema !== null ? $schema->getType($typeName) : null;
```
spawnia, line 49:
```suggestion
                $existingType = $schema === null
                    ? null
                    : $schema->getType($typeName);
```

src/Validator/Rules/UniqueFieldDefinitionNames.php
```diff
@@ -0,0 +1,90 @@
...
+            NodeKind::INTERFACE_TYPE_EXTENSION => $checkFieldUniqueness,
+            NodeKind::OBJECT_TYPE_DEFINITION => $checkFieldUniqueness,
+            NodeKind::OBJECT_TYPE_EXTENSION => $checkFieldUniqueness,
+        ];
+    }
+
+    private static function hasField(?NamedType $type, string $fieldName): bool
+    {
+        if ($type instanceof ObjectType || $type instanceof InterfaceType || $type instanceof InputObjectType) {
+            return isset($type->getFields()[$fieldName]);
```
spawnia, line 85:
```suggestion
            return $type->hasField($fieldName);
```

src/Validator/Rules/UniqueFieldDefinitionNames.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
```
spawnia, line 3:
```suggestion
<?php declare(strict_types=1);
```

tests/Validator/UniqueFieldDefinitionNamesTest.php
```diff
@@ -0,0 +1,508 @@
+<?php
+
+declare(strict_types=1);
```
spawnia, line 3:
```suggestion
<?php declare(strict_types=1);
```

tests/Validator/UniqueFieldDefinitionNamesTest.php
```diff
@@ -0,0 +1,508 @@
...
+
+namespace GraphQL\Tests\Validator;
+
+use GraphQL\Type\Schema;
+use GraphQL\Utils\BuildSchema;
+use GraphQL\Validator\Rules\UniqueFieldDefinitionNames;
+
+class UniqueFieldDefinitionNamesTest extends ValidatorTestCase
```
spawnia, line 11:
```suggestion
final class UniqueFieldDefinitionNamesTest extends ValidatorTestCase
```

spawnia (changes requested):
Looks pretty good overall

src/Validator/Rules/UniqueFieldDefinitionNames.php
```diff
@@ -0,0 +1,90 @@
...
+ */
+class UniqueFieldDefinitionNames extends ValidationRule
+{
+    public function getSDLVisitor(SDLValidationContext $context): array
+    {
+        $schema = $context->getSchema();
+        /** @var array<string, NameNode[]> $knownFieldNames */
+        $knownFieldNames = [];
+        $checkFieldUniqueness = static function ($node) use ($context, $schema, &$knownFieldNames): VisitorOperation {
+            assert($node instanceof InputObjectTypeDefinitionNode || $node instanceof InterfaceTypeDefinitionNode || $node instanceof ObjectTypeDefinitionNode || $node instanceof InputObjectTypeExtensionNode || $node instanceof InterfaceTypeExtensionNode || $node instanceof ObjectTypeExtensionNode);
```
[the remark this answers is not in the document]
vhenzl: No, there is no common interface. They are the 6 types of nodes supported by this rule: https://github.com/webonyx/graphql-php/pull/1111/files#diff-b20b47356cdad418fb8a83dd0e978923a68e02857888235e4b6fa9a03c5d303dR73-R78

Since the union can't be used as the parameter type and `$node` is just `mixed`, I added it here to get help from IDE and PHPStan while working on the code. And forgot to remove it afterwards. 

Unless you want me to keep it there, I will remove the `assert` before the PR is ready.  The problem with `$node` being `mixed` is common to all/most rules. We had a chat on this topic a while back: https://github.com/webonyx/graphql-php/pull/1111/files#diff-b20b47356cdad418fb8a83dd0e978923a68e02857888235e4b6fa9a03c5d303dR73-R78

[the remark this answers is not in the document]
spawnia: Got it, thanks for the explanation.

The intent behind my question was to see if we could define the code to be more self-explanatory that way. For example, if the assert were `assert($node instanceof NodeWithFields)`, it would be obvious why those node types were selected.

I am fine with this implementation and would like to keep the `assert()` - perhaps formatted to be split across multiple lines for readiblilty.

src/Validator/Rules/UniqueFieldDefinitionNames.php
```diff
@@ -0,0 +1,90 @@
...
+            $typeName = $node->name->value;
+
+            if (! isset($knownFieldNames[$typeName])) {
+                $knownFieldNames[$typeName] = [];
+            }
+
+            $fieldNames = &$knownFieldNames[$typeName];
```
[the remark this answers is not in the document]
vhenzl: This suggestion is invalid, only variables can be assigned by reference. Must be:
```
$knownFieldNames[$typeName] ??= [];
$fieldNames = &$knownFieldNames[$typeName];
```

spawnia (approved):

PR #1119, opened 2022-04-25
Run phpstan only on one php version

simPod:
Running phpstan 3 times produces 3 inline comments

image

Those can be e.g. filtered to be produced by a single PHP version run only but I wonder what is the reason to run in on multiple versions in the first place. AFAIK it will not show different errors when run on different php versions since we have PHP 7.4 code.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -11,13 +11,6 @@ jobs:
     name: "Static Analysis with PHPStan"
     runs-on: "ubuntu-20.04"
 
-    strategy:
-      matrix:
-        php-version:
-          - "7.4"
-          - "8.0"
-          - "8.1"
-
     steps:
       - name: "Checkout code"
         uses: "actions/checkout@v2"
@@ -26,7 +19,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          php-version: "${{ matrix.php-version }}"
+          php-version: "7.4"
           tools: "cs2pr"
 
       - name: "Install dependencies with Composer"
@@ -39,11 +32,6 @@ jobs:
     name: "Static Analysis with Psalm"
     runs-on: "ubuntu-20.04"
 
-    strategy:
-      matrix:
-        php-version:
-          - "8.1"
-
     steps:
       - name: "Checkout code"
         uses: "actions/checkout@v2"
@@ -52,7 +40,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          php-version: "${{ matrix.php-version }}"
+          php-version: "7.4"
           tools: "cs2pr"
 
       - name: "Install dependencies with Composer"


mfn:
> AFAIK it will not show different errors when run on different php versions since we have PHP 7.4 code.

In my experience from rebing/graphql-laravel it actually does. Some errors were only reported when running on PHP8, I therefore limited to only a single/latest version.

spawnia:
I am fine with this. The workflow definition can be simplified to not use `matrix` then.

simPod:
@mfn when you stumble upon this again, can you show example? 

@spawnia I'll change the php version to 7.4 though. E.g. on 8.0+ there might be some changes to return types (sprintf etc. not returning `false` on failure)

spawnia:
@simPod let's actively try and provoke a PHPStan failure that only occurs when running on a specific PHP version: https://github.com/webonyx/graphql-php/pull/1124

mfn:
@simPod no, I could not find a way to satisfy it (unless I would use per-php version phpstan configs and therefore baseline files), this I just kept it at one (the latest) version.

simPod:
@mfn so is there actually a point to run it on multiple versions if errors cannot easily co-exist in baseline? We can't solve it anyway (the effort is not worth it).

mfn:
> so is there actually a point to run it on multiple versions if errors cannot easily co-exist in baseline?

I couldn't see one 🤷🏼

spawnia:
> is there actually a point to run it on multiple versions if errors cannot easily co-exist in baseline

I think this is a logical fallacy. The point of running on multiple versions is to catch errors. Only if it were proven to catch no additional errors could we omit it without any downside.

I actually found a case where the 7.4 analysis caught an incompatibility that the 8.0 analysis did not catch (usage of `: static`):
- https://github.com/webonyx/graphql-php/runs/6172372984?check_suite_focus=true
- https://github.com/webonyx/graphql-php/runs/6172373068?check_suite_focus=true

Same with usage of named arguments, caught only in the PHP 7.4 analysis: https://github.com/webonyx/graphql-php/runs/6172412791?check_suite_focus=true

Null-safe access: https://github.com/webonyx/graphql-php/runs/6172490811?check_suite_focus=true

Only the 8+ analyses caught that `each()` was removed:
- https://github.com/webonyx/graphql-php/runs/6172531542?check_suite_focus=true
- https://github.com/webonyx/graphql-php/runs/6172514866?check_suite_focus=true

It may be difficult in certain corner cases, but that is expected when maintaining compatibility with multiple major versions of a language, but it is definitely useful.

The aesthetics of inline comments matter less than ensuring correctness. If PHPStan adds an option to analyse multiple PHP versions at once and produce a nicer output, I am happy to add that. Until then, let's keep making sure this library is error free.

simPod:
> I actually found a case where the 7.4 analysis caught an incompatibility that the 8.0 analysis did not catch (usage of : static):
> Only the 8+ analyses caught that each() was removed:

That's why we run the code, e.g. phpunit handles these.

https://github.com/webonyx/graphql-php/runs/6172515206?check_suite_focus=true#step:5:698

The tests you've performed are php version incompatibility tests that are handled by phpunit php-version matrix and are not necessary to fail the same way with phpstan.

--------------

The only valid case was the one posted by @mfn but there's no easy way to solve it anyway. We would need `n` phpstan configurations for that. Therefore, the conclusion is that running phpstan on multiple php versions is not viable without some work overhead. It can be done but not worth.  
The thing we're doing right now is wrong in any case. If it spots any error (e.g. the case @mfn mentioned), we have no means to suppress or handle it anyways.

spawnia:
We do not have 100% test coverage, so PHPUnit does not catch every possible error.

I get the want to simplify and slim down CI, but in this instance it demonstrably worsens the analysis capabilities. I also don't see a hard reason why we must do this, I feel like the argument you are building up is a strawman. Currently static analysis passes, and I have not had issues with it that were impossible to overcome.

simPod:
Slimming the CI down is not a concern, it runs fast, it's free so whatever. I had noticed we have run it 3 times while we have no means to baseline errors reported by phpstan on different php versions anyways. We can run it on highest php version so it will catch issues introduced by php BC breaks.

spawnia:
> we have no means to baseline errors reported by phpstan on different php versions anyways

That would be nice to have, sure. Fortunately there are other ways to deal with errors: write universally compatible code, hand-write an ignore rule, ignore the entire line.

> We can run it on highest php version so it will catch issues introduced by php BC breaks.

That would not catch usage of PHP 8 only features, as shown by my examples.

PR #1120, opened 2022-04-25
Drop PHP 7.4

simPod:
I was wondering that for master we could drop PHP 7.4 already.


ruudk:
V15 is released since some time and works pretty well. Would it be a good moment to start developing V16?

spawnia:
I personally still have a need for PHP 7.4 compatibility and am not eager to drop it.

ruudk:
Sorry to ask, but are there any changes in your situation @spawnia? I'm currently working on #1426 and I feel quiet silly to not be able to use union types and have to fallback to no type safety at all. PHP 7.4 is EOL for a long time now. In my opinion it does not make any sense to keep supporting this.

spawnia:
I still have an important project to maintain that uses this library through https://github.com/spawnia/sailor, and probably will for a couple of years. I guess GraphQL is pretty slow to change aspects that affect the client, so I might not have to worry about backwards compatibility that much.

I can definitely see the advantages of using a newer PHP version for development. Perhaps we can implement something like https://getrector.com/blog/how-all-frameworks-can-bump-to-php-81-and-you-can-use-older-php to get the best of both?

ruudk:
Just wondering, but why does Sailor have to keep supporting PHP 7.4?

I don't really think it's worth the effort to downgrade this project with Rector. Gives you a lot of other issues to deal with, for what effort?

spawnia:
I just explained why, and I am starting to get annoyed by your repeated questioning of my reasons.

ruudk:
Alright, I misread it. Not my intention to annoy you ☺️

simPod:
@spawnia maybe the legacy project of yours can use Rector to downgrade the sailor lib? Or create sailor-legacy repo that autodowngrades sailor upstream lib.

.github/workflows/static-analysis.yml
```diff
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
-          - "7.4"
           - "8.0"
```
simPod, line 18: @spawnia I don't know what version do you need to support but I'd support only currently supported ones - 8.2+  
https://www.php.net/supported-versions.php

PR #1124, opened 2022-04-25
Provoke PHP 8 only PHPStan failure

spawnia:
https://github.com/webonyx/graphql-php/pull/1119


src/Executor/ExecutionResult.php
```diff
@@ -8,7 +8,6 @@
 use GraphQL\Error\Error;
 use GraphQL\Error\FormattedError;
 use JsonSerializable;
-use ReturnTypeWillChange;
```
simPod, line 11: this will provoke warning in php, unrelated to phpstan

spawnia: This was just one idea I had, feel free to push another attempt to this PR.

PR #1136, opened 2022-05-02
Add unique argument names validation rule

vhenzl:
Adds SDL validation rule `UniqueArgumentDefinitionNames`. 

Corresponding change in the reference implementation: https://github.com/graphql/graphql-js/commit/2170e4a485e067e68870aabb551f5cd44bd088d4

With this PR, `graphql-php` contains all SDL validation rules from `graphql-js`.


spawnia:
> With this PR, graphql-php contains all SDL validation rules from graphql-js.

🎉

src/Validator/Rules/UniqueArgumentDefinitionNames.php
```diff
@@ -0,0 +1,80 @@
...
+        $checkArgUniquenessPerField = static function ($node) use ($context): VisitorOperation {
+            assert(
+                $node instanceof InterfaceTypeDefinitionNode
+                || $node instanceof InterfaceTypeExtensionNode
+                || $node instanceof ObjectTypeDefinitionNode
+                || $node instanceof ObjectTypeExtensionNode
+            );
+
+            foreach ($node->fields as $fieldDef) {
+                self::checkArgUniqueness($node->name->value . '.' . $fieldDef->name->value, $fieldDef->arguments, $context);
```
spawnia, line 39:
```suggestion
                self::checkArgUniqueness("{$node->name->value}.{$fieldDef->name->value}", $fieldDef->arguments, $context);
```

src/Validator/Rules/UniqueArgumentDefinitionNames.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * @param NodeList<InputValueDefinitionNode> $arguments
+     */
+    private static function checkArgUniqueness(string $parentName, NodeList $arguments, SDLValidationContext $context): VisitorOperation
+    {
+        $seenArgs = [];
+        foreach ($arguments as $value) {
+            $seenArgs[$value->name->value] ??= [];
```
spawnia, line 63:
```suggestion
```

Should not be necessary, since you can do this in PHP:

```php
>>> $a = []
=> []
>>> $a['b']['c'][] = 'd' 
=> "d"
>>> $a
=> [
     "b" => [
       "c" => [
         "d",
       ],
     ],
   ]
```

src/Validator/Rules/UniqueArgumentDefinitionNames.php
```diff
@@ -0,0 +1,80 @@
...
+        ];
+    }
+
+    /**
+     * @param NodeList<InputValueDefinitionNode> $arguments
+     */
+    private static function checkArgUniqueness(string $parentName, NodeList $arguments, SDLValidationContext $context): VisitorOperation
+    {
+        $seenArgs = [];
+        foreach ($arguments as $value) {
```
spawnia, line 62:
```suggestion
        foreach ($arguments as $argument) {
```

src/Validator/Rules/UniqueArgumentDefinitionNames.php
```diff
@@ -0,0 +1,80 @@
...
+            $seenArgs[$value->name->value] ??= [];
+            $seenArgs[$value->name->value][] = $value;
+        }
+
+        foreach ($seenArgs as $argName => $argNodes) {
+            if (count($argNodes) > 1) {
+                $context->reportError(
+                    new Error(
+                        'Argument "' . $parentName . '(' . $argName . ':)" can only be defined once.',
+                        array_map(static fn (InputValueDefinitionNode $node) => $node->name, $argNodes),
```
spawnia, line 72: Is this mapping necessary?

spawnia (changes requested):

src/Validator/Rules/UniqueArgumentDefinitionNames.php
vhenzl: Right. Switching between PHP and JS all the time, I find this confusing, so I tend to initialize arrays explicitly.

src/Validator/Rules/UniqueArgumentDefinitionNames.php
vhenzl: Not really. Simple `$argNodes` works. Not sure if there are some nuances between passing an array of name nodes and definition nodes. The mapping comes from `graphlq-js`.

src/Validator/Rules/UniqueArgumentDefinitionNames.php
```diff
@@ -0,0 +1,78 @@
...
+                self::checkArgUniqueness("{$node->name->value}.{$fieldDef->name->value}", $fieldDef->arguments, $context);
+            }
+
+            return Visitor::skipNode();
+        };
+
+        return [
+            NodeKind::DIRECTIVE_DEFINITION => static function (DirectiveDefinitionNode $node) use ($context) {
+                return self::checkArgUniqueness('@' . $node->name->value, $node->arguments, $context);
+            },
```
spawnia, line 47:
```suggestion
            NodeKind::DIRECTIVE_DEFINITION => static function (DirectiveDefinitionNode $node): VisitorOperation => self::checkArgUniqueness("@{$node->name->value}", $node->arguments, $context),
```

src/Validator/Rules/UniqueArgumentDefinitionNames.php
spawnia: I would have guessed that perhaps the reported line/columns might differ, but apparently not.

spawnia:
@vhenzl Do you plan any more changes that we should include before cutting a v15 release?

vhenzl:
@spawnia Yeah, there are a few things I would like to solve. I'll put together a summary.

PR #1137, opened 2022-05-04
Never resolve additional types in Schema constructor

ruudk:
When you construct a Schema and pass an array of type callables like this:
```php
Schema(['types' => [
  'TypeName' => function() { },
]);
```
the callables are immediately resolved in the constructor.

If you want the types to not be resolved, you must use a callable that returns the types array with callables like this:
```php
Schema(['types' => fn() => [
  'TypeName' => function() { },
]);
```

I was surprised by that because I thought my array of callables was already enough for lazy loading.

When looking at the source code I want the distinction a bit weird and I think it can be handled in the same way.

When `Schema::getTypeMap` is called, it will call `Schema::collectAllTypes` which will call `Schema::resolveAdditionalTypes`.


spawnia:
I am trying top minimize type loading in https://github.com/webonyx/graphql-php/pull/1104

ruudk:
@spawnia Hmm, not really sure how. Maybe you can include this change in that PR when you finish it.

ruudk:
I need to register some types that are not referenced by other types, how is that possible without using the `types` array?

spawnia:
Can you illustrate why you would need many such types? If there is a small number of them, eager loading them should be no problem.

Anyways, because `types` currently does not require to be a map with type names as keys, it is not really possible to load the types within lazily. As soon as the first unknown type needs to be found, we need to traverse and resolve the types within. We could change this, but that would be a big breaking change. Not sure if it is worth it.

PR #1138, opened 2022-05-04
Fix broken SchemaPrinter

ruudk:
In 914d91e717ac2b5e5dc7b4253726b00185ac745b a breaking change was introduced while removing `Utils::every`.

Since that change, it only includes descriptions if ALL of the arguments have a description.

If one argument has a description and one has not, no descriptions will be added.

That doesn't make sense.

It should be the opposite.

When none of the arguments have a description, then inline them.


spawnia:
Please add a test case to https://github.com/webonyx/graphql-php/blob/master/tests/Utils/SchemaPrinterTest.php

ruudk:
@spawnia Done!

tests/Utils/SchemaPrinterTest.php
```diff
@@ -420,6 +420,37 @@ public function testPrintsStringFieldWithMultipleArgsLastIsDefault(): void
...
+        self::assertPrintedSchemaEquals(
+            <<<'GRAPHQL'
+            type Query {
+              singleField(
+                """This is the first argument"""
+                argOne: Int
+
+                """This is the second argument"""
+                argTwo: String
+                argThree: Boolean
```
ruudk, line 445: Personally, I would find it nicer if there was an extra newline after the arg that has a description.
```suggestion
                """This is the second argument"""
                argTwo: String

                argThree: Boolean
```

tests/Utils/SchemaPrinterTest.php
```diff
@@ -420,6 +420,37 @@ public function testPrintsStringFieldWithMultipleArgsLastIsDefault(): void
         );
     }
 
+    /**
+     * @see it('Prints Fields With Description Args')
```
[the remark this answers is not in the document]
ruudk: Fixed

tests/Utils/SchemaPrinterTest.php
spawnia: Me too! Would you like to make that change in this PR? I am fine with diverging from the reference implementation here.

tests/Utils/SchemaPrinterTest.php
```diff
@@ -420,6 +420,34 @@ public function testPrintsStringFieldWithMultipleArgsLastIsDefault(): void
         );
     }
 
+    public function testPrintsFIeldsWithDescriptionArgs(): void
```
spawnia, line 423:
```suggestion
    public function testPrintsArgsWithDescriptionsInMultipleLines(): void
```

tests/Utils/SchemaPrinterTest.php
ruudk: I'll try it in a new PR :)

spawnia:
Nicely done, thank you

tests/Utils/SchemaPrinterTest.php
ruudk: Done in https://github.com/webonyx/graphql-php/pull/1141

PR #1139, opened 2022-05-04
Add support for type extensions to schema build

vhenzl:
Type extensions (e.g. `extend type MyType { newField: String }`) currently works only with the schema extender, they are silently ignored by the schema builder (types are not extended, `extensionASTNodes` is not set). This PR fixes that and adds support for type extensions to schema build.

Similar changes in the reference implementation: https://github.com/graphql/graphql-js/commit/1283c8431b561c529adfdfc0ade845fad2a09ade

Tests by origin from `graphql-js` come from #928 and #929; this PR brings these two a bit closer to closing.

I'll later do a similar change for schema extension.  

(Draft PR for now as I still need to self-review, but should be good in general.)

### DIFF
--- a/src/Utils/ASTDefinitionBuilder.php
+++ b/src/Utils/ASTDefinitionBuilder.php
@@ -146,6 +146,22 @@ class ASTDefinitionBuilder
     }
 
     /**
+     * @param array<InputObjectTypeDefinitionNode|InputObjectTypeExtensionNode> $nodes
+     *
+     * @return array<string, UnnamedInputObjectFieldConfig>
+     */
+    private function makeInputFields(array $nodes): array
+    {
+        /** @var array<int, InputValueDefinitionNode>> $fields */
+        $fields = [];
+        foreach ($nodes as $node) {
+            \array_push($fields, ...$node->fields);
+        }
+
+        return $this->makeInputValues(new NodeList($fields));
+    }
+
+    /**
      * @param ListTypeNode|NonNullTypeNode|NamedTypeNode $typeNode
      */
     private function buildWrappedType(TypeNode $typeNode): Type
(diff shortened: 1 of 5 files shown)


src/Utils/ASTDefinitionBuilder.php
```diff
@@ -256,25 +285,32 @@ private function makeSchemaDef(Node $def): Type
...
+        $extensionASTNodes = $this->typeExtensionsMap[$name] ?? [];
+
         return new ObjectType([
-            'name' => $def->name->value,
+            'name' => $name,
             'description' => $def->description->value ?? null,
-            'fields' => fn (): array => $this->makeFieldDefMap($def),
-            'interfaces' => fn (): array => $this->makeImplementedInterfaces($def),
+            'fields' => fn (): array => $this->makeFieldDefMap([$def, ...$extensionASTNodes]),
+            'interfaces' => fn (): array => $this->makeImplementedInterfaces([$def, ...$extensionASTNodes]),
```
spawnia, line 296: `[$def, ...$extensionASTNodes]` can be extracted to a variable.

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -338,72 +376,101 @@ private function makeImplementedInterfaces($def): array
...
+        $extensionASTNodes = $this->typeExtensionsMap[$name] ?? [];
+
         return new InterfaceType([
-            'name' => $def->name->value,
+            'name' => $name,
             'description' => $def->description->value ?? null,
-            'fields' => fn (): array => $this->makeFieldDefMap($def),
-            'interfaces' => fn (): array => $this->makeImplementedInterfaces($def),
+            'fields' => fn (): array => $this->makeFieldDefMap([$def, ...$extensionASTNodes]),
+            'interfaces' => fn (): array => $this->makeImplementedInterfaces([$def, ...$extensionASTNodes]),
```
spawnia, line 387: Extract `[$def, ...$extensionASTNodes]`

src/Utils/BuildSchema.php
```diff
@@ -144,7 +145,12 @@ public function buildSchema(): Schema
                     $schemaDef = $definition;
                     break;
                 case $definition instanceof TypeDefinitionNode:
-                    $this->nodeMap[$definition->name->value] = $definition;
+                    $typeDefinitionsMap[$definition->name->value] = $definition;
+                    break;
+                case $definition instanceof TypeExtensionNode:
+                    $extendedTypeName = $definition->name->value;
+                    $existingTypeExtensions = $typeExtensionsMap[$extendedTypeName] ?? [];
+                    $typeExtensionsMap[$extendedTypeName] = [...$existingTypeExtensions, $definition];
```
spawnia, line 153:
```suggestion
                    $typeExtensionsMap[$$definition->name->value][] = $definition;
```

tests/Utils/BuildSchemaTest.php
```diff
@@ -713,6 +732,222 @@ enum: MyEnum
         self::assertEquals($rootFields['field2']->deprecationReason, 'Because I said so');
     }
 
+    /**
+     * it('Correctly extend scalar type').
+     */
+    public function testCorrectlyExtendScalarType(): void
```
spawnia, line 738: This does not seem to match the latest test suite from graphql-js, did you match an older version?

tests/Utils/BuildSchemaTest.php
```diff
@@ -713,6 +732,222 @@ enum: MyEnum
...
+            GRAPHQL;
+
+        $schema = BuildSchema::build("
+            ${scalarSDL}
+            directive @foo on SCALAR
+            directive @bar on SCALAR
+        ");
+
+        /** @var ScalarType $someScalar */
+        $someScalar = $schema->getType('SomeScalar');
```
spawnia, line 756:
```suggestion
        $someScalar = $schema->getType('SomeScalar');
        assert($someScalar instanceof ScalarType);
```

https://github.com/webonyx/graphql-php/blob/master/CONTRIBUTING.md#type-assertions

tests/Utils/BuildSchemaTest.php
vhenzl: https://github.com/webonyx/graphql-php/pull/928 and https://github.com/webonyx/graphql-php/pull/929 match graphql-js v15.x as v16 were still in making when I worked on them. I'll check what has changed.

vhenzl: The relevant change that makes tests here and in #928 differ from the current graphql-js is https://github.com/graphql/graphql-js/commit/33c069ce64ebad1379d8c6a7937f98952ad40e63

I don't see it as something relevant for this PR. It can be solved across the whole `BuildSchemaTest` afterwards in #928.

PR #1142, opened 2022-05-05
Replace ${..} string interpolation by {$...}

simPod:
https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation


spawnia:
Good catch. I hope that interpolating constants and function calls will become possible too at some point.

PR #1145, opened 2022-05-12
Add newlines around fields with description

ruudk:
Same as #1141 but now for fields.


src/Utils/SchemaPrinter.php
```diff
@@ -344,7 +344,12 @@ protected static function printFields(array $options, $type): string
     {
         $fields = [];
         $firstInBlock = true;
+        $previousHasDescription = false;
         foreach ($type->getFields() as $f) {
+            if ($previousHasDescription && $f->description === null) {
```
spawnia, line 349: Let's extract `$hasDescription = $f->description !== null;`

spawnia (changes requested):

PR #1147, opened 2022-05-23
Allow converting an input object array to a value object using `parseValue`

### LINKED ISSUE #1146: Allow input object types values to be converted to value objects
When working with mutations you often want to use an input object type. Currently, the provided input values are passed as an array to the resolver. Working with arrays is tedious and error prone. I'd rather have real / typed objects to work with.

It would be great if I could add a `parseValue` method to my `InputObjectType` configuration so that it can convert to a real object instead.

Something like this:
```php
class Tag {
    public function __construct(
        public readonly string $name,
        public readonly string $value,
    ) {}
}
$tag = new InputObjectType([
    'name' => 'Tag',
    'fields' => [
        'name' => [
            'type' => Type::string(),
        ],
        'value' => [
            'type' => Type::string(),
        ],
    ],
    'parseValue' => function(array $values) {
        return new Tag(
            $values['name'],
            $values['value'],
        );
    }
])

class StoryFiltersInput {
    public function __construct(
        public readonly string $author,
        public readonly boolean $popular,
        public readonly Tag $tag,
    ) {}
}
$input = new InputObjectType([
    'name' => 'StoryFiltersInput',
    'fields' => [
        'author' => [
            'type' => Type::id(),
        ],
        'popular' => [
            'type' => Type::boolean(),
        ],
        'tag' => [
            'type' => $tag,
        ]
    ],
    'parseValue' => function(array $values) {
        return new StoryFiltersInput(
            $values['author'],
            $values['popular'],
            $values['tag'],
        );
    }
])
```

As you can see, by the time the `StoryFiltersInput` is constructed, the `$values['tag']` is already parsed to an instance of `Tag`.

I checked to see if I could add this, and I think it should happen somewhere here:
https://github.com/webonyx/graphql-php/blob/cc1ce0572461dec9bb75bbc48cb7eb0049ae3b2e/src/Utils/Value.php#L48

@spawnia Do you think this is something that can be added? I can try to create a PR and see how far I come.

ruudk:
Fixes #1146

### DIFF
--- a/docs/type-definitions/inputs.md
+++ b/docs/type-definitions/inputs.md
@@ -35,13 +35,14 @@ Every field may be of other InputObjectType (thus complex hierarchies of inputs 
 
 ## Configuration options
 
-The constructor of InputObjectType accepts an array with only 3 options:
+The constructor of `InputObjectType` accepts an `array` with the following options:
 
-| Option      | Type                  | Notes                                                                                                                                           |
-| ----------- | --------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
-| name        | `string`              | **Required.** Unique name of this object type within Schema                                                                                     |
-| fields      | `array` or `callable` | **Required**. An array describing object fields or callable returning such an array (see below).                                                |
-| description | `string`              | Plain-text description of this type for clients (e.g. used by [GraphiQL](https://github.com/graphql/graphiql) for auto-generated documentation) |
+| Option      | Type                                    | Notes                                                                                                                                           |
+| ----------- | --------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
+| name        | `string`                                | **Required.** Unique name of this object type within Schema                                                                                     |
+| fields      | `array` or `callable`                   | **Required**. An array describing object fields or callable returning such an array (see below).                                                |
+| description | `string`                                | Plain-text description of this type for clients (e.g. used by [GraphiQL](https://github.com/graphql/graphiql) for auto-generated documentation) |
+| parseValue  | `callable(array<string, mixed>): mixed` | Converts incoming values from their array representation to something else (e.g. a value object)                                                |
 
 Every field is an array with following entries:
 
@@ -108,3 +109,51 @@ $variables = [
 
 **graphql-php** will validate the input against your InputObjectType definition and pass it to your
 resolver as `$args['filters']`
+
+## Converting input object array to value object
+
+If you want more type safety you can choose to parse the input array into a value object.
+
+```php
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Definition\InputObjectType;
+
+final class StoryFiltersInput
+{
+    public string $author;
+    public ?bool $popular;
+    public array $tags;
+
+    public function __construct(string $author, ?bool $popular, array $tags)
+    {
+        $this->author = $author;
+        $this->popular = $popular;
+        $this->tag = $tag;
+    }
+}
+
+$filters = new InputObjectType([
+    'name' => 'StoryFiltersInput',
+    'fields' => [
+        'author' => [
+            'type' => Type::nonNull(Type::string()),
+        ],
+        'popular' => [
+            'type' => Type::boolean(),
+        ],
+        'tags' => [
+            'type' => Type::nonNull(Type::listOf(Type::string())),
+        ]
+    ],
+    'parseValue' => fn(array $values) => new StoryFiltersInput(
+        $values['author'],
+        $values['popular'] ?? null,
+        $values['tags']
+    ),
+]);
+```
+
+The value of `$args['filters']` will now be an instance of `StoryFiltersInput`.
+
+The incoming values are converted using a depth-first traversal. Thus, nested input values
+will be passed through their respective `parseValue` functions before the parent receives their value.
--- a/src/Type/Definition/InputObjectType.php
+++ b/src/Type/Definition/InputObjectType.php
@@ -20,6 +20,7 @@ use function is_string;
  *   name?: string|null,
  *   description?: string|null,
  *   fields: iterable<FieldConfig>|callable(): iterable<FieldConfig>,
+ *   parseValue?: callable(array<string, mixed>): mixed,
  *   astNode?: InputObjectTypeDefinitionNode|null,
  *   extensionASTNodes?: array<int, InputObjectTypeExtensionNode>|null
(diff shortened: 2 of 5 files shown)


ruudk:
@spawnia It was actually pretty easy to do this. It's ready for review.

src/Type/Definition/InputObjectType.php
```diff
@@ -142,6 +143,24 @@ protected function initializeField($nameOrIndex, $field): void
         $this->fields[$field->name] = $field;
     }
 
+    /**
+     * Parses an externally provided value (query variable) to use as an input.
+     *
+     * Should throw an exception with a client friendly message on invalid values, @see ClientAware.
+     *
+     * @param mixed $value
```
spawnia, line 151:
```suggestion
     * @param array<string, mixed> $value
```

I don't think it can ever be anything else.

spawnia (changes requested):

src/Type/Definition/InputObjectType.php
ruudk: You're right! Changed it.

ruudk:
Updated the PR, ready 🚀

src/Type/Definition/InputObjectType.php
```diff
@@ -20,6 +20,7 @@
  *   name?: string|null,
  *   description?: string|null,
  *   fields: iterable<FieldConfig>|callable(): iterable<FieldConfig>,
+ *   parseValue?: callable(mixed): mixed,
```
spawnia, line 23:
```suggestion
 *   parseValue?: callable(array<string, mixed>): mixed,
```

ruudk:
Like this?

tests/Type/InputObjectTypeTest.php
```diff
@@ -0,0 +1,132 @@
...
+
+use GraphQL\Examples\Blog\Types;
+use GraphQL\GraphQL;
+use GraphQL\Type\Definition\InputObjectType;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Schema;
+use PHPUnit\Framework\TestCase;
+
+class Tag
```
spawnia, line 13: https://github.com/webonyx/graphql-php/blob/master/CONTRIBUTING.md#final-classes

tests/Type/InputObjectTypeTest.php
```diff
@@ -0,0 +1,132 @@
...
+    {
+        $this->author = $author;
+        $this->popular = $popular;
+        $this->tag = $tag;
+    }
+}
+
+class InputObjectTypeTest extends TestCase
+{
+    public function testParseValue(): void
```
spawnia, line 44: Let's add a test that ensures `parseValue` is not called when explicitly passing `null` as the value for an input object.

spawnia (changes requested):

docs/type-definitions/inputs.md
```diff
@@ -108,3 +108,47 @@ $variables = [
 
 **graphql-php** will validate the input against your InputObjectType definition and pass it to your
 resolver as `$args['filters']`
+
+## Converting input object array to value object
+
+If you want more type safety you can choose to parse the input array into a value object.
+
+```php
```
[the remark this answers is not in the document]
ruudk: Better?

ruudk:
Applied your feedback.

docs/type-definitions/inputs.md
```diff
@@ -35,13 +35,14 @@ Every field may be of other InputObjectType (thus complex hierarchies of inputs
 
 ## Configuration options
 
-The constructor of InputObjectType accepts an array with only 3 options:
+The constructor of InputObjectType accepts an array with 4 options:
```
spawnia, line 38:
```suggestion
The constructor of `InputObjectType` accepts an `array` with the following options:
```

docs/type-definitions/inputs.md
```diff
@@ -35,13 +35,14 @@ Every field may be of other InputObjectType (thus complex hierarchies of inputs
...
+| ----------- | --------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
+| name        | `string`                                | **Required.** Unique name of this object type within Schema                                                                                     |
+| fields      | `array` or `callable`                   | **Required**. An array describing object fields or callable returning such an array (see below).                                                |
+| description | `string`                                | Plain-text description of this type for clients (e.g. used by [GraphiQL](https://github.com/graphql/graphiql) for auto-generated documentation) |
+| parseValue  | `callable(array<string, mixed>): mixed` | Callback to convert the array to something else (e.g. a value object)                                                                           |
```
spawnia, line 45:
```suggestion
| parseValue  | `callable(array<string, mixed>): mixed` | Converts incoming values from their array representation to something else (e.g. a value object)                                                                           |
```

docs/type-definitions/inputs.md
```diff
@@ -108,3 +109,48 @@ $variables = [
...
+        'popular' => [
+            'type' => Type::boolean(),
+        ],
+        'tags' => [
+            'type' => Type::listOf(Type::string()),
+        ]
+    ],
+    'parseValue' => fn(array $values) => new StoryFiltersInput(
+        $values['author'],
+        $values['popular'],
```
spawnia, line 150:
```suggestion
        $values['popular'] ?? null,
```

docs/type-definitions/inputs.md
```diff
@@ -108,3 +109,48 @@ $variables = [
...
+        $values['tags'] ?? []
```
spawnia, line 151: Defaulting to an empty array seems quite opinionated, especially given that the argument could have an explicit default value of `[]` transparent to clients. Why shouldn't it default to `null`? Perhaps we can just make the type of tags non-nullable and avoid deciding.

docs/type-definitions/inputs.md
```diff
@@ -108,3 +109,48 @@ $variables = [
...
+    ),
+]);
+```
+
+The value of `$args['filters']` will now be an instance of `StoryFiltersInput`.
```
spawnia, line 156: Perhaps we can mention how nested inputs are handled. How about the following?

The incoming values are converted using a depth-first traversal. Thus, nested input values
will be passed through their respective `parseValue` functions before the parent receives their value.

ruudk:
Can it be merged like this? Would love to start using it 🤩

spawnia (approved):
Nice work!

Issue #1149, opened 2022-05-23
Introduce `DeferredType` that allows linking types together without having to pass a reference to the actual type object or use callbacks


ruudk:
When specifying a field type, I have to either provide the actual object instance, or use a callback that delegates the call to a registry to fetch the single instance of that type.

```php
$filters = new InputObjectType([
    'name' => 'StoryFiltersInput',
    'fields' => [
        'tags' => [
            'type' => fn() => Type::nonNull( Type::listOf( $registry->get(SomeOtherType::class) ) ),
        ]
    ]
]);
```

I wonder why it works like that. Maybe this is for legacy reasons. But I would find it way easier if I could just link types by a FQCN or GraphQL type name.

So what if we create something like this:
```php
final class DeferredType extends GraphQLType
{
    public function __construct(public readonly string $name)
    {
    }

    public function toString() : string
    {
        return $this->name;
    }
}
```

```php
$filters = new InputObjectType([
    'name' => 'StoryFiltersInput',
    'fields' => [
        'tags' => [
            'type' => Type::nonNull( Type::listOf( new DeferredType(SomeOtherType::class) ) ),
        ]
    ]
]);
```

Now the type can be added to the schema. As soon as the type is needed, it needs to resolve the DeferredType to the actual type. 
But that only happens when it is needed.

From an API perspective, it feels so much nicer to define things like this than to go with the callbacks + registry approach.

spawnia:
How about `TypeReference` over `DeferredType`?

> As soon as the type is needed, it needs to resolve the DeferredType to the actual type.

That would require including a type loader with the schema, not every schema has one. I guess we could throw when encountering a type reference but the schema has no type loader.

ruudk:
> How about `TypeReference` over [DeferredType](url)?

I like it!

> That would require including a type loader with the schema, not every schema has one. I guess we could throw when encountering a type reference but the schema has no type loader.

Sounds good

ruudk:
Updated the PR to reflect `TypeReference`.

ruudk:
That's true, but why should it be fully possible to work on its own? Isn't that design decision (made long time) one of the reasons why we now have all these callbacks to make things lazy again? 

I hope that for a future version of this _wonderful library_, a rewrite will undo all of this, and let types be simple and stupid with the Schema connecting everything together.

spawnia:
[...]

PR #1151, opened 2022-05-24
Introduce `TypeReference`

ruudk:
See #1149

### DIFF
--- a/src/Type/Definition/TypeReference.php
+++ b/src/Type/Definition/TypeReference.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Type\Definition;
+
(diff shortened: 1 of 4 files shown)


src/Type/Definition/TypeReference.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace GraphQL\Type\Definition;
+
+class TypeReference extends Type implements NullableType, NamedType
```
spawnia, line 7: I don't see why this has to extend `Type`, it can not be used as such.

spawnia:
Just merged https://github.com/webonyx/graphql-php/pull/1104 which changes `Schema` significantly.

PR #1153, opened 2022-05-25
Utils > Replace Schema with type loader callable in `AST::typeFromAST()`

ruudk:
This allows it to be used independently from the Schema. One use case could be to parse a GraphQL type string and resolve it the type via a container / registry.

@spawnia Acceptable as breaking change or do we want to keep the old one, and introduce a new one?


src/Utils/AST.php
```diff
@@ -540,24 +539,25 @@ public static function valueFromASTUntyped(Node $valueNode, ?array $variables =
...
      *
+     * @param callable(string): ?Type $loader
      * @param NamedTypeNode|ListTypeNode|NonNullTypeNode $inputTypeNode
      *
      * @throws Exception
      *
      * @api
      */
-    public static function typeFromAST(Schema $schema, Node $inputTypeNode): ?Type
+    public static function typeFromAST(callable $loader, Node $inputTypeNode): ?Type
```
spawnia, line 549:
```suggestion
    public static function typeFromAST(callable $typeLoader, Node $inputTypeNode): ?Type
```

src/Utils/TypeInfo.php
```diff
@@ -349,7 +349,7 @@ private static function getFieldDefinition(Schema $schema, Type $parentType, Fie
      */
     public static function typeFromAST(Schema $schema, Node $inputTypeNode): ?Type
```
spawnia, line 350:
```suggestion
    public static function typeFromAST(callable $typeLoader, Node $inputTypeNode): ?Type
```

Let's bubble up that change even further and make it throughout.

spawnia (commented):
I like the idea, makes the utility function more generally useful.

ruudk:
Updated the PR, I removed `GraphQL\Utils\TypeInfo::typeFromAST()` because you can use `GraphQL\Utils\AST::typeFromAST()` directly.

What do you think?

src/Utils/AST.php
```diff
@@ -540,24 +539,25 @@ public static function valueFromASTUntyped(Node $valueNode, ?array $variables =
     /**
      * Returns type definition for given AST Type node.
      *
+     * @param callable(string): ?Type                    $typeLoader
```
spawnia, line 542:
```suggestion
     * @param callable(string): ?Type $typeLoader
```

spawnia:
I like it, removes unnecessary indirection.

PR #1155, opened 2022-05-31
Allow retrieving directive values from type definitions

bojanbogdanovic:
When using `Values::getDirectiveValues()` with a  type definition `$info->returnType->astNode`; PHPStan was giving the following error:

`Parameter #2 $node of static method GraphQL\Executor\Values::getDirectiveValues() expects GraphQL\Language\AST\EnumValueDefinitionNode|GraphQL\Language\AST\FieldDefinitionNode|GraphQL\Language\AST\FieldNode|GraphQL\Language\AST\FragmentSpreadNode|GraphQL\Language\AST\InlineFragmentNode, GraphQL\Language\AST\TypeDefinitionNode given.`

It should be possible to retrieve directive values from type definitions, e.g: `type Test @testDirective(testValue: "test")`

### DIFF
--- a/src/Executor/Values.php
+++ b/src/Executor/Values.php
@@ -15,6 +15,7 @@ use GraphQL\Language\AST\InlineFragmentNode;
 use GraphQL\Language\AST\Node;
 use GraphQL\Language\AST\NodeList;
 use GraphQL\Language\AST\NullValueNode;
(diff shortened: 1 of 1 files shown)


spawnia:
Thanks. Due to a lack of an interface that defines the property `$directives`, I simply enumerated all options.

PR #1167, opened 2022-06-09
Do not change HTTP status code in `StandardServer`

spawnia:
Resolves https://github.com/webonyx/graphql-php/issues/309

This change allows users to set the status code they want themselves, usually defaulting to 200.

### DIFF
--- a/src/Server/Helper.php
+++ b/src/Server/Helper.php
@@ -442,42 +442,6 @@ class Helper
     }
 
     /**
-     * @param ExecutionResult|array<ExecutionResult> $result
-     */
-    protected function resolveHttpStatus($result): int
-    {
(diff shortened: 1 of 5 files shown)


src/Server/Helper.php
```diff
@@ -415,30 +415,20 @@ public function sendResponse($result): void
...
+            $this->emitResponse($result);
         }
     }
 
-    /**
-     * @param ExecutionResult|array<ExecutionResult> $result
-     */
-    protected function doSendResponse($result): void
-    {
-        $httpStatus = $this->resolveHttpStatus($result);
```
simPod, line 430: is this method still used now?

simPod (approved):

PR #1169, opened 2022-06-11
14.x

Chrisdowson:
1.add graphql with reactphp


spawnia:
Thanks for this example. I am generally open to adding something like it to illustrate usage with react-php, but a few things need to be different.

The example is way too large and mostly just a copy of the blog example. This is bad for continued maintenance, and I also think it is unnecessarily complex. Please trim it down to a minimal example that illustrates relevant bits of code - preferrably all in a single file.

Also, please target `master` with your PR. `14.x` is on maintenance mode only.

PR #1171, opened 2022-06-12
Avoid calling `FormattedError::addDebugEntries()` twice when using default error formatting

Warxcell:
Its already added inside `createFromException`

https://github.com/webonyx/graphql-php/blob/master/src/Error/FormattedError.php#L182-L184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Clone `NodeList` in `Node::cloneDeep()`
 - Calling `Schema::getType()` on a schema built from SDL returns `null` for unknown types (#1068)
 - Avoid crash on typeless inline fragment when using `QueryComplexity` rule
+- Avoid calling `FormattedError::addDebugEntries()` twice when using default error formatting
 
 ### Removed
 
--- a/src/Error/FormattedError.php
+++ b/src/Error/FormattedError.php
@@ -245,7 +245,12 @@ class FormattedError
      */
     public static function prepareFormatter(?callable $formatter, int $debug): callable
     {
-        $formatter ??= [self::class, 'createFromException'];
+        // Prevents double work of adding debug entries
+        if(null !== $formatter && $debug !== DebugFlag::NONE) {
+            $formatter = static fn (Throwable $e): array => self::addDebugEntries($formatter($e), $e, $debug);
+        } else {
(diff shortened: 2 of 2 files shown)


mfn:
> Its already added inside createFromException

Whilst this is true, the change is not correct: `createFromException` is called as a default value. When you provide your own `$formatter`, the debug flag check is still relevant.

Warxcell:
Isn't responsibility of those who provides `$formatter` to handle debug flags on his own? Like some other things in this library.

spawnia:
I don't know what problem this is supposed to solve. Please provide a test case that shows how the current behaviour is problematic.

Warxcell:
> I don't know what problem this is supposed to solve. Please provide a test case that shows how the current behaviour is problematic.

 its not really bug or problem - just some code executed twice - what's the point of it?

spawnia:
Apparently there is a point - see the failing test cases in https://github.com/webonyx/graphql-php/runs/6857237019?check_suite_focus=true

Warxcell:
> Apparently there is a point - see the failing test cases in https://github.com/webonyx/graphql-php/runs/6857237019?check_suite_focus=true

Maybe tests are wrong? They probably re-define formatter, without adding appropriate debug info inside overriden formatter.

If that's the case - I will remove duplicated code inside formatter, not the decorated one, but I think that if you take over the formatter - its your responsibility to format and errors aswell. 

Please advice.

spawnia:
The failing tests are not even particular to error handling, they mostly use the default formatter and leverage the debug flag.

After some review, I found that the current approach is documented: https://webonyx.github.io/graphql-php/error-handling/#custom-error-handling-and-formatting

I can see that adding debug information twice performs unnecessary work, and I could also see how it could be useful to customize how debug information is added. We would have to break the API to change this though, and I am not sure it is worth it to do so.

Warxcell:
> The failing tests are not even particular to error handling, they mostly use the default formatter and leverage the debug flag.
> 
> After some review, I found that the current approach is documented: https://webonyx.github.io/graphql-php/error-handling/#custom-error-handling-and-formatting
> 
> I can see that adding debug information twice performs unnecessary work, and I could also see how it could be useful to customize how debug information is added. We would have to break the API to change this though, and I am not sure it is worth it to do so.

Ok, added check in prepareFormatter - if formatter is provided - decorate it - otherwise  no, because debug entries already exists in createFromException.

PR #1180, opened 2022-06-17
Ensure `InputObjectType::parseValue()` works with literals and extended schemas

### LINKED ISSUE #1178: InputObjectType parse value not working
Its documented under: https://webonyx.github.io/graphql-php/type-definitions/inputs/#converting-input-object-array-to-value-object

But it simply arrives as array.

I was able to follow it down to https://github.com/webonyx/graphql-php/blob/master/src/Utils/AST.php#L392-L435

its simply not called at all.

Warxcell:
Fixes #1178

### DIFF
--- a/src/Utils/AST.php
+++ b/src/Utils/AST.php
@@ -432,7 +432,7 @@ class AST
                 $coercedObj[$fieldName] = $fieldValue;
             }
 
-            return $coercedObj;
+            return $type->parseValue($coercedObj);
         }
 
         if ($type instanceof EnumType) {
--- a/src/Utils/SchemaExtender.php
+++ b/src/Utils/SchemaExtender.php
@@ -253,7 +253,7 @@ class SchemaExtender
             'fields' => fn (): array => $this->extendInputFieldMap($type),
             'astNode' => $type->astNode,
             'extensionASTNodes' => $extensionASTNodes,
-            'parseValue' => $type->config['parseValue'] ?? null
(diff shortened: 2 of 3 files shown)


Warxcell:
@spawnia saw the actual bug - when extended parseValue is not transferred into new InputObjectType. Fixed it.

spawnia:
@Warxcell great job on the test and the fix, thank you.

I am not quite sure if `parseValue` would work when providing the values as GraphQL literals (not using variables). Could you add a test for that too and make sure it works?

Warxcell:
Yep, seems its not working.

spawnia:
I think the place where this conversion was previously added was fundamentally wrong, because variable values are kept separate from value literals after that. A client might send a query like this:

```graphql
query ($foo: Int) {
  foo(input: { foo: $foo, bar: 2 })
}
```

In that case, both sources of values will have to be combined - only then can `InputObjectType::parseValue()` be called. I think the stacktrace you did in https://github.com/webonyx/graphql-php/issues/1178#issuecomment-1158819536 pretty much points to the correct place to do this.

Warxcell:
Seems its needed on both places. That way probably it won't work when using mixed (raw variables and literals). 

If I put it inside `Values.php` - i should traverse from deeper level in order for upper levels to receive correct types.

spawnia:
Thanks @Warxcell for your work on this.

PR #1184, opened 2022-06-23
Add `TypeInfo::getParentTypeStack()`

mostafa-rz:
Hi,
The TypeInfo can be used in Visitor and in my use case I need the path to the current node. The information is already on private property there and I only need to get it.


CHANGELOG.md
```diff
@@ -65,6 +65,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Add `parseValue` config option to InputObjectType to parse input value to custom value object
 - Add option `sortTypes` to have `SchemaPrinter` order types alphabetically
 - Allow constructing `EnumType` from PHP enum
+- Add `getParentTypeStack()` method to `TypeInfo.php`
```
spawnia, line 68:
```suggestion
- Add `TypeInfo::getParentTypeStack()`
```

PR #1188, opened 2022-06-28
Use `"` instead of `"""` for single line descriptions

simPod:
I've checked the spec and it says to use single double quotes to quote single line descriptions.

See https://spec.graphql.org/October2021/#sec-Descriptions

### DIFF
--- a/src/Language/BlockString.php
+++ b/src/Language/BlockString.php
@@ -139,10 +139,13 @@ class BlockString
             : $value;
         if ($printAsMultipleLines) {
             $result .= "\n";
+            $quoting = '"""';
+        } else {
+            $quoting = '"';
         }
 
-        return '"""'
-            . str_replace('"""', '\\"""', $result)
-            . '"""';
+        return $quoting
+            . str_replace($quoting, '\\' . $quoting, $result)
+            . $quoting;
     }
 }
--- a/tests/Language/SchemaPrinterTest.php
+++ b/tests/Language/SchemaPrinterTest.php
@@ -63,9 +63,13 @@ of the `Foo` type.
 """
 type Foo implements Bar & Baz & Two {
   one: Type
-  """This is a description of the `two` field."""
+  """
+  This is a description of the `two` field.
+  """
   two(
-    """This is a description of the `argument` argument."""
+    """
+    This is a description of the `argument` argument.
+    """
     argument: InputType!
   ): Type
   three(argument: InputType, other: String): Int
--- a/src/Language/Printer.php
+++ b/src/Language/Printer.php
@@ -410,7 +410,7 @@ class Printer
 
             case $node instanceof StringValueNode:
                 if ($node->block) {
-                    return BlockString::print($node->value, $isDescription ? '' : '  ');
+                    return BlockString::print($node->value, $isDescription ? '' : '  ', true);
                 }
 
                 return json_encode($node->value, JSON_THROW_ON_ERROR);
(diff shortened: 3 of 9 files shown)


tests/Language/SchemaPrinterTest.php
```diff
@@ -63,9 +63,13 @@ public function testPrintsKitchenSink(): void
 """
 type Foo implements Bar & Baz & Two {
   one: Type
-  """This is a description of the `two` field."""
+  """
+  This is a description of the `two` field.
+  """
```
spawnia, line 68: Why are those not one-liners with `"`?

src/Language/BlockString.php
```diff
@@ -139,10 +139,13 @@ public static function print(
...
+            $quoting = '"""';
+        } else {
+            $quoting = '"';
         }
 
-        return '"""'
-            . str_replace('"""', '\\"""', $result)
-            . '"""';
+        return $quoting
+            . str_replace($quoting, '\\' . $quoting, $result)
```
spawnia, line 148: Is this tested somewhere?

tests/Language/SchemaPrinterTest.php
simPod: I tried to change as little as possible. Here it was not a one-liner in `schema-kitchen-sink.graphql` fixture.

simPod:
I see ref impl has `"""` for one liners but it's against the spec as well. Will try to raise an issue there.

tests/Language/SchemaPrinterTest.php
spawnia: Got it. How would a description like the following be transformed:

```
"""one liner in multi quotes"""
type Foo
```

I guess it would not stay the same, but rather switch to either a one liner with single quotes or a block with triple quotes.

simPod: Yes, there are two options according to spec

```
"one liner in multi quotes"
type Foo
```

```
"""
one liner in multi quotes
"""
type Foo
```

spawnia: I think we should go with option one and add a test to ensure consistent behaviour.

simPod: What do you mean 🤔 Both are possible, depends on the user what he defines.

I can adapt our tests (fixtures) to always use single quotes.

When you set `description` with single line value with no trailing quote or slash, `"` is used.

spawnia: Please re-read https://github.com/webonyx/graphql-php/pull/1188#discussion_r908691474.

The example would be a user-defined schema. I am assuming that during parsing some information will be lost, so when the schema is then printed it will not output exactly the same.

simPod: Ah now I get the initial question.

```gql
"""one liner in multi quotes"""
type Foo
```
is parsed as a BlockString token since our parser cares about triple quotes.

spawnia: Fine, that would mean the printed output would be option 2. That is fine with me, but we should cover this behaviour with a test.

simPod: So schema printer test now covers it

https://github.com/webonyx/graphql-php/pull/1188/files#diff-61fa379df85242a4f4a4f5029dc7fde8f6d8bd0c3989e50124aa9421de9644d5R17

Issue #1193, opened 2022-07-05
Naming a Type like a callable function in scope calls the existing function even if its unintended


oneserv-heuser:
I have a Type called 'Link'. When resolving the schema with webonyx/graphql-php since 14.0.0 I get the following error:
```
.ERROR: link() expects exactly 2 parameters, 0 given
```
This happens because of the following lines, especially line 1028:

https://github.com/webonyx/graphql-php/blob/d0b7c5270e155c9d7369c5d35dd4731d98e8ba52/src/Executor/ReferenceExecutor.php#L1026-L1032

The [is_callable doc](https://www.php.net/manual/en/function.is-callable.php) says that it returns whether a value can be called as a function from the current scope. As the function is used in line 1028 any Type with a name same as any function callable from the ReferenceExecutor class scope will be handled as callable instead of a Type.

That leads to the execution of Schema::resolveType() and the passed argument is now "Link". Now the php "link" function is called and leads to the error as no parameters are given.
https://github.com/webonyx/graphql-php/blob/d0b7c5270e155c9d7369c5d35dd4731d98e8ba52/src/Type/Schema.php#L318-L325

This works with any function callable from the ReferenceExecutor class scope.
An easy fix I tested for my use case would be to add a !is_string check to [line 1028 from the ReferenceExecutor.](https://github.com/webonyx/graphql-php/blob/d0b7c5270e155c9d7369c5d35dd4731d98e8ba52/src/Executor/ReferenceExecutor.php#L1028)
```PHP
if ($typeCandidate === null) {
    $runtimeType = static::defaultTypeResolver($result, $exeContext->contextValue, $info, $returnType);
/** added !is_string check**/
} elseif (is_callable($typeCandidate) && !is_string($typeCandidate)) {
    $runtimeType = $typeCandidate();
} else {
    $runtimeType = $typeCandidate;
}
```

spawnia:
Thanks for the report. Please try and add a pull request with a failing test, probably in https://github.com/webonyx/graphql-php/blob/master/tests/Type/LazyTypeLoaderTest.php or https://github.com/webonyx/graphql-php/blob/master/tests/Executor/ExecutorLazySchemaTest.php.

oneserv-heuser:
@spawnia Added a test, as requested in #1194

PR #1194, opened 2022-07-06
Avoid calling defined functions named like lazily loaded types

oneserv-heuser:
As requested, this adds a failing test for #1193

Proposed solution:
Add a !is_string check to [line 1028 from the ReferenceExecutor.](https://github.com/webonyx/graphql-php/blob/d0b7c5270e155c9d7369c5d35dd4731d98e8ba52/src/Executor/ReferenceExecutor.php#L1028)
```PHP
if ($typeCandidate === null) {
    $runtimeType = static::defaultTypeResolver($result, $exeContext->contextValue, $info, $returnType);
/** added !is_string check**/
} elseif (is_callable($typeCandidate) && !is_string($typeCandidate)) {
    $runtimeType = $typeCandidate();
} else {
    $runtimeType = $typeCandidate;
}
```

Resolves https://github.com/webonyx/graphql-php/issues/1193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Calling `Schema::getType()` on a schema built from SDL returns `null` for unknown types (#1068)
 - Avoid crash on typeless inline fragment when using `QueryComplexity` rule
 - Avoid calling `FormattedError::addDebugEntries()` twice when using default error formatting
+- Avoid calling defined functions named like lazily loaded types
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


spawnia:
Cool, thanks for providing the test. I modified it a bit to ensure the cause of failure is actually what was intended - see https://github.com/webonyx/graphql-php/runs/7210529657?check_suite_focus=true. Can you try and add the fix now?

oneserv-heuser:
@spawnia Thanks for clarifying my code, looks way better now.
Added the fix, locally all tests are running successfully.

spawnia:
Thanks for the fix!

PR #1195, opened 2022-07-07
Coercion error path.

Warxcell:
I know that's not best place to put it, I'm looking for more elegant solution.

The idea is to have path of error in structured way, so we can format it to the frontend - and frontend can locate the error easier - otherwise frontend should parse strings to locate the error.


spawnia:
I do not understand what this is trying to achieve. Please add a test that illustrates the effect of this change.

Warxcell:
The idea is client of api to avoid parsing strings from error message like that
image

But directly receive path as array.

spawnia:
We will implement nothing that breaks or extends the error format specified in the GraphQL specification, see https://spec.graphql.org/draft#sec-Errors. I think what you are trying to do will require an addition to `extensions`.

src/Utils/Value.php
```diff
@@ -256,12 +259,22 @@ private static function coercionError(
...
+            } while ($path !== null);
+        }
+
         return new Error(
             $fullMessage,
             $blameNode,
             null,
             [],
-            null,
+            $atPath,
```
spawnia, line 277: This is not what `$path` is intended for, it is meant to represent the path of fields leading up to the error - not arguments! https://github.com/webonyx/graphql-php/blob/6f4e702c810f2902e43671701fb5a3e5a48bb68f/src/Error/Error.php#L40-L46

spawnia:
> Otherwise I should preg_match from error message

That might be the easiest. Alternatively, I could see wrapping the `$originalError` with an exception such as `CoercionError` and expose structured data there - or perhaps have `CoercionError` extend `Error`.

Warxcell:
> 

Sounds perfect to me.

spawnia:
Let's go with `class CoercionError extends Error` and reduce the number of nested exceptions.

src/Error/CoercionError.php
```diff
@@ -0,0 +1,40 @@
+<?php declare(strict_types=1);
+
+namespace GraphQL\Error;
+
+use GraphQL\Language\AST\Node;
+use GraphQL\Utils\Value;
+use Throwable;
+
+/**
+ * @phpstan-type Path list<string>
```
spawnia, line 10: The path for values in an input list contains `int`. Also, let's differentiate this from the Path used for fields - how about `InputPath`? Same for the name of the property.

src/Utils/Value.php
```diff
@@ -1,12 +1,10 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
```
spawnia, line 3: Please revert unnecessary formatting changes

src/Error/CoercionError.php
```diff
@@ -0,0 +1,40 @@
...
+        ?array $valuePath = null
+    ) {
+        parent::__construct($message, $nodes, null, [], null, $previous);
+        $this->valuePath = $valuePath;
+    }
+
+    /**
+     * @return Path|null
+     */
+    public function getValuePath(): ?array
```
spawnia, line 36: Let's just make the property public

tests/Utils/CoerceValueTest.php
```diff
@@ -76,17 +80,23 @@ public function testCoercingAnArrayToGraphQLIDProducesAnError(): void
      * Describe: for GraphQLInt.
      *
      * @param mixed $result returned result
+     * @phpstan-param Path|null $expectedPath
```
spawnia, line 83: Regular `@param` is fine here

tests/Utils/CoerceValueTest.php
```diff
@@ -375,11 +385,11 @@ public function testReturnsErrorForIncorrectValueType(): void
     }
 
     /**
-     * @see it('returns no error for a valid input')
-     *
      * @param mixed $input
      *
      * @dataProvider validInputObjects
+     * @see          it('returns no error for a valid input')
```
spawnia, line 391: revert

spawnia (changes requested):

src/Utils/Value.php
Warxcell: Formatting is PITA :( 
Every project should have commited their editorconfig / idea settings. Reverting

src/Error/CoercionError.php
Warxcell: ValuePath? Since its not only for input but also for enum && scalar.

spawnia: Returned results are also values, but GraphQL distinguishes between input and output - see https://spec.graphql.org/draft/#sec-Input-and-Output-Types

Warxcell: Renamed to InputPath

tests/Utils/CoerceValueTest.php
```diff
@@ -76,17 +87,23 @@ public function testCoercingAnArrayToGraphQLIDProducesAnError(): void
...
         self::assertIsArray($errors);
         self::assertCount(1, $errors);
+        self::assertInstanceOf(CoercionError::class, $errors[0]);
         self::assertSame($expected, $errors[0]->getMessage());
 
         self::assertNull($result['value']);
+
+        if ($expectedPath !== null) {
+            self::assertSame($expectedPath, $errors[0]->inputPath);
+        }
```
spawnia, line 106: Shouldn't those be equivalent even when `null`?

tests/Utils/CoerceValueTest.php
```diff
@@ -375,11 +392,11 @@ public function testReturnsErrorForIncorrectValueType(): void
     }
 
     /**
-     * @see it('returns no error for a valid input')
```
spawnia, line 378: please self-review and revert changes like this

tests/Utils/CoerceValueTest.php
```diff
@@ -397,27 +414,32 @@ public function validInputObjects(): iterable
...
     /**
-     * @return iterable<int, array{mixed}>
+     * @return iterable<int, array{0: mixed, 1: string, 2: InputPath|null}>
      */
     public function invalidInputObjects(): iterable
     {
-        yield [123];
+        yield [['foo' => 1234, 'nested' => ['foobar' => null]], 'Expected non-nullable type Int! not to be null at value.nested.foobar.', ['nested', 'foobar']];
+        yield [['foo' => null, 'bar' => 1234], 'Expected non-nullable type Int! not to be null at value.foo.', ['foo']];
+        yield [123, 'Expected type TestInputObject to be an object.', null];
```
spawnia, line 437:
```suggestion
        yield [
            ['foo' => 1234, 'nested' => ['foobar' => null]],
            'Expected non-nullable type Int! not to be null at value.nested.foobar.',
            ['nested', 'foobar'],
        ];
        yield [
            ['foo' => null, 'bar' => 1234],
            'Expected non-nullable type Int! not to be null at value.foo.',
            ['foo'],
        ];
        yield [
            123,
            'Expected type TestInputObject to be an object.',
            null,
        ];
```

tests/Utils/CoerceValueTest.php
```diff
@@ -397,27 +414,44 @@ public function validInputObjects(): iterable
...
     /**
-     * @return iterable<int, array{mixed}>
+     * @return iterable<int, array{0: mixed, 1: string, 2: InputPath|null}>
      */
     public function invalidInputObjects(): iterable
     {
-        yield [123];
+        yield [
+            ['foo' => 1234, 'nested' => ['foobar' => null]],
+            'Expected non-nullable type Int! not to be null at value.nested.foobar.',
```
spawnia, line 437: Where does `value` come from? Seems misleading to me.

tests/Utils/CoerceValueTest.php
Warxcell: I can `vendor/bin/php-cs-fixer fix` before committing... seems its not working :shrug:

tests/Utils/CoerceValueTest.php
Warxcell: Removed check.

Issue #1199, opened 2022-07-15
StandardServer response dont support UTF-8


Kagayakashi:
StandardServer response dont support UTF-8.
Where json_encode function calls, need to add UTF8 support.

spawnia:
How is the current behaviour problematic?

Kagayakashi:
> How is the current behaviour problematic?

At this moment we are working on Graphql API and using StandardServer.
We have multilanguage response and cyrillic is not supported.

We can make a temporary solution if graphql-php release is delayed, but preferable avoid temporary solutions.

PR #1200, opened 2022-07-15
Add to emitResponse utf-8 support #1199

Kagayakashi:
In json emitResponse added utf8 support for issue #1199

### DIFF
--- a/tests/Server/HelperTest.php
+++ b/tests/Server/HelperTest.php
@@ -12,17 +12,15 @@ final class HelperTest extends TestCase {
      */
     public function testSendResponseWithUtf8Support(): void
     {
-        $dataArray = [
-            'data' => [
-                'user' => 'Петя'
-            ]
-        ];
-        
-        $expected = json_encode($dataArray, JSON_UNESCAPED_UNICODE);
+        $expected = '{"data":{"name":"Петя"}}';
+        $this->expectOutputString($expected);
 
         $helper = new Helper();
-        $helper->sendResponse($dataArray);
-
-        $this->expectOutputString($expected);
+        $data = [
+            'data' => [
+                'name' => 'Петя'
+            ]
+        ];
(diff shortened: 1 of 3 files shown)


spawnia:
Please add a failing test case first.

Kagayakashi:
> Please add a failing test case first.

Helper -> emitResponse method dont have return value, there is direct "echo".
So I added test semi hardcoded.

Will be fine?

spawnia:
https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#testing-output

tests/Server/HelperTest.php
```diff
@@ -0,0 +1,28 @@
...
+     */
+    public function testSendResponseWithUtf8Support(): void
+    {
+        $dataArray = [
+            'data' => [
+                'user' => 'Петя'
+            ]
+        ];
+        
+        $expected = json_encode($dataArray, JSON_UNESCAPED_UNICODE);
```
spawnia, line 21: That test is repeating the implementation, it should rather spell out the correct expected result.

tests/Server/HelperTest.php
```diff
@@ -0,0 +1,28 @@
...
+
+        $helper = new Helper();
+        $helper->sendResponse($dataArray);
+
+        $this->expectOutputString($expected);
```
spawnia, line 26:
```suggestion
 
        $this->expectOutputString($expected);
        $helper->sendResponse($dataArray);
```

tests/Server/HelperTest.php
Kagayakashi: I just also added "JSON_UNESCAPED_UNICODE" to actual code along with test.
Current master version dont have this.
Should it be replaced with strictly written json?

Like {"data":{"user":"Петя"}}

PR #1202, opened 2022-07-17
Sync Value::coerceInputValue() with reference implementation

spawnia:


src/Executor/Values.php
```diff
@@ -110,16 +110,23 @@ public static function getVariableValues(Schema $schema, NodeList $varDefNodes,
...
+                                $invalidValue = $error->printInvalidValue();
+
+                                $inputPath = $error->printInputPath();
+                                $message = "Variable \"\${$varName}\" got invalid value {$invalidValue}"
+                                    . ($inputPath !== null
+                                        ? " at \"{$varName}{$inputPath}\""
+                                        : '')
+                                    . '; ' . $error->getMessage();
 
                                 $errors[] = new Error(
```
Warxcell, line 127: I just came accross this - it actually replaces the error - shouldn't it be just 
```
$errors[] = $error;
```
rather than creating new error which hides the original one? or at least previous be `$error` not `$error->getPrevious()` ?

PR #1212, opened 2022-08-10
Show actual error in debug entries

### LINKED ISSUE #1211: debug entries doesn't include all the info
It doesn't show the origin of the error. 

Expected output:

image

Actual output:

image

Warxcell:
Fixes #1211


spawnia:
Please add a test case to ensure we do not introduce regressions in the future.

PR #1214, opened 2022-08-22
Preserve errors.

Warxcell:
https://github.com/webonyx/graphql-php/pull/1202#pullrequestreview-1041470219

### DIFF
--- a/src/Executor/Values.php
+++ b/src/Executor/Values.php
@@ -123,7 +123,7 @@ class Values
                                     $error->getSource(),
                                     $error->getPositions(),
                                     $error->getPath(),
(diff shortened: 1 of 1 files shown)


spawnia:
I don't see how this improves this library. The message from the coercion error is included within the new error.

Warxcell:
> I don't see how this improves this library. The message from the coercion error is included within the new error. Anyway, isn't this PR obsolete with #1202?

Messages is preversed, but not the original exception - which creates a few issues when formatting exceptions. 

Otherwise yes, this change is included in linked PR, I just hoped to get it a bit faster into master. :)

PR #1233, opened 2022-10-18
Add ability to resolve union type of object via __typename

N-Silbernagel:
We can currently specify the type of a graphlql interface/union type through the __typename key on arrays. We cannot however do the same for objects, except for through the ArrayAccess interface.

To make this possible, I added a "Resolvable" interface. Users of the library can make use of this by implementing the sole method of the interface "resolveGraphQLType". Another option would be to check for a __typename public property but that is IMHO the inferior solution.

I have got one failing test which is EnumTypeTest/testAllowsSimpleArrayAsValues. That test was already broken when I forked the project tho.

Looking forward to hearing your feedback.

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -30,7 +30,6 @@ use GraphQL\Language\AST\Node;
 use GraphQL\Language\AST\OperationDefinitionNode;
 use GraphQL\Language\AST\SelectionNode;
 use GraphQL\Language\AST\SelectionSetNode;
-use GraphQL\Resolvable;
 use GraphQL\Type\Definition\AbstractType;
 use GraphQL\Type\Definition\Directive;
 use GraphQL\Type\Definition\FieldDefinition;
@@ -1102,8 +1101,10 @@ class ReferenceExecutor implements ExecutorImplementation
             return $value['__typename'];
         }
 
-        if ($value instanceof Resolvable) {
-            return $value->resolveGraphQLType();
+        if (is_object($value)
+            && isset($value->__typename)
+        ) {
+            return $value->__typename;
         }
 
         if ($abstractType instanceof InterfaceType && $info->schema->getConfig()->typeLoader !== null) {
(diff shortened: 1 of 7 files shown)


src/Executor/ReferenceExecutor.php
```diff
@@ -1094,15 +1095,17 @@ protected function completeAbstractValue(
...
         // First, look for `__typename`.
-        if (
-            $value !== null
-            && (is_array($value) || $value instanceof ArrayAccess)
+        if ((is_array($value) || $value instanceof ArrayAccess)
             && isset($value['__typename'])
             && is_string($value['__typename'])
         ) {
             return $value['__typename'];
         }
```
spawnia, line 1103: Instead of introducing a new interface, I think we should rather extend this part to have parity with the way the default field resolver accesses keys on the value - see https://github.com/webonyx/graphql-php/blob/03c630d21858eae50898996a40650d6abca3da23/src/Executor/Executor.php#L193-L201

This will allow objects with a property `__typename` - public or virtual - to work, which I think is conceptually much simpler then a separate interface.

N-Silbernagel: You are right, it would be more in sync how the rest of the library works.
My rational was that I think an interface makes the intent much clearer.

Im gonna adjust my PR to use the public property approach you proposed.

N-Silbernagel:
> If you closely read the documentation for [Interfaces](https://webonyx.github.io/graphql-php/type-definitions/interfaces) and [Unions](https://webonyx.github.io/graphql-php/type-definitions/unions), I think you will find there is not really a need for such an interface. Implementing `resolveType` is a more elegant approach to resolve abstract types, as it does not force your data objects to even be aware of the GraphQL layer. As described in my other comment, we can (and I think we should) extend how access to `__typename` is attempted, but I see no need to introduce additional complexity beyond that.

If I understand the documentation correctly, the `resolveType` method does only work when defining graphql schemas via php files. We are using the Schema definition language and I find it tedious to define another decorator for types defined in .grahql files.

spawnia:
You can do a combination of SDL and programmatic implementations of types - [nuwave/lighthouse](https://github.com/nuwave/lighthouse) works that way.

I generalized the solution and testing strategy in https://github.com/webonyx/graphql-php/pull/1234.

PR #1234, opened 2022-10-18
Add ability to resolve abstract type of object via `__typename`

spawnia:
Resolves https://github.com/webonyx/graphql-php/pull/1233

Thanks @N-Silbernagel


N-Silbernagel:
Thanks for accepting this change and the fast responses, appreciate it a lot. 
Am I right in thinking this is going to get merged into 15.x? Is there a roadmap for 15.x?

spawnia:
Yeah, this will be in 15.x. I am currently gathering feedback for the alpha release, will do a beta in a couple weeks and then go for a stable release.

N-Silbernagel:
Nice, thanks again :)

PR #1241, opened 2022-11-03
fix: allow to have queryplans with different options

simPod:
Currently, I'm unable to instantiate two different query plans.


spawnia:
Makes sense.

PR #1242, opened 2022-11-03
chore: remove psalm

simPod:
The project got "stale". It helped to resolve some issues when we introduced phpdoc types but it [causes another issues](https://github.com/webonyx/graphql-php/actions/runs/3389829138/jobs/5633332863). Phpstan matured significantly meanwhile.  
I don't see a benefit of keeping it anymore.


spawnia:
I was waiting for you to pick up on the failing analysis, after spending some time fiddling around with the type definitions I really did not feel like investing more time to satisfy psalm was a worthwhile activity.

I automated class-reference regeneration, you can just revert the file to the latest changes in `master` and let GitHub actions handle it.

simPod:
Yup, I noticed it's failing just right now in #1241.

I investigated related failures before, isolated them and opened issues in psalm but it's broken still (no blame on psalm guys). They're parsing docblocks using regex which seems like a way to hell to me 🤷🏾‍♂️.

simPod:
_BTW you can ping me next time if there's something I missed and can solve_

PR #1251, opened 2022-11-16
run tests on php8.2 + fix deprecation

fezfez:


spawnia:
I am actually fixing the deprecations in https://github.com/webonyx/graphql-php/pull/1252 instead of ignoring them.

PR #1252, opened 2022-11-16
Run CI on PHP 8.2

spawnia:


simPod:
@spawnia only with `AllowDynamicProperties`

https://phpstan.org/r/d48f3a62-0cb0-44eb-9a31-90cc3a6f5d91

PR #1255, opened 2022-12-10
Replace AST interfaces `@property` declarations with methods

LastDragon-ru:
Interfaces cannot have properties. Moreover,

1) `@property` doesn't work in PHPStan on PHP 8.2 (https://github.com/phpstan/phpstan/issues/8501)
2) psalm report "ERROR: [NoInterfaceProperties](https://psalm.dev/028) - 33:6 - Interfaces cannot have properties" (https://psalm.dev/r/68960d2827)

So probably will be better to use methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,9 +52,9 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Store rules exclusively by class name in `DocumentValidator`
 - Reorder standard types as described in the GraphQL specification
 - Improve runtime performance by moving checks for duplicate/mismatching type instances to `assert()` or schema validation
-- Replace `HasSelectionSet::$selectionSet` to `HasSelectionSet::getSelectionSet()`
-- Replace `TypeDefinitionNode::$name` to `TypeDefinitionNode::getName()`
-- Replace `TypeExtensionNode::$name` to `TypeExtensionNode::getName()`
+- Replace `HasSelectionSet::$selectionSet` with `HasSelectionSet::getSelectionSet()`
+- Replace `TypeDefinitionNode::$name` with `TypeDefinitionNode::getName()`
+- Replace `TypeExtensionNode::$name` with `TypeExtensionNode::getName()`
 
 ### Added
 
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -52,6 +52,9 @@ You can find and compare releases at the [GitHub release page](https://github.co
 - Store rules exclusively by class name in `DocumentValidator`
 - Reorder standard types as described in the GraphQL specification
 - Improve runtime performance by moving checks for duplicate/mismatching type instances to `assert()` or schema validation
+- Replace `HasSelectionSet::$selectionSet` to `HasSelectionSet::getSelectionSet()`
+- Replace `TypeDefinitionNode::$name` to `TypeDefinitionNode::getName()`
+- Replace `TypeExtensionNode::$name` to `TypeExtensionNode::getName()`
```
spawnia, line 57:
```suggestion
- Replace `HasSelectionSet::$selectionSet` with `HasSelectionSet::getSelectionSet()`
- Replace `TypeDefinitionNode::$name` with `TypeDefinitionNode::getName()`
- Replace `TypeExtensionNode::$name` with `TypeExtensionNode::getName()`
```

src/Utils/ASTDefinitionBuilder.php
```diff
@@ -188,11 +188,17 @@ private function buildWrappedType(TypeNode $typeNode): Type
...
     public function buildType($ref): Type
     {
-        if (is_string($ref)) {
-            return $this->internalBuildType($ref);
+        $type = null;
+
+        if ($ref instanceof TypeDefinitionNode) {
+            $type = $this->internalBuildType($ref->getName()->value, $ref);
+        } elseif ($ref instanceof NamedTypeNode) {
+            $type = $this->internalBuildType($ref->name->value, $ref);
```
spawnia, line 196: Why don't we add `getName()` to that as well and go back to the simple implementation?

LastDragon-ru: Properties are fine for classes 🤷‍♂️

PR #1256, opened 2022-12-13
PHP 8.2 support

emteknetnz:
Add the `#[\AllowDynamicProperties]` attribute to squash deprecation warning for [dynamic properties](https://php.watch/versions/8.2/dynamic-properties-deprecated) e.g. `$this->$name = $value;` within `__set()` method

Change deprecated string interpolation `"${var}"` to `sprintf()` -  I did not use `"$var"` or `"{$var}" syntax as that failed linting e.g. `Squiz.Strings.DoubleQuoteUsage.ContainsVar e.g. Variable "$message" not allowed in double quoted`

Adds PHP 8.2 to CI matrix

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


emteknetnz:
OK fair enough - do you know roughly when the stable for 15 will be released?

Issue #1259, opened 2022-12-16
PHP 8.2 compatibility issue on 14.11.8


ldebrouwer:
When using version 14.11.8 on PHP 8.2, deprecation warnings are thrown for the following;

```
PHP Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in /var/www/html/vendor/webonyx/graphql-php/src/Validator/Rules/ValuesOfCorrectType.php on line 182
```

spawnia:
I am no longer maintaining v14, this will be fixed in the upcoming v15.

PR #1260, opened 2022-12-16
Using `${var}` in strings is deprecated, use `{$var}` instead.

ldebrouwer:
See issue #1259.


spawnia:
I am no longer maintaining v14, this will be fixed in the upcoming v15.

PR #1280, opened 2023-01-17
Avoid deprecated interpolation syntax

jasonvarga:
This prevents a deprecation message in PHP 8.2.


spawnia:
I no longer maintain 14.x.

PR #1287, opened 2023-01-24
Simplify blog example TypeRegistry

spawnia:


examples/01-blog/Blog/Type/CommentType.php
```diff
@@ -17,33 +18,33 @@ public function __construct()
         parent::__construct([
             'name' => 'Comment',
             'fields' => static fn (): array => [
-                'id' => Types::id(),
+                'id' => Type::id(),
                 'author' => [
-                    'type' => Types::user(),
+                    'type' => TypeRegistry::type(UserType::class),
```
shmax, line 23: Can I ask what the motivation was to get rid of the `Types` helper? Personally, I kind of like the pattern; I can right-click something like `user` and choose "Find usages" in my IDE to quickly see where it's being used. I suppose you can technically do the same thing with the class name itself, but you will get cross-pollution from the class itself, static usages, and so on.

shmax: Okay, so you lose about a hundred lines of code, but you also lose some handy tooling. If your goal is just to drop lines of code, you could always tighten up Types.php. I use a similar pattern in my own code, and my Types.php looks like this:

```
    public static function action() : callable { return static::get(ActionType::class);}
    public static function actionSort() : callable { return static::get(ActionSortType::class);}
    public static function actions() : callable { return static::get(ActionsType::class);}
    public static function appraisal() : callable { return static::get(AppraisalType::class);}
    public static function appraisalSet() : callable { return static::get(AppraisalSetType::class);}
```

simPod (approved):

PR #1288, opened 2023-01-24
Stanning for tests

shmax:
Stanning for some more low-hanging fruit.

### DIFF
--- a/tests/Language/VisitorTest.php
+++ b/tests/Language/VisitorTest.php
@@ -68,7 +68,8 @@ final class VisitorTest extends ValidatorTestCase
         if ($parent instanceof NodeList) {
             self::assertEquals($node, $parent[$key]);
         } else {
-            self::assertEquals($node, $parent->{$key});
+            $vars = get_object_vars($parent);
+            self::assertEquals($node, $vars[$key]);
         }
 
         self::assertEquals($node, $this->getNodeByPath($ast, $path));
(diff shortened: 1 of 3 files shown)


tests/Language/VisitorTest.php
```diff
@@ -68,7 +68,8 @@ private function checkVisitorFnArgs(DocumentNode $ast, array $args, bool $isEdit
         if ($parent instanceof NodeList) {
             self::assertEquals($node, $parent[$key]);
         } else {
-            self::assertEquals($node, $parent->{$key});
+            $vars = get_object_vars($parent);
```
shmax, line 71: This isn't any more type-safe than what was here before and is probably less performant. Any better ideas?

spawnia: Perhaps check `propery_exists()` before to ensure the dynamic access is fine.

shmax: Hmm, wouldn't that make it harder to catch runtime errors?

spawnia: Well check and fail the test if it does not exist.

spawnia: We can also just ignore phpstan here.

shmax: SGTM. bd02774f

shmax:
@spawnia thanks very much for reviewing 👍

PR #1289, opened 2023-01-30
ci: rework phpstan to support language version specific configs

simPod:


phpstan/phpstan/include-by-php-version.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
```
spawnia, line 3:
```suggestion
<?php declare(strict_types=1);
```

phpstan/phpstan/php-82.neon
```diff
@@ -0,0 +1,13 @@
...
+        # but all implementors of the interfaces actually have those properties.
+        - "~Access to an undefined property .*GraphQL\\\\Type\\\\Definition\\\\NamedType.*::\\$name~"
+        - "~Access to an undefined property .*GraphQL\\\\Type\\\\Definition\\\\NamedType.*::\\$description~"
+        - "~Access to an undefined property .*GraphQL\\\\Type\\\\Definition\\\\NamedType.*::\\$astNode~"
+        - "~Access to an undefined property .*GraphQL\\\\Type\\\\Definition\\\\NamedType.*::\\$extensionASTNodes~"
+        - "~Access to an undefined property .*GraphQL\\\\Language\\\\AST\\\\TypeDefinitionNode.*::\\$name~"
+        - "~Access to an undefined property .*GraphQL\\\\Language\\\\AST\\\\TypeExtensionNode.*::\\$name~"
+
+        # In PHP 8.2, PHPStan seems to not allow dynamic checks for properties on classes, even when they are not final.
+        - "~Call to function property_exists\\(\\) with .* will always evaluate to false\\.~"
```
spawnia, line 13: I think this is or will be fixed: https://github.com/phpstan/phpstan/issues/8474

simPod: We can then remove the ignores but I'd keep this _version-based include framework_ so we can handle similar issues more easily in the future without having `reportUnmatched: false`.

simPod: In order to avoid this

image

spawnia: sounds good

PR #1292, opened 2023-02-01
Allow sorting more items in SchemaPrinter

ruudk:
This change allows you to configure the SchemaPrinter and define what should be sorted and what not.

`sortArguments` allows to sort arguments used in fields on objects and interfaces.

`sortFields` allows to sort fields inside objects and interfaces.

`sortInputFields` allows to sort fields inside input objects.

`sortEnumValues` allows to sort enum values.

It's important to know that changing the order of fields inside input objects can have severe consequences for clients that have generated code based on that order. For example, Apollo for Kotlin and Swift takes the ordering from the input object to create the input objects.

When the order in the schema changes, so does the order of the constructor arguments for those objects. When the calling code doesn't use named arguments, they can pass wrong values.

That's why I made sorting of input object fields a separate option.

### DIFF
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -27,7 +27,15 @@ use GraphQL\Type\Schema;
 /**
  * Prints the contents of a Schema in schema definition language.
  *
- * @phpstan-type Options array{sortTypes?: bool}
+ * All sorting options sort alphabetically. If not given or `false`, the original schema definition order will be used.
+ *
+ * @phpstan-type Options array{
+ *   sortArguments?: bool,
+ *   sortEnumValues?: bool,
+ *   sortFields?: bool,
+ *   sortInputFields?: bool,
+ *   sortTypes?: bool,
+ * }
  */
 class SchemaPrinter
 {
@@ -239,6 +247,10 @@ class SchemaPrinter
             return '';
         }
 
+        if (isset($options['sortArguments']) && $options['sortArguments'] === true) {
+            usort($args, static fn (Argument $left, Argument $right): int => $left->name <=> $right->name);
+        }
+
         $allArgsWithoutDescription = true;
         foreach ($args as $arg) {
             $description = $arg->description;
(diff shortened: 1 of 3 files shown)


ruudk:
@spawnia After a few rebases the PR is now green. If you have time to review that would be really appreciated! Thanks 💙

src/Utils/SchemaPrinter.php
```diff
@@ -337,16 +341,22 @@ protected static function printObject(ObjectType $type, array $options): string
      */
     protected static function printFields(array $options, $type): string
     {
-        $fields = [];
+        $items = [];
```
spawnia, line 344: Let's keep this as `$fields` and change the other variable to `$fieldDefinitions`. Same with the other changes like this.

ruudk: @spawnia Done!

src/Utils/SchemaPrinter.php
```diff
@@ -27,7 +27,7 @@
 /**
  * Prints the contents of a Schema in schema definition language.
  *
- * @phpstan-type Options array{sortTypes?: bool}
+ * @phpstan-type Options array{sortTypes?: bool, sortArguments?: bool, sortFields?: bool, sortInputFields?: bool, sortEnumValues?: bool}
```
spawnia, line 30:
```suggestion
 * All sortX options sort alphabetically. If not given or `false`, the original schema definition order will be used.
 *
 * @phpstan-type Options array{
 *   sortTypes?: bool,
 *   sortArguments?: bool,
 *   sortFields?: bool,
 *   sortInputFields?: bool,
 *   sortEnumValues?: bool,
 * }
```

ruudk: Done

spawnia:
Thanks, released with https://github.com/webonyx/graphql-php/releases/tag/v15.1.0

simPod:
I have to sort already in SchemaConfig because tools like Sailor does not use SchemaPrinter

WDYT about moving sorting to SchemaConfig?

spawnia:
Sorting in the schema might be difficult with lazyness, but I could see adding it in some way.

PR #1300, opened 2023-02-03
Use PHPUnit 10

spawnia:
Awaiting https://github.com/rdohms/phpunit-arraysubset-asserts/pull/70


ruudk:
Just an idea: why not drop the dependency and own the implementation?

spawnia:
> Just an idea: why not drop the dependency and own the implementation?

I don't see a big benefit to using PHPUnit 10 just now, but see a cost to inlining the implementation.

PR #1313, opened 2023-02-15
Add method NodeList::reindex() for AST manipulation

spawnia:
See https://github.com/spawnia/sailor/pull/88

### DIFF
--- a/src/Language/AST/NodeList.php
+++ b/src/Language/AST/NodeList.php
@@ -102,7 +102,7 @@ class NodeList implements \ArrayAccess, \IteratorAggregate, \Countable
     }
 
     /**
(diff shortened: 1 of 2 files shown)


simPod (approved):
This seems like the best solution at the moment.

Though, I'd rather have NodeList to behave like a list of nodes all the time. But to achieve that, we would have needed to remove ArrayAccess and then have something like `remove($node)` instead of supporting `unset($nodelist[$index])`. And such change would have been significantly breaking. But I can look into that if you want.

spawnia:
I could see removing `ArrayAccess` in favor of an API with explicit methods, sure.

PR #1314, opened 2023-02-15
fix: do not add `false` to NodeList

simPod:
Visitor like this one https://github.com/webonyx/graphql-php/blob/76e165607d1a0f5eb47e8dbd8d653fa9661d0281/src/Validator/QueryValidationContext.php#L132 may return `false` or `void` https://github.com/webonyx/graphql-php/blob/76e165607d1a0f5eb47e8dbd8d653fa9661d0281/src/Language/Visitor.php#L275 which is interpreted as `false` as well.

However, we were pushing the false into a `NodeList` which is wrong I believe since it should contain nodes only.

And also based on tests I came to a conclusion that `false` is valid return value from NodeVisitor so I handled it only for `NodeList`.

### DIFF
--- a/src/Language/Visitor.php
+++ b/src/Language/Visitor.php
@@ -6,6 +6,7 @@ use GraphQL\Language\AST\Node;
 use GraphQL\Language\AST\NodeKind;
 use GraphQL\Language\AST\NodeList;
 use GraphQL\Utils\TypeInfo;
+use GraphQL\Utils\Utils;
 
 /**
  * Utility for efficient AST traversal and modification.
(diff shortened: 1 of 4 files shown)


simPod:
@spawnia I believe this is correct visitor type. Can you confirm?

Also, having `A|mixed` is `mixed` so we should get rid of it everywhere I think.

spawnia:
I think extracting a type variable should be possible here, let me know when you get it right.

src/Language/Visitor.php
```diff
@@ -224,7 +225,9 @@ public static function visit(object $root, array $visitor, ?array $keyMap = null
                             ++$editOffset;
                         } else {
                             if ($node instanceof NodeList || \is_array($node)) {
-                                $node[$editKey] = $editValue;
+                                if ($editValue !== false) {
```
spawnia, line 228: I think we should throw if `editValue` is not a `Node` here

simPod: See the commit, that breaks tests since it happens very often (most of visitors return void, which is coerced to false).

spawnia: > coerced to false

`void` will become `null`.

simPod: You're right, I got lost in callbacks.

But if I throw, everything kind of breaks apart. `false` seems to be a valid value, just not something that needs to be added to the NodeList

spawnia: Found and fixed the culprit.

simPod: nice

PR #1316, opened 2023-02-15
Update visitor docs

spawnia:
Regarding https://github.com/webonyx/graphql-php/pull/1314 - I believe the docs to be misleading. Returning `false` or any other value other than a `VisitorOperation`, `null` or a `Node` seems invalid to me.

### DIFF
--- a/docs/class-reference.md
+++ b/docs/class-reference.md
@@ -1033,34 +1033,29 @@ Utility for efficient AST traversal and modification.
 the visitor's enter function at each node in the traversal, and calling the
 leave function after visiting that node and all of its child nodes.
 
-By returning different values from the enter and leave functions, the
-behavior of the visitor can be altered, including skipping over a sub-tree of
-the AST (by returning false), editing the AST by returning a value or null
-to remove the value, or to stop the whole traversal by returning BREAK.
+By returning different values from the `enter` and `leave` functions, the
+behavior of the visitor can be altered.
+
+- no return (`void`) or return `null`: no action
+- `Visitor::skipNode()`: skips over the subtree at the current node of the AST
+- `Visitor::stop()`: stop the Visitor completely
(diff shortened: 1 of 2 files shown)


docs/class-reference.md
```diff
@@ -1033,34 +1033,29 @@ Utility for efficient AST traversal and modification.
...
-the AST (by returning false), editing the AST by returning a value or null
-to remove the value, or to stop the whole traversal by returning BREAK.
+By returning different values from the `enter` and `leave` functions, the
+behavior of the visitor can be altered.
+
+- no return (`void`) or return `null`: no action
+- `Visitor::skipNode()`: skips over the subtree at the current node of the AST
+- `Visitor::stop()`: stop the Visitor completely
+- `Visitor::removeNode()`: remove the current node
+- return any other value: replace this node with the returned value
```
simPod, line 1043: What about introducing something like `Visitor::replaceNode($value)`? So be don't allow returning everything from a Visitor

spawnia: I like it a lot!

spawnia: Do you want to put that together in https://github.com/webonyx/graphql-php/pull/1314?

spawnia: Hm actually, that breaks just about any user of the visitor and does not really improve type safety. Primitive values should also be allowed, for example when renaming stuff. I just think that returning `false` does not make sense (unless changing the value of a `BooleanValueNode`).

simPod: So do we keep return type `mixed` then?

spawnia: I think so, yes. `Visitor::replaceNode()` would have to take `mixed` as well - while the intent is clearer, I am not sure it is worth the breakage.

PR #1326, opened 2023-02-27
Initialize `directives` and `arguments` in `Node` with empty `NodeList`

shmax:
Hey folks, I finally got around to updating to 15.0 today, and a few things exploded. Similar to the [issue I reported a while ago](https://github.com/webonyx/graphql-php/pull/1272), I run into issues when calling `AST::fromArray` on AST arrays that are generated without empty placeholder arrays for things like `directives` and `arguments`. 

My previous approach was to coalesce those fields to empty array in the consuming code, but this time there were a lot more of those access points to guard against, so I thought I'd try just initializing `directives` and `arguments` in the constructors.

There are, of course, several other ways we could approach this:
* make these fields private and make getters that coalesce to empty lists
* put `isset` and coalesce checks everywhere where these things are accessed

We could also discuss modifying our AST generation so that it doesn't add empty placeholders (~thus mirroring the JavaScript version of GraphQL~ edit: actually done by a [3rd-party optimization tool](https://www.npmjs.com/package/@graphql-tools/optimize)), which would make it harder for this kind of problem to slip past us.

I'm open to anything. Please let me know what you think.

### DIFF
--- a/src/Language/AST/DirectiveDefinitionNode.php
+++ b/src/Language/AST/DirectiveDefinitionNode.php
@@ -17,4 +17,10 @@ class DirectiveDefinitionNode extends Node implements TypeSystemDefinitionNode
 
     /** @var NodeList<NameNode> */
     public NodeList $locations;
+
+	public function __construct(array $vars)
+	{
+		parent::__construct($vars);
+		$this->arguments = $this->arguments ?? new NodeList([]);
+	}
 }
--- a/src/Language/AST/DirectiveDefinitionNode.php
+++ b/src/Language/AST/DirectiveDefinitionNode.php
@@ -21,6 +21,6 @@ class DirectiveDefinitionNode extends Node implements TypeSystemDefinitionNode
     public function __construct(array $vars)
     {
         parent::__construct($vars);
-        $this->arguments = $this->arguments ?? new NodeList([]);
+        $this->arguments ??= new NodeList([]);
     }
 }
--- a/tests/Utils/AstFromArrayTest.php
+++ b/tests/Utils/AstFromArrayTest.php
@@ -1,0 +1,142 @@
+<?php declare(strict_types=1);
+
+namespace GraphQL\Tests\Utils;
+
+use GraphQL\Error\SerializationError;
+use GraphQL\Language\AST\BooleanValueNode;
+use GraphQL\Language\AST\EnumValueNode;
+use GraphQL\Language\AST\FieldNode;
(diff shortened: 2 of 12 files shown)


src/Language/AST/DirectiveDefinitionNode.php
```diff
@@ -17,4 +17,10 @@ class DirectiveDefinitionNode extends Node implements TypeSystemDefinitionNode
 
     /** @var NodeList<NameNode> */
     public NodeList $locations;
+
+    public function __construct(array $vars)
+    {
+        parent::__construct($vars);
+        $this->arguments = $this->arguments ?? new NodeList([]);
```
spawnia, line 24:
```suggestion
        $this->arguments ??= new NodeList([]);
```

tests/Utils/AstFromArrayTest.php
```diff
@@ -0,0 +1,128 @@
...
+use GraphQL\Utils\AST;
+use PHPUnit\Framework\TestCase;
+
+final class AstFromArrayTest extends TestCase
+{
+    public function testVariableDefinitionArrayFromArray(): void
+    {
+        /**
+         * @var VariableDefinitionNode
+         */
```
spawnia, line 21: use `assert()` over `@var`

spawnia (changes requested):

spawnia:
I can see benefits to having a smaller AST, e.g. caching.

src/Language/AST/DirectiveDefinitionNode.php
shmax: Ah, forgot that was a thing, thank you ece7502bd75ac0274fa51a6f65ffe72691881a04

tests/Utils/AstFromArrayTest.php
shmax: 👍 2f0340bd

shmax:
> SGTM. Would you like me to tackle that in this PR, or in a different one?

Let me see if I can weasel out of that one, for now. I did a little debugging, and it's nothing to do with the reference library--it's actually coming from a setting (that defaults to `true`) in the graphql-codeden package I'm working with:
https://github.com/dotansimha/graphql-code-generator/blob/f08e4f29f173ee704250cf999d4037232a7779fd/website/public/config.schema.json#L263

spawnia (approved):

shmax:
Beautiful, thanks much for the quick review 👍

PR #1330, opened 2023-03-02
Before fetching a type, first check if it exists

ruudk:
When upgrading from 15.1.0 to 15.2.1 I noticed the following error in our setup:

```
  Type "Subscription" not found in name to fully qualified mapping
```

We have a very strict service locator. Any type that is retrieved should exist, or fail.

This works fine, and allows us to quickly spot invalid referenced types.

But this broke in f82c04f191906150bfb0ca3458f238d24312d839

I think it's better to first check if a type exists, before comparing it.

### DIFF
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -186,19 +186,9 @@ class SchemaPrinter
      */
     protected static function hasDefaultRootOperationTypes(Schema $schema): bool
     {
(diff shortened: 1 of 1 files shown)


ruudk:
Pushed a second commit, that is way simpler. By first checking if it's not null, we ensure that the `getType` call doesn't happen.

ruudk:
I'll update my loader to return null on failure :(

PR #1331, opened 2023-03-02
feat: add missing @throws to Type constructors

simPod:
I started using this https://phpstan.org/blog/bring-your-exceptions-under-control and having `@throws` in the library is better for users since they can infer it and don't have to explicitly specify in their codebases.

### DIFF
--- a/src/Type/Definition/EnumType.php
+++ b/src/Type/Definition/EnumType.php
@@ -63,6 +63,8 @@ class EnumType extends Type implements InputType, OutputType, LeafType, Nullable
     private array $nameLookup;
 
     /**
+     * @throws InvariantViolation
+     *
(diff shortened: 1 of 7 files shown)


spawnia:
I think adding `@throws` is a very slippery slope to take. Higher level methods accumulate a lot of `@throws`.

simPod:
Sure, when not handled it does. But it's valuable information, at least user knows what to catch.

I have decided to go that way when I discovered I was catching wrong exceptions that were not thrown at all and instead allowing bubbling up those I wanted to catch. And it can be easily analysed with phpstan so I think it's valuable.

spawnia:
We might as well delete them all as a first step, at least that would not give the wrong impression that they are somewhat informative.

PR #1335, opened 2023-03-03
Differentiate between const and non-const AST nodes

spawnia:
- [x] Check where `ConstValueNode` needs to be used

### DIFF
--- a/src/Validator/Rules/UniqueInputFieldNames.php
+++ b/src/Validator/Rules/UniqueInputFieldNames.php
@@ -3,6 +3,7 @@
 namespace GraphQL\Validator\Rules;
 
 use GraphQL\Error\Error;
+use GraphQL\Language\AST\ConstObjectFieldNode;
 use GraphQL\Language\AST\NameNode;
 use GraphQL\Language\AST\NodeKind;
 use GraphQL\Language\AST\ObjectFieldNode;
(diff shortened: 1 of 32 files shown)


src/Validator/Rules/UniqueInputFieldNames.php
```diff
@@ -54,7 +55,8 @@ public function getASTVisitor(ValidationContext $context): array
                     $this->knownNames = $knownNames;
                 },
             ],
-            NodeKind::OBJECT_FIELD => function (ObjectFieldNode $node) use ($context): VisitorOperation {
+            NodeKind::OBJECT_FIELD => function ($node) use ($context): VisitorOperation {
```
simPod, line 58:
```suggestion
            NodeKind::OBJECT_FIELD => function (mixed $node) use ($context): VisitorOperation {
```

is this already supported in this php version?

simPod (commented):
TBH I'm not really able to evaluate this addition since I don't have the required gql expertise but I don't see any harm in this either.

src/Validator/Rules/UniqueInputFieldNames.php
spawnia: nah, `mixed` is PHP 8. We could do `Node` or something, but I think just the `assert()` is cleanest.

PR #1336, opened 2023-03-05
Don't unnecessarily invoke type loader when printing schema

mfn:
## Summary
Before https://github.com/webonyx/graphql-php/pull/1303 , the type loader was not invoked when printing the schema and not mutation or subscription is defined.

This brings back the previous behaviour in this regard.

I'm not sure what is really wanted here, it was discovered as feedback in the preparation for graphql-laravel for the graphql-php v15 upgrade in https://github.com/rebing/graphql-laravel/pull/953#issuecomment-1449855145 and we added a test case which did not trigger the typeloader in v14 ( https://github.com/rebing/graphql-laravel/pull/995 , though at that point it wasn't yet clear what's going on).

It's not a big issue, but I figured I open the PR as discussion, if we really need to invoke the type loader if we already see that there's not such top level type defined.

The test is just quickly put together: if you change back to the code in master, the test throws an exception.

Let me know what you think, cheers.

### DIFF
--- a/src/Utils/SchemaPrinter.php
+++ b/src/Utils/SchemaPrinter.php
@@ -186,9 +186,12 @@ class SchemaPrinter
      */
     protected static function hasDefaultRootOperationTypes(Schema $schema): bool
     {
+        $mutationType = $schema->getMutationType();
+        $subscriptionType = $schema->getSubscriptionType();
+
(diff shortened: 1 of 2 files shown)


spawnia:
The type loader should not throw when a type is not found. Such behaviour would be problematic because types can be requested by the client, e.g. in a fragment spread.

PR #1337, opened 2023-03-06
test: run all NonNullTests

simPod:


simPod:
I don't know of any. Maybe some dead code analysis based on finding unused methods while considering `test*()`, dataproviders and phpunit's methods as used.

Mb with this https://psalm.dev/articles/better-unused-variable-detection

PR #1339, opened 2023-03-06
Add rector SetList::DEAD_CODE

spawnia:


spawnia:
@simPod sadly this did not find the issue with dead test code either.

src/Error/Error.php
```diff
@@ -19,6 +19,8 @@
  *
  * Class extends standard PHP `\Exception`, so all standard methods of base `\Exception` class
  * are available in addition to those listed below.
+ *
+ * @see \GraphQL\Tests\Error\ErrorTest
```
simPod, line 23: I think we should not add these.

1. this is nonexistent symbol when installed via composer since we don't bundle tests
2. it's better to use `@covers` to link test to the classes it covers

src/Executor/ExecutionResult.php
```diff
@@ -66,14 +66,14 @@ class ExecutionResult implements \JsonSerializable
      *
      * @phpstan-var ErrorFormatter|null
      */
-    private $errorFormatter = null;
+    private $errorFormatter;
```
simPod, line 69: Not sure if this is correct? I don't see it initialized in constructor

spawnia: It is an untyped property, so it does not need initialization.

PR #1363, opened 2023-04-14
Update lazy docs

shmax:
For no particular reason, I just realized that after all that work we did on lazy loading, we didn't remember to update the docs.

I have no idea if what I've done here is on the right track, or if it's maybe going to be a bit too hairy for casual readers to easily understand, but as always my main goal here is to get the conversation started.

Side note: I'm not sure if the "fields" callable thing provides any appreciable benefit once you are using callables for everything, anyway. If not, perhaps we could downplay that section in the docs...


shmax:
Er, I don't know why the fixer is going berserk and touching everything. Any insight?

shmax:
>  I applied the changes in master.

Thank you, buddy. Updated.

docs/schema-definition.md
```diff
@@ -160,8 +160,64 @@ $schema = new Schema([
 You can automate this registry if you wish to reduce boilerplate or even
 introduce a Dependency Injection Container if your types have other dependencies.
 
-Alternatively, all methods of the registry could be static - then there is no need
-to pass it in the constructor - instead use **TypeRegistry::myAType()** in your type definitions.
+This approach has its drawbacks; while using a callable for the **fields** property does provide some protection, once the fields property is accessed each individual field under its governance is still evaluated whether it's needed for the current query or not, and this can result in many superfluous field and argument instantiations. For a more robust solution, you can skip the type loader altogether and instead craft a type registry populated with static helpers that return callables for each individual type:
```
spawnia, line 163: > skip the type loader altogether

I don't think that will work, unless the schema is constructed with `types`. Retrieving types by name is necessary in cases where the client defines a type name, e.g. variables, fragments.

shmax: Seems to work just fine in my own code, but to be fair I didn't ensure I've covered all edge cases. For now I'll reword that...

docs/schema-definition.md
```diff
@@ -160,8 +160,64 @@ $schema = new Schema([
...
+    {
+        // by moving our custom types to separate files and returning a callable here, we can prevent
+        // even more needless work from happening
+        return static fn () => self::byClassName($classname);
+    }
+
+    private static function byClassName(string $classname): Type
+    {
+        $parts = \explode('\\', $classname);
+        $cacheName = \strtolower(\preg_replace('~Type$~', '', $parts[\count($parts) - 1]));
```
shmax, line 210: I don't actually remember why I do this; seems like I should be able to just use the classname as a cache key...

spawnia: Seems like it converts to the type name that is used in the schema, thus allowing lookup from the type loader function.

docs/schema-definition.md
shmax: Is that for selecting on a union? That seems to work just fine, too, presumable because I use the same static helpers everywhere, eg:

```sdl
// a piece of a query from my code
fragment skuEdges on Skus {
    items {
      edges {
        ... on SkuRow {
          sku {
            ...smallSku
          }
        }
      }
    }
  }
```

```php
// my union resolver:
class SortedSkuType extends UnionType
{
    public function __construct()
    {
        parent::__construct([
            'types' => [
                Types::skuRow(),
                Types::sortSeparator()
            ],
            'resolveType' => static function ($value): callable {
                if (is_object($value)) {
                    return Types::skuRow();
                } else {
                    return Types::sortSeparator();
                }
            }
        ]);
    }
}
```

At any rate, I would not presume to push back on this too hard, as you know this stuff far better than me, and there's no harm in dropping the suggestion from the documentation, which I've already done.

docs/schema-definition.md
shmax: Ah, I'm sure you're right, thank you.

docs/schema-definition.md
spawnia: Have you tried exactly the query that I suggested? I think that looking for an unknown type would fail without a type loader. How do you construct your `Schema`?

shmax: I'm not sure I understand what you're getting at with this. Yes, if I run your exact query without a type loader, I get an error:

```
{
  "errors": [
    {
      "message": "Unknown type \"SomeUnknownType\".",
      "locations": [
        {
          "line": 1,
          "column": 20
        }
      ],
      "code": 0,
      "extensions": {
        "file": "/var/www/html/vendor/webonyx/graphql-php/src/Validator/Rules/KnownTypeNames.php",
        "line": 87
      }
    }
  ]
}
```

I also get an error _with_ a type loader. 
```
{
  "error": {
    "message": "Unknown graphql type: SomeUnknownType"
  }
}
```

Other than one having nicer formatting than the other, it's not clear to me what we've proven, here. What is the use case for querying for types that don't exist?

Regardless, I removed the bit you didn't like 4 days ago. What is left to do?

shmax:
@spawnia Hey there, do you have any further comments on this?

docs/schema-definition.md
shmax: Just replying to myself, here--I was experimenting with removing the type loader again, and it's definitely not advised; removing it doesn't break anything, but it _does_ lead to invocation of `TypeInfo::extractTypes`, which tours the entire tree sniffing for types and is probably not what you want whether you're using lazy loading or not.

spawnia: Ah right, that makes sense. As of now, your new example does not function as a type loader - whereas the existing example at https://webonyx.github.io/graphql-php/schema-definition/#lazy-loading-of-types does. Can we get the best of both worlds? Also, I would like to avoid overwhelming readers with different implementation variants and rather just describe a single approach that is optimal for most use cases.

shmax: Yeah, let me review how everything works again and then I'll rewrite the whole section. I would have done that in the first place but I didn't want to step on any toes. I'll work on it when I get a chance.

shmax:
@spawnia is this any better? I now downplay the bit about `fields`, merged the two code samples, and added a link to the sample that has a working version of this structure.

docs/schema-definition.md
```diff
@@ -105,63 +105,80 @@ $schema = new Schema($config);
...
+2. Define each custom type as a callable that returns a type, rather than an object instance. Then, the work of instantiating them will only happen as they are needed by each query.
+
+3. Define all of your object **fields** and \*_args_ properties as callbacks. If you're already doing #2 then this isn't needed, but it's a quick and easy precaution.
+
+It is recommended to centralize this kind of functionality in a type loader. A typical example might look like the following:
 
 ```php
-use GraphQL\Type\Definition\Type;
-use GraphQL\Type\Definition\ObjectType;
-use GraphQL\Type\Schema;
```
spawnia, line 120: Let's keep the imports.

docs/schema-definition.md
```diff
@@ -105,63 +105,80 @@ $schema = new Schema($config);
...
 
+// MyAType.php
+class MyAType extends \GraphQL\Type\Definition\ObjectType {
+    public __construct() {
+        parent::__construct([
+            "name"=>"MyA",
```
spawnia, line 124: formatting is off

docs/schema-definition.md
```diff
@@ -105,63 +105,80 @@ $schema = new Schema($config);
...
+            "fields" => fn() => [
+                'b' => Types::MyB()
```
spawnia, line 126: There is no class `Types`

docs/schema-definition.md
```diff
@@ -105,63 +105,80 @@ $schema = new Schema($config);
...
+
+    public static function MyA(): callable { return self::get(MyAType::class); }
+    public static function MyB(): callable { return self::get(MyBType::class); }
+
+    ...
 }
 
+// ...
+
 $typeRegistry = new TypeRegistry();
```
spawnia, line 171: All static now, no instantiation needed

docs/schema-definition.md
```diff
@@ -105,63 +105,80 @@ $schema = new Schema($config);
...
 
 $schema = new Schema([
-    'query' => $typeRegistry->get('Query'),
-    'typeLoader' => static fn (string $name): Type => $typeRegistry->get($name),
+    'query' => new Query(),
+    'mutation' => new Mutation(),
+    'typeLoader' => static fn (string $name): Type =>
+        $typeRegistry->get($name),
```
spawnia, line 177: function is private

docs/schema-definition.md
```diff
@@ -105,63 +105,80 @@ $schema = new Schema($config);
...
 ]);
 ```
 
-You can automate this registry if you wish to reduce boilerplate or even
-introduce a Dependency Injection Container if your types have other dependencies.
-
-Alternatively, all methods of the registry could be static - then there is no need
-to pass it in the constructor - instead use **TypeRegistry::myAType()** in your type definitions.
+A working demonstration of this kind of architecture can be found in the [01-blog](../examples/01-blog) sample.
```
spawnia, line 181: Good, because this example is not working at all 😜

docs/schema-definition.md
```diff
@@ -105,63 +105,80 @@ $schema = new Schema($config);
...
-of your object **fields** as callbacks.
+1. Use a type loader. This will put you in a position to implement your own caching and lookup strategies, and GraphQL won't need to preload a map of all known types to do its work.
 
-Type loading is very similar to PHP class loading, but keep in mind that the **typeLoader** must
-always return the same instance of a type. A good way to ensure this is to use a type registry:
+2. Define each custom type as a callable that returns a type, rather than an object instance. Then, the work of instantiating them will only happen as they are needed by each query.
+
+3. Define all of your object **fields** and \*_args_ properties as callbacks. If you're already doing #2 then this isn't needed, but it's a quick and easy precaution.
+
+It is recommended to centralize this kind of functionality in a type loader. A typical example might look like the following:
```
spawnia, line 116:
```suggestion
It is recommended to centralize this kind of functionality in a type registry. A typical example might look like the following:
```

docs/schema-definition.md
```diff
@@ -105,63 +105,80 @@ $schema = new Schema($config);
...
-By default, the schema will scan all of your type, field and argument definitions to serve GraphQL queries.
-It may cause performance overhead when there are many types in the schema.
+If your schema makes use of a large number of complex or dynamically-generated types, they can become a performance concern. There are a few best practices that can lessen their impact:
 
-In this case, it is recommended to pass the **typeLoader** option to the schema constructor and define all
-of your object **fields** as callbacks.
+1. Use a type loader. This will put you in a position to implement your own caching and lookup strategies, and GraphQL won't need to preload a map of all known types to do its work.
 
-Type loading is very similar to PHP class loading, but keep in mind that the **typeLoader** must
-always return the same instance of a type. A good way to ensure this is to use a type registry:
```
spawnia, line 115: This constraint is important and no longer mentioned

spawnia (changes requested):
The numbered points are fine, but the example is broken as of now and needs some more work.

docs/schema-definition.md
shmax: Whoops! Fixed, thank you

docs/schema-definition.md
shmax: Thank you 👍

docs/schema-definition.md
shmax: Well, to be fair I'm not sure the old one worked, either (it references "Query" on  `$typeRegistry`).

docs/schema-definition.md
shmax: fixed, I think..?

docs/schema-definition.md
shmax: Right you are, thank you 👍

shmax:
Did my best to clean up the sample a bit more. Still not sure if it works, but I'll put aside some time to actually build a working version, and then I'll make sure the markdown sample is complete and copy/paste-ready

shmax:
I rewrote the sample to use non-abstract model types (ie "Story" and "User"), and even mocked it all up as a working example (not checked-in) to make sure it all works. I guess the only last gotcha with it is that I didn't stub out the data source stuff. I can do that, or I can simplify more by going back to `MyA` and `MyB`. If nothing else, it's now conceptually just a subset of what you see in the example/blog-01 folder, which should help comprehension. Let me know what you think.

docs/schema-definition.md
shmax: Moved to a comment:
39f35486

Good enough?

docs/schema-definition.md
shmax: By the way, will this relative link work once the documentation site gets built? Should I use an absolute link?

shmax:
How are we doing on this?

spawnia (approved):

spawnia:
Did a pass over the example, looks good now. If you approve of the changes I made, we can merge.

shmax:
LGTM, thanks much 👍

shmax:
Thanks again for all the support and quick reviews.

PR #1365, opened 2023-04-17
Merge selected fields in `ResolveInfo::getFieldSelection()`

### LINKED ISSUE #1364: ResolveInfo::getFieldSelection() doesn't merge fields correctly in some specific cases
Resolving selection sets for a Relay-type connection, I've got this piece of code:
```php
  $queryStructure = $info->getFieldSelection(2);
  return array_keys($queryStructure['edges']['node'] ?? []);
```
It works correctly for most cases, but recently I've found out that there are valid queries, which result in missing fields. Namely, queries with duplicates like:
```graphql
query duplicateSelection {
  productConnection(first: 1) {
    edges {
      node { name }
      node { updatedAt }
    }
  }
}
```
In this case, I only get `updatedAt`, and the `name` is gone (overwritten). 

I've found the place to fix it, so a PR is on the way. :-)

jacor84:
# Description

Folding a selection set in `ResolveInfo` needs to preserve all fields from a given type/level. Currently, there is a possibility to overwrite it with a new set of fields. See an example below, which is a new test case, where the sub-selection from the first `level2` would be lost.

```graphql
    query deepMerge {
      level1 {
        level2 { # used here
          scalar1
        }
        level2 { # and here
          scalar2
        }
        scalar1
        scalar2
      }
    }
```

The solution is to merge selected fields for `FieldNode`, which is done in the second commit.

Closes #1364

### DIFF
--- a/src/Type/Definition/ResolveInfo.php
+++ b/src/Type/Definition/ResolveInfo.php
@@ -227,7 +227,10 @@ class ResolveInfo
         foreach ($selectionSet->selections as $selectionNode) {
             if ($selectionNode instanceof FieldNode) {
                 $fields[$selectionNode->name->value] = $descend > 0 && $selectionNode->selectionSet !== null
-                    ? $this->foldSelectionSet($selectionNode->selectionSet, $descend - 1)
+                    ? \array_merge_recursive(
+                      $fields[$selectionNode->name->value] ?? [],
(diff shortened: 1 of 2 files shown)


spawnia:
Thanks, released with https://github.com/webonyx/graphql-php/releases/tag/v15.2.5.

PR #1366, opened 2023-04-17
Adjust return type in methods doc block

henser:
Change createFromException and addDebugEntries return type in doc block from SerializableError to array

### DIFF
--- a/src/Error/FormattedError.php
+++ b/src/Error/FormattedError.php
@@ -122,7 +122,7 @@ class FormattedError
      *
      * For a list of available debug flags @see \GraphQL\Error\DebugFlag constants.
      *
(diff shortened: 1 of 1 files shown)


spawnia:
No, `SerializableError` is a type alias that describes the returned value more precisely.

henser:
@spawnia This is causing psalm to list the use of createFromException with errors on my implementation, as the method is expected to return array from it's definition (`:array`) but it's doc'd as `@return SerializableError`. Could we at least make it `@return SerializableError|mixed[]`

spawnia:
Psalm should understand type imports, I know PHPStan does. I am not going to make this libraries type definitions worse because of inadequate third-party tooling.

PR #1372, opened 2023-04-27
Call persistedQueryLoader even when query was passed

### LINKED ISSUE #1370: Always call `loadPersistedQuery` if `queryId` was provided
Hey.

`graphql-php` conveniently provides means of implementing persisted queries - namely, it parses `queryId` from the request. The usual strategy is called Apollo automatic persisted queries - find the query in cache, return if found, cache if not found. This is implemented by all major GraphQL libraries: Apollo Server, Hot Chocolate, ExpediaGroup/graphQL. 

The process is described in detail here: https://www.apollographql.com/docs/apollo-server/performance/apq/

This would have worked in `graphql-php` too, but there's a small stopper which prevents us from doing so: https://github.com/webonyx/graphql-php/blob/master/src/Server/Helper.php#L258. If `queryId` and `query` are both provided (as is the case when a query by ID wasn't initially found), then persisted query loader isn't executed, meaning there's no way to actually save the query into cache.

This can be worked around in userland by overwriting `Helper::promiseToExecuteOperation`, but that isn't an elegant solution to what seems to be just a precaution of the past. Instead, we can simply change the condition to NOT check for `query` at all, and instead always use `persistedQueryLoader` whenever a `queryId` is provided. 

Would this kind of change be accepted? If so, I'll PR the change.

oprypkhantc:
Closes #1370.

This partially reverses #790. That PR was _supposed_ to fix the support of [Automatic persisted queries](https://www.apollographql.com/docs/apollo-server/performance/apq/), but it has only partially done so. By adding the `$query === null` condition, it contradicts the flow proposed by Apollo and @cwdt in their original PR: 

```
3. If the backend doesn't recognize the hash, it will ask the client to send the hash and the query text so it can store them mapped together for future lookups. During this request, the backend will also fulfill the data request.
```

The important part here is `so it can store them mapped together for future lookups` - this is currently not possible because `persistedQueryLoader` is not called when both `queryId` and `query` were passed in a request.

Since `webonyx/graphql-php` parses Apollo-style automatic persisted query IDs (see https://github.com/webonyx/graphql-php/blob/master/src/Server/OperationParams.php#L123), it makes sense to allow them to be implemented in userland.

Although this is technically a fix of the previous PR, I've marked it as a "change" in the changelog because it might be a breaking change for some.


spawnia:
Makes sense, thanks for the fix. I am releasing this as a bug fix, as the previous logic never really worked.

PR #1375, opened 2023-05-04
Normalize phpenum type

shmax:
I'm starting to make heavier use of `PhpEnumType`, and its lack of configurability is starting to become an issue. Someone added support for a custom name recently, but I need the same support for 'description', and rather than start an optional argument chain I thought I'd see what you think about just doing the same config-based setup we do for everything else. In the spirit of brevity, I added a static `fromPath` helper that will allow that same one-liner feel (or we can ditch it). A backwards-compatible solution using a union type on the first argument may be possible, but I figure it might be best to just rip the band-aid off. Let me know what you think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+## v15.14.3
+
+### Fixed
+
+- Serialize `BackedEnum` instances by name, not value https://github.com/webonyx/graphql-php/pull/1618
+
 ## v15.14.2
 
 ### Changed
--- a/src/Type/Definition/EnumType.php
+++ b/src/Type/Definition/EnumType.php
@@ -137,7 +137,7 @@ class EnumType extends Type implements InputType, OutputType, LeafType, Nullable
         }
 
         if (is_a($value, \BackedEnum::class)) {
-            return $value->value;
+            return $value->name;
         }
 
         if (is_a($value, \UnitEnum::class)) {
--- a/tests/Type/PhpEnumType/BackedPhpEnum.php
+++ b/tests/Type/PhpEnumType/BackedPhpEnum.php
@@ -4,7 +4,7 @@ namespace GraphQL\Tests\Type\PhpEnumType;
 
 enum BackedPhpEnum: string
 {
-    case A = 'A';
+    case A = 'a';
     case B = 'B';
-    case C = 'C';
+    case C = 'the value does not really matter';
 }


tests/Type/PhpEnumType/MyCustomPhpEnum.php
```diff
@@ -6,7 +6,7 @@
 use GraphQL\Type\Definition\Description;
 
 #[Description(description: 'foo')]
-enum PhpEnum
+enum MyCustomPhpEnum
```
shmax, line 9: This doesn't have anything to do with the feature, it just always vaguely confuses me when sample code uses variable names that sound like language features or the name of the thing being demonstrated, eg:

```php
$float = 5;
```

src/Type/Definition/PhpEnumType.php
```diff
@@ -4,31 +4,35 @@
...
 use GraphQL\Utils\Utils;
 
 /**
+ * @see EnumValueDefinitionNode
  * @phpstan-import-type PartialEnumValueConfig from EnumType
+ *
+ * @phpstan-type PhpEnumTypeConfig array{
+ *   name?: string|null,
+ *   description?: string|null,
+ *   enumPath: class-string<\UnitEnum>,
```
spawnia, line 19:
```suggestion
 *   enumClass: class-string<\UnitEnum>,
```

path is something that relates to the filesystem

src/Type/Definition/PhpEnumType.php
```diff
@@ -39,22 +43,29 @@ public function __construct(string $enum, ?string $name = null)
         }
 
         parent::__construct([
-            'name' => $name ?? $this->baseName($enum),
+            'name' => $config['name'] ?? $this->baseName($config['enumPath']),
```
spawnia, line 46: Repeated array access to `$config['enumPath']`

src/Type/Definition/PhpEnumType.php
```diff
@@ -39,22 +43,29 @@ public function __construct(string $enum, ?string $name = null)
...
-
         return $value->name;
     }
 
+    /**
+    /* @param class-string $path
+     */
+    public static function fromPath($path): PhpEnumType {
+        return new static(["enumPath"=>$path]);
+    }
```
spawnia, line 67: I don't think this is necessary

tests/Type/PhpEnumType/MyCustomPhpEnum.php
spawnia: Makes sense to me.

tests/Type/PhpEnumTypeTest.php
```diff
@@ -30,11 +30,13 @@ protected function setUp(): void
 
     public function testConstructEnumTypeFromPhpEnum(): void
     {
-        $enumType = new PhpEnumType(PhpEnum::class);
+        $enumType = new PhpEnumType(
+            ['enumPath' => MyCustomPhpEnum::class]
+        );
```
spawnia, line 35:
```suggestion
        $enumType = new PhpEnumType([
            'enumPath' => MyCustomPhpEnum::class,
        ]);
```

spawnia (changes requested):

src/Type/Definition/PhpEnumType.php
shmax: Heh, yeah I wasn't crazy about it, either. 👍

shmax:
Thanks for the comments. 👍 I'm happy with this if you are.

spawnia (approved):
Looks good, I am happy to include this in the next major version. It will be a few months until then, if you really want this feature included in the library before then we can discuss ways to include it in a non-breaking way.

shmax:
Merged latest. Still no rush, or anything, but it should be ready for next release.

Issue #1376, opened 2023-05-04
PHP 8.2 compatibility issue: Using ${var} in strings is deprecated, use {$var} instead


kirkmadera:
This seems like a quick fix, so I am making an issue. This happens when making any call to the GraphQL endpoint before opcache has stored it.

> Deprecated:  Using ${var} in strings is deprecated, use {$var} instead in vendor/webonyx/graphql-php/src/Validator/Rules/ValuesOfCorrectType.php on line 183

kirkmadera:
Confirmed that my PR fixes the issue I am running into

spawnia:
See https://github.com/webonyx/graphql-php#versioning:

> Older versions are no longer supported.

PR #1377, opened 2023-05-04
Fix to deprecated string interpolation syntax

kirkmadera:
Fixes https://github.com/webonyx/graphql-php/issues/1376


spawnia:
See https://github.com/webonyx/graphql-php#versioning:

> Older versions are no longer supported.

kirkmadera:
Did not see version 15. Thanks. Will migrate

PR #1385, opened 2023-05-26
Allow context to be scoped

ruudk:
By implementing the `ScopedContext` interface on your context object, you can ensure that
scope is only passed downwards and not shared with other fields.

Example:
```graphql
query {
  a {
    b {
      c
    }
    d {
      e {
        f
      }
    }
  }
  g {
    h {
      i
    }
  }
}
```

In the above situation, the following will happen:
* `a` receives the cloned context from the executor
* `b` receives the cloned context from `a`
* `c` receives the cloned context from `b`
* `d` receives the cloned context from `a`
* `e` receives the cloned context from `d`
* `f` receives the cloned context from `e`
* `g` receives the cloned context from the executor
* `h` receives the cloned context from `g`
* `i` receives the cloned context from `h`

References:
- https://github.com/graphql/graphql-js/issues/2692
- https://github.com/rmosolgo/graphql-ruby/pull/2634

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -294,8 +294,8 @@ class ReferenceExecutor implements ExecutorImplementation
         // Similar to completeValueCatchingError.
         try {
             $result = $operation->operation === 'mutation'
-                ? $this->executeFieldsSerially($type, $rootValue, $path, $fields)
-                : $this->executeFields($type, $rootValue, $path, $fields);
+                ? $this->executeFieldsSerially($type, $rootValue, $path, $fields, $this->exeContext->contextValue)
+                : $this->executeFields($type, $rootValue, $path, $fields, $this->exeContext->contextValue);
 
             $promise = $this->getPromise($result);
             if ($promise !== null) {
@@ -522,22 +522,23 @@ class ReferenceExecutor implements ExecutorImplementation
      *
      * @param mixed             $rootValue
      * @param array<string|int> $path
+     * @param mixed             $contextValue
      *
      * @phpstan-param Fields $fields
      *
      * @return array<mixed>|Promise|\stdClass
      */
-    protected function executeFieldsSerially(ObjectType $parentType, $rootValue, array $path, \ArrayObject $fields)
+    protected function executeFieldsSerially(ObjectType $parentType, $rootValue, array $path, \ArrayObject $fields, $contextValue)
     {
         $result = $this->promiseReduce(
             \array_keys($fields->getArrayCopy()),
-            function ($results, $responseName) use ($path, $parentType, $rootValue, $fields) {
+            function ($results, $responseName) use ($contextValue, $path, $parentType, $rootValue, $fields) {
                 $fieldNodes = $fields[$responseName];
                 assert($fieldNodes instanceof \ArrayObject, 'The keys of $fields populate $responseName');
 
                 $fieldPath = $path;
                 $fieldPath[] = $responseName;
-                $result = $this->resolveField($parentType, $rootValue, $fieldNodes, $fieldPath);
+                $result = $this->resolveField($parentType, $rootValue, $fieldNodes, $fieldPath, $contextValue);
                 if ($result === static::$UNDEFINED) {
                     return $results;
                 }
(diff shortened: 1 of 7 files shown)


src/Executor/ScopedContext.php
```diff
@@ -0,0 +1,12 @@
+<?php declare(strict_types=1);
+
+namespace GraphQL\Executor;
+
+/**
+ * This interface can be used to indicate that your context should be scoped.
+ *
+ * A scoped context will only delegate information to children.
+ */
+interface ScopedContext
```
SavageTiger, line 10: Should we maybe forbid the use of the non-scoped one? Since it doens't really make sense to use it in 99% of cases I think.

SavageTiger (approved):
That was fast 🚀

src/Executor/ScopedContext.php
ruudk: This PR targets `webonyx/graphql-php`, we'll only use scoped context in our implementation.

ruudk:
@spawnia The PR is ready for review. I'd love to hear what you think. If you have suggestions on how to improve this, please let me know

src/Executor/ReferenceExecutor.php
```diff
@@ -1393,4 +1417,18 @@ protected function ensureValidRuntimeType(
...
+
+    /**
+     * @param mixed $contextValue
+     *
+     * @return mixed
+     */
+    private function maybeScopeContext($contextValue)
+    {
+        if ($contextValue instanceof ScopedContext) {
+            return clone $contextValue;
```
spawnia, line 1429:
```suggestion
            return $contextValue->clone();
```

This let's the interface be more than a marker and have more control, otherwise the implementation is also trivial.

src/Executor/ScopedContext.php
```diff
@@ -0,0 +1,12 @@
+<?php declare(strict_types=1);
+
+namespace GraphQL\Executor;
+
+/**
+ * This interface can be used to indicate that your context should be scoped.
```
spawnia, line 6: How does it indicate that? (I know, it is by implementing it in the value that is passed as the context, but that is not obvious).

src/Executor/ScopedContext.php
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * A scoped context will only delegate information to children.
+ */
+interface ScopedContext
+{
```
spawnia, line 11: I think we should add a `public function clone(): void;` here and use it to describe when exactly it will be called. That will make the mechanics of how this works much more obvious.

tests/Executor/TestClasses/MyContext.php
```diff
@@ -0,0 +1,9 @@
+<?php declare(strict_types=1);
+
+namespace GraphQL\Tests\Executor\TestClasses;
+
+class MyContext
```
spawnia, line 5:
```suggestion
final class MyContext
```

All classes in `tests`

spawnia (changes requested):
Let's figure out where this can be documented. I think at least the PHPDoc that describes `$contextValue` should be extended, perhaps an entire new section of docs can be added?

src/Executor/ScopedContext.php
ruudk: Done

src/Executor/ScopedContext.php
ruudk: Fixed

ruudk:
@spawnia Ready. Added a few lines to the documentation. I think it's sufficient.

spawnia:
Concerning docs, I would like to provide an example. The pull request description does not seem right:

> d receives the cloned context from d
> f receives the cloned context from d

Again, I think the following is still true:

>  I think at least the PHPDoc that describes $contextValue should be extended,

See https://github.com/webonyx/graphql-php/blob/01d7b25022cf9b846f0faf9b36f2b44009ff36f9/src/GraphQL.php#L47-L51. A more detailed description can be available in the interface. Also, the interface should be added to the class reference generation.

ruudk:
@spawnia  Updated the PR. If you have better wording, please let me know.

PR #1388, opened 2023-05-30
Prepare v15.5.0

spawnia:


ruudk (approved):
Much better wording. Thanks!

PR #1411, opened 2023-07-21
Allow disabling short form query in Printer

ruudk:
When using the Printer to tidy/reformat GraphQL operations, I noticed that it would always choose the short form. In our company, we'd like to be explicit, and always define the type of the operation.

Therefore I want to be able to turn this off.

### DIFF
--- a/src/Language/Printer.php
+++ b/src/Language/Printer.php
@@ -62,6 +62,8 @@ use GraphQL\Language\AST\VariableNode;
  */
 class Printer
 {
+    private bool $useShortFormQueryWhenPossible;
+
     /**
      * Converts the AST of a GraphQL node to a string.
      *
@@ -72,12 +74,15 @@ class Printer
     public static function doPrint(Node $ast): string
     {
         static $instance;
-        $instance ??= new static();
+        $instance ??= new static(true);
 
         return $instance->printAST($ast);
     }
 
-    protected function __construct() {}
+    public function __construct(bool $useShortFormQueryWhenPossible)
+    {
+        $this->useShortFormQueryWhenPossible = $useShortFormQueryWhenPossible;
+    }
 
     /**
      * Recursively traverse an AST depth-first and produce a pretty string.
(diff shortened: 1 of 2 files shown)


spawnia:
I don't see a hard technical reason to make this change. However, I do value ease of maintenance - particularly staying close to the reference implemenation `graphql-js`, which also does not have such an option - see https://graphql.org/graphql-js/language/#print. Thanks, but I am going to pass on this.

spawnia:
Perhaps https://github.com/LastDragon-ru/lara-asp/tree/master/packages/graphql-printer is the way to go for you, it seems to follow the philosophy of having everything configurable.

ruudk:
😢 I don't want to include other dependencies. 

As long as the Printer won't change to a final class, I can just extend it and override the `p` function, intercept it when `$node instanceof OperationDefinitionNode` and then make sure `^{` is replaced with `^query {`. 

To make that a bit easier, would it be possible if we break up this switch, and introduce protected methods for every condition?
https://github.com/webonyx/graphql-php/blob/63e01bbc7c327575dfa72fb30c7379c42b2ebb31/src/Language/Printer.php#L93-L99

```php
    protected function p(?Node $node, bool $isDescription = false): string
    {
        if ($node === null) {
            return '';
        }

        switch (true) {
            case $node instanceof ArgumentNode:
                return $this->printArgumentNode($node);
```

That would allow others to be flexible?

ruudk:
This works:
```php
class CustomPrinter extends Printer 
{
    protected function p(?Node $node, bool $isDescription = false) : string
    {
        if ($node instanceof OperationDefinitionNode) {
            $value = parent::p($node, $isDescription);

            if (str_starts_with($value, '{')) {
                return 'query ' . $value;
            }

            return $value;
        }

        return parent::p($node, $isDescription);
    }
}
```

spawnia:
I plan to keep this library extensible, see https://github.com/webonyx/graphql-php/blob/master/CONTRIBUTING.md.

PR #1412, opened 2023-07-21
Also set queryId from sha256 hash if query is provided

aboks:
This is necessary to support automatic persisted queries (https://www.apollographql.com/docs/apollo-server/performance/apq/) in the way outlined by #1372. Without this fix, the Apollo-style automatic query hash is not converted to a query id when a query is also present. This means the persistedQueryLoader is still not called to allow saving the query, even after the fix from #1372.


spawnia:
Thanks, released in https://github.com/webonyx/graphql-php/releases/tag/v15.5.3.

PR #1413, opened 2023-07-21
Do not set raw vars on query complexity rule if no validation rules are used

aboks:
If an empty array of validation rules is passed, no validations will be performed. It seems unnecessary to set the raw variable values on the query complexity rule in such a case.

In these situations, this change saves the unnecessary overhead from DocumentValidator::getRule(), which internally instantiates all rules through DocumentValidator::allRules().


spawnia:
Thanks.

PR #1418, opened 2023-08-02
Allow lazy root types (query, mutation, subscription)

ruudk:
This makes it possible to create a schema with the root types lazy. This way, only when the root type is needed, the type will be initialized by calling the callable.

Why would a Schema have an instance of the mutation and subscription root type, when it's only serving a query operation? Visa versa. That does not make sense to me.

This is in line with the rest of the codebase, where all the types can be passed as lazy.

```php
$schema = new Schema([
    'query' => fn () => $query,
    'mutation' => fn () => $mutation,
    'subscription' => fn () => $subscription,
]);
```

### Alternative solution

An alternative solution could be to allow passing a string with the name of the root type. When needed, the Schema will call the typeLoader with the given name. For example:

```php
new Schema([
    'query' => 'Query',
    'typeLoader' => $typeLoader,
]);
```


shmax:
Nice concept. Don't know why I didn't think of it when I was working on this stuff. 👍

src/Type/Schema.php
```diff
@@ -214,7 +214,17 @@ public function getOperationType(string $operation): ?ObjectType
...
     {
-        return $this->config->query;
+        $query = $this->config->query;
+
+        if ($query === null) {
+            return null;
+        }
+
+        if (is_callable($query)) {
+            return $this->config->query = ($query)();
```
ruudk, line 224:
```suggestion
            return $this->config->query = $query();
```

src/Type/Schema.php
```diff
@@ -224,7 +234,17 @@ public function getQueryType(): ?ObjectType
...
     {
-        return $this->config->mutation;
+        $mutation = $this->config->mutation;
+
+        if ($mutation === null) {
+            return null;
+        }
+
+        if (is_callable($mutation)) {
+            return $this->config->mutation = ($mutation)();
```
ruudk, line 244:
```suggestion
            return $this->config->mutation = $mutation();
```

src/Type/Schema.php
```diff
@@ -234,7 +254,17 @@ public function getMutationType(): ?ObjectType
...
     {
-        return $this->config->subscription;
+        $subscription = $this->config->subscription;
+
+        if ($subscription === null) {
+            return null;
+        }
+
+        if (is_callable($subscription)) {
+            return $this->config->subscription = ($subscription)();
```
ruudk, line 264:
```suggestion
            return $this->config->subscription = $subscription();
```

ruudk:
@spawnia nice, added some tweaks as comments

spawnia:
Thank you, released with https://github.com/webonyx/graphql-php/releases/tag/v15.6.0.

stayallive:
Very nice 🤘

spawnia:
> Is it possible to leverage this lazy loading using the schema definition language?

Sure, depends on the implementation.

PR #1421, opened 2023-08-06
add link to validation toolkit

shmax:
Do you mind if I pimp my validation library? I'm obviously biased, but I can't imagine life without it, and I'd love to see it get more usage.

https://github.com/shmax/graphql-php-validation-toolkit


shmax:
Thank you!

PR #1423, opened 2023-08-08
Use type names when possible

ruudk:
Instead of calling Type::builtInTypes which calls Introspection::getTypes() that creates a lot of objects, we only need the type names.

I noticed this while working on something else and I think this is a good improvement.

### DIFF
--- a/src/Type/Definition/NamedTypeImplementation.php
+++ b/src/Type/Definition/NamedTypeImplementation.php
@@ -41,10 +41,9 @@ trait NamedTypeImplementation
         throw new InvariantViolation('Must provide name for Type.');
     }
 
-    /** @throws InvariantViolation */
     public function isBuiltInType(): bool
     {
-        return \array_key_exists($this->name, Type::builtInTypes());
+        return \in_array($this->name, Type::BUILT_IN_TYPE_NAMES, true);
     }
 
     public function name(): string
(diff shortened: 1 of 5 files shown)


src/Type/Definition/NamedTypeImplementation.php
```diff
@@ -44,7 +45,7 @@ protected function inferName(): string
     /** @throws InvariantViolation */
     public function isBuiltInType(): bool
     {
-        return \array_key_exists($this->name, Type::builtInTypes());
+        return \in_array($this->name, [...Type::STANDARD_TYPE_NAMES, ...Introspection::TYPE_NAMES], true);
```
spawnia, line 48: Let's optimize further and extract the array to a `const BUILT_IN_TYPE_NAMES`, maybe in `Type`?

spawnia (approved):
Great change.

src/Type/Definition/NamedTypeImplementation.php
ruudk: On it.

ruudk:
@spawnia Although the php-cs-fixer bot is super nice, the fact that it runs as a GitHub Actions causes it to not trigger the CI workflows. It can only be done like that when you use a Personal Access Token.

ruudk:
@spawnia It's green now!

shmax:
Nice one. 👍

PR #1426, opened 2023-08-14
Introduce StandardTypeRegistry

### LINKED ISSUE #1424: Replace static variables and statically referenced types with instance based TypeRegistry
First of all, we ❤️ this amazing library. We've been using it since early 2017 and have 2 large schema's consisting over 1000+ types.

The library sometimes uses static variables to store objects that it creates. This way, the object is instantiated once, and then returned from cache. 

https://github.com/webonyx/graphql-php/blob/86d5a659445225c8cbf410ac596090e3ebb0752c/src/Utils/Utils.php#L13-L15
https://github.com/webonyx/graphql-php/blob/86d5a659445225c8cbf410ac596090e3ebb0752c/src/Language/Visitor.php#L344-L346
https://github.com/webonyx/graphql-php/blob/86d5a659445225c8cbf410ac596090e3ebb0752c/src/Language/Printer.php#L76-L77

The above examples are fine, because these objects don't hold state that can ever change. 

But it becomes a problem when this happens:
https://github.com/webonyx/graphql-php/blob/86d5a659445225c8cbf410ac596090e3ebb0752c/src/Type/Definition/Type.php#L114-L119

Here, the Introspection types (several objects, enums, and referenced built in types) and the standard types are stored statically.

Same here: 
https://github.com/webonyx/graphql-php/blob/86d5a659445225c8cbf410ac596090e3ebb0752c/src/Type/Introspection.php#L51C1-L52

And here:
https://github.com/webonyx/graphql-php/blob/86d5a659445225c8cbf410ac596090e3ebb0752c/src/Executor/ReferenceExecutor.php#L664-L667

Let's say you have multiple independent schema's. One of the schema's uses a [Type::overrideStandardTypes](https://github.com/webonyx/graphql-php/blob/86d5a659445225c8cbf410ac596090e3ebb0752c/src/Type/Definition/Type.php#L145) to replace the standard `String` 
type with a custom one. This now becomes a problem depending when the second schema loads. Because it uses statically cached types, they reference to a different `String` type, resulting in:
```
Found duplicate type in schema at Viewer.__typename: String. Ensure the type loader returns the same instance. See https://webonyx.github.io/graphql-php/type-definitions/#type-registry. in /Volumes/CS/www/cosmos/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php:878
 
 Caused by
 AssertionError: Found duplicate type in schema at Viewer.__typename: String. Ensure the type loader returns the same instance. See https://webonyx.github.io/graphql-php/type-definitions/#type-registry.
 
```

This problem pops up when running our test suite. When I run a single test, everything is fine. But when I run multiple tests together, every time the Schema loads, it gets a new instance of our standard `String` type. Even if I would also statically cache this default type, the problem with also happen when I run 2 tests that both operate on a different schema. Therefore I think the static references are not the way to go.

I tried to create a PR to fix this, but it's a bit hard for me. Some code directly references methods on [Introspection](https://github.com/search?q=repo%3Awebonyx%2Fgraphql-php%20%22Introspection%3A%3A%22&type=code) that return the cached definitions. 

What I would like to see for a next version of this library, is that instead of statically referenced types we switch to an instance based type registry. 

For example:

```php
interface TypeRegistry {
  public function (protected callable $typeLoader) {}
  public function get($typeName) : Type&NamedType {}
  public function nonNull($wrapped) : NonNullType {}
  public function listOf($wrapped) : ListOfType {}
  public function string() : ScalarType {}
  public function int() : ScalarType {}
  // ... etc
}
class DefaultTypeRegistry implements TypeRegistry {
  protected $resolvedTypes = [];  
  // ...
}

$typeRegistry = new DefaultTypeRegistry(
    typeLoader: function(string $typeName): Type&NamedType {
        return match ($typeName) {
            'Boolean' => new BooleanType(),
            'Float' => new FloatType(),
            'ID' => new IDType(),
            'Int' => new IntType(),
            'String' => new StringType(),
            'Query' => new QueryType(),
            'HelloWorld' => new HelloWorldType(),
        }
    }
);
$schema = new Schema([
    'typeRegistry' => $typeRegistry
]);
```

Instead of referencing `Type::string()` to get a StringType, you use `$typeRegistry->get('String')` that returns that instance. 

If needed, we can create specific methods for standard types like `$typeRegistry->string()`. 
The `Type::listOf` and `Type::nonNull` can also move to this registry: `$typeRegistry->listOf($typeRegistry->nonNull($typeRegistry->string()))`. 

The [Introspection](https://github.com/webonyx/graphql-php/blob/master/src/Type/Introspection.php) class becomes an instance within the Schema. The Schema can instantiate it like `$this->introspection = new Introspection($this->typeRegistry);`. 

The statically called [Directive](https://github.com/webonyx/graphql-php/blob/86d5a659445225c8cbf410ac596090e3ebb0752c/src/Type/Definition/Directive.php#L91-L93C11) also switched to an instance created inside the Schema.

Since every Schema has their own registry, there are no collisions when multiple schema's coexist. 

The `Type::overrideStandardTypes` can be removed, as this is now controlled in your type loader. It will first go to the type loader to load `String`. It can fallback to a default StringType later. 

I also think that `Type::getStandardTypes` can be removed, as it doesn't really matter if something is standard or not. Just always go through the type loader and cache it inside the instance of the type registry.

ruudk:
Currently, types are referenced and cached statically. This is problematic when using multiple schema's that have different standard types that share the same memory. For example, when running them in un-isolated unit tests or when there is a long running PHP process that serves GraphQL requests.

To solve this problem, we introduce a `StandardTypeRegistry` interface with a `DefaultTypeRegistry` implementation. People are allowed to create their own registries by implementing the interface. Every Schema should be constructed with a `typeRegistry` that is an instance of `StandardTypeRegistry`. From there on, all types are queried from the registry. The registry will be responsible for caching the types to make sure subsequent calls to the same type will return the same instance.

Internally, all static calls to the standard types (Type::string(), Type::int(), Type::float(), Type::boolean(), Type::id()) have been replaced with dynamic calls on the type registry. Also calls to the introspection objects and internal directives are using the type registry now.

As most people probably have only one schema, we keep the static methods on the Type call. These now forward to `DefaultTypeRegistry::getInstance()`. This way, we don't break existing installations.

The reason for creating a `StandardTypeRegistry` interface as opposed to just a non-final implementation is that it allows people to use composition instead of inheritance to extend the functionality of the registry. For example, in my project I'd like to have a registry that holds all types and that allows me to query any type by their name (instead of FQCN). I can then delegate the interface methods to the decorated `StandardTypeRegistry`.

Resolves #1424

### DIFF
--- a/src/Type/Definition/Type.php
+++ b/src/Type/Definition/Type.php
@@ -4,6 +4,7 @@ namespace GraphQL\Type\Definition;
 
 use GraphQL\Error\InvariantViolation;
 use GraphQL\Type\Introspection;
+use GraphQL\Type\Registry\DefaultStandardTypeRegistry;
 use GraphQL\Utils\Utils;
 
 /**
@@ -30,9 +31,6 @@ abstract class Type implements \JsonSerializable
         ...Introspection::TYPE_NAMES,
     ];
 
-    /** @var array<string, ScalarType> */
-    protected static array $standardTypes;
-
     /**
      * @api
      *
@@ -40,7 +38,7 @@ abstract class Type implements \JsonSerializable
      */
     public static function int(): ScalarType
     {
-        return static::$standardTypes[self::INT] ??= new IntType();
+        return DefaultStandardTypeRegistry::instance()->int();
     }
 
     /**
@@ -50,7 +48,7 @@ abstract class Type implements \JsonSerializable
      */
     public static function float(): ScalarType
     {
-        return static::$standardTypes[self::FLOAT] ??= new FloatType();
+        return DefaultStandardTypeRegistry::instance()->float();
     }
 
     /**
@@ -60,7 +58,7 @@ abstract class Type implements \JsonSerializable
      */
     public static function string(): ScalarType
     {
-        return static::$standardTypes[self::STRING] ??= new StringType();
+        return DefaultStandardTypeRegistry::instance()->string();
     }
 
     /**
@@ -70,7 +68,7 @@ abstract class Type implements \JsonSerializable
      */
     public static function boolean(): ScalarType
     {
-        return static::$standardTypes[self::BOOLEAN] ??= new BooleanType();
+        return DefaultStandardTypeRegistry::instance()->boolean();
     }
 
     /**
@@ -80,7 +78,7 @@ abstract class Type implements \JsonSerializable
      */
     public static function id(): ScalarType
     {
-        return static::$standardTypes[self::ID] ??= new IDType();
+        return DefaultStandardTypeRegistry::instance()->id();
     }
 
     /**
@@ -108,23 +106,6 @@ abstract class Type implements \JsonSerializable
     }
 
     /**
-     * Returns all builtin in types including base scalar and introspection types.
-     *
-     * @throws InvariantViolation
-     *
-     * @return array<string, Type&NamedType>
-     */
-    public static function builtInTypes(): array
-    {
-        static $builtInTypes;
-
-        return $builtInTypes ??= \array_merge(
-            Introspection::getTypes(),
-            self::getStandardTypes()
-        );
-    }
-
-    /**
      * Returns all builtin scalar types.
      *
      * @throws InvariantViolation
@@ -133,22 +114,20 @@ abstract class Type implements \JsonSerializable
      */
     public static function getStandardTypes(): array
     {
-        return [
-            self::INT => static::int(),
-            self::FLOAT => static::float(),
-            self::STRING => static::string(),
-            self::BOOLEAN => static::boolean(),
-            self::ID => static::id(),
-        ];
+        return DefaultStandardTypeRegistry::instance()->standardTypes();
     }
 
     /**
+     * @deprecated use a custom instance of `DefaultStandardTypeRegistry` on the `Schema` instead
+     *
      * @param array<ScalarType> $types
      *
      * @throws InvariantViolation
      */
     public static function overrideStandardTypes(array $types): void
     {
+        $standardTypes = DefaultStandardTypeRegistry::instance()->standardTypes();
+
         foreach ($types as $type) {
             // @phpstan-ignore-next-line generic type is not enforced by PHP
             if (! $type instanceof ScalarType) {
@@ -163,8 +142,16 @@ abstract class Type implements \JsonSerializable
                 throw new InvariantViolation("Expecting one of the following names for a standard type: {$standardTypeNames}; got {$notStandardTypeName}");
             }
 
-            static::$standardTypes[$type->name] = $type;
+            $standardTypes[$type->name] = $type;
         }
+
+        DefaultStandardTypeRegistry::register(new DefaultStandardTypeRegistry(
+            $standardTypes[Type::INT],
+            $standardTypes[Type::FLOAT],
+            $standardTypes[Type::STRING],
+            $standardTypes[Type::BOOLEAN],
+            $standardTypes[Type::ID],
+        ));
     }
 
     /**
--- a/src/Type/Definition/Type.php
+++ b/src/Type/Definition/Type.php
@@ -32,6 +32,8 @@ abstract class Type implements \JsonSerializable
     ];
 
     /**
+     * @deprecated use the `typeRegistry` on the `Schema` instead
+     *
      * @api
      *
      * @throws InvariantViolation
@@ -42,6 +44,8 @@ abstract class Type implements \JsonSerializable
     }
 
     /**
+     * @deprecated use the `typeRegistry` on the `Schema` instead
+     *
      * @api
      *
      * @throws InvariantViolation
@@ -52,6 +56,8 @@ abstract class Type implements \JsonSerializable
     }
 
     /**
+     * @deprecated use the `typeRegistry` on the `Schema` instead
+     *
      * @api
      *
      * @throws InvariantViolation
@@ -62,6 +68,8 @@ abstract class Type implements \JsonSerializable
     }
 
     /**
+     * @deprecated use the `typeRegistry` on the `Schema` instead
+     *
      * @api
      *
      * @throws InvariantViolation
@@ -72,6 +80,8 @@ abstract class Type implements \JsonSerializable
     }
 
     /**
+     * @deprecated use the `typeRegistry` on the `Schema` instead
+     *
      * @api
      *
      * @throws InvariantViolation
@@ -108,6 +118,8 @@ abstract class Type implements \JsonSerializable
     /**
      * Returns all builtin scalar types.
      *
+     * @deprecated use the `typeRegistry` on the `Schema` instead
+     *
      * @throws InvariantViolation
      *
      * @return array<string, ScalarType>
@@ -118,7 +130,7 @@ abstract class Type implements \JsonSerializable
     }
 
     /**
-     * @deprecated use a custom instance of `DefaultStandardTypeRegistry` on the `Schema` instead
+     * @deprecated use the `typeRegistry` on the `Schema` instead
      *
      * @param array<ScalarType> $types
      *
--- a/src/Type/Registry/DefaultStandardTypeRegistry.php
+++ b/src/Type/Registry/DefaultStandardTypeRegistry.php
@@ -74,7 +74,7 @@ class DefaultStandardTypeRegistry implements StandardTypeRegistry
      *
      * @return T|callable():T
      */
-    private function objectOrLazyInitialize($objectOrClassName)
+    protected function objectOrLazyInitialize($objectOrClassName)
     {
         return is_object($objectOrClassName) ? $objectOrClassName : fn () => new $objectOrClassName();
     }
--- a/src/Type/Registry/DefaultStandardTypeRegistry.php
+++ b/src/Type/Registry/DefaultStandardTypeRegistry.php
@@ -12,7 +12,6 @@ use GraphQL\Type\Definition\IntType;
 use GraphQL\Type\Definition\ScalarType;
 use GraphQL\Type\Definition\StringType;
 use GraphQL\Type\Definition\Type;
-use GraphQL\Type\Introspection;
 
 /**
  * A default implementation of the type registry.
@@ -40,8 +39,6 @@ class DefaultStandardTypeRegistry implements StandardTypeRegistry, BuiltInDirect
     /** @var array<string, Directive> */
     protected array $directives = [];
 
-    protected Introspection $introspection;
-
     /**
      * @param ScalarType|class-string<ScalarType> $intType
      * @param ScalarType|class-string<ScalarType> $floatType
@@ -63,8 +60,6 @@ class DefaultStandardTypeRegistry implements StandardTypeRegistry, BuiltInDirect
             Type::BOOLEAN => $this->objectOrLazyInitialize($booleanType),
             Type::ID => $this->objectOrLazyInitialize($idType),
         ];
-
-        $this->introspection = new Introspection($this);
     }
 
     /**
@@ -227,10 +222,5 @@ class DefaultStandardTypeRegistry implements StandardTypeRegistry, BuiltInDirect
                 ],
             ],
         ]);
-    }
-
-    public function introspection(): Introspection
-    {
-        return $this->introspection;
     }
 }
--- a/src/Type/Registry/DefaultStandardTypeRegistry.php
+++ b/src/Type/Registry/DefaultStandardTypeRegistry.php
@@ -22,7 +22,7 @@ use GraphQL\Type\Introspection;
  * When a type by class is requested, it will return a callable that will instantiate the type on first call.
  * On subsequent calls, it will return the same instance immediately.
  */
-class DefaultStandardTypeRegistry implements StandardTypeRegistry
+class DefaultStandardTypeRegistry implements StandardTypeRegistry, BuiltInDirectiveRegistry
 {
     private static self $instance;
 
--- a/src/Type/Registry/DefaultStandardTypeRegistry.php
+++ b/src/Type/Registry/DefaultStandardTypeRegistry.php
@@ -1,0 +1,236 @@
+<?php declare(strict_types=1);
+
+namespace GraphQL\Type\Registry;
+
+use GraphQL\Error\InvariantViolation;
+use GraphQL\Language\DirectiveLocation;
+use GraphQL\Type\Definition\BooleanType;
+use GraphQL\Type\Definition\Directive;
+use GraphQL\Type\Definition\FloatType;
+use GraphQL\Type\Definition\IDType;
+use GraphQL\Type\Definition\IntType;
+use GraphQL\Type\Definition\ScalarType;
+use GraphQL\Type\Definition\StringType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Introspection;
+
+/**
+ * A default implementation of the type registry.
+ *
+ * It will return the standard types by default, but can be initialized with custom types.
+ *
+ * When a type by class is requested, it will return a callable that will instantiate the type on first call.
+ * On subsequent calls, it will return the same instance immediately.
+ */
+class DefaultStandardTypeRegistry implements StandardTypeRegistry
+{
+    private static self $instance;
+
+    /**
+     * @var array{
+     *   Int: (ScalarType|callable(): ScalarType),
+     *   Float: (ScalarType|callable(): ScalarType),
+     *   String: (ScalarType|callable(): ScalarType),
+     *   Boolean: (ScalarType|callable(): ScalarType),
+     *   ID: (ScalarType|callable(): ScalarType),
+     * }
+     */
+    protected array $standardTypes;
+
+    /** @var array<string, Directive> */
+    protected array $directives = [];
+
+    protected Introspection $introspection;
+
+    /**
+     * @param ScalarType|class-string<ScalarType> $intType
+     * @param ScalarType|class-string<ScalarType> $floatType
+     * @param ScalarType|class-string<ScalarType> $stringType
+     * @param ScalarType|class-string<ScalarType> $booleanType
+     * @param ScalarType|class-string<ScalarType> $idType
+     */
+    public function __construct(
+        $intType = IntType::class,
+        $floatType = FloatType::class,
+        $stringType = StringType::class,
+        $booleanType = BooleanType::class,
+        $idType = IDType::class
+    ) {
+        $this->standardTypes = [
+            Type::INT => $this->objectOrLazyInitialize($intType),
+            Type::FLOAT => $this->objectOrLazyInitialize($floatType),
+            Type::STRING => $this->objectOrLazyInitialize($stringType),
+            Type::BOOLEAN => $this->objectOrLazyInitialize($booleanType),
+            Type::ID => $this->objectOrLazyInitialize($idType),
+        ];
+
+        $this->introspection = new Introspection($this);
+    }
+
+    /**
+     * @template T of object
+     *
+     * @param T|class-string<T> $objectOrClassName
+     *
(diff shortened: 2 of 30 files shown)


src/Type/Registry/DefaultTypeRegistry.php
```diff
@@ -0,0 +1,233 @@
...
+ * It will return the standard types by default, but can be initialized with custom types.
+ *
+ * When a type by class is requested, it will return a callable that will instantiate the type on first call.
+ * On subsequent calls, it will return the same instance immediately.
+ */
+class DefaultTypeRegistry implements StandardTypeRegistry
+{
+    private static self $instance;
+
+    /** @var array{
```
spawnia, line 31:
```suggestion
    /**
     *@var array{
```

src/Type/Registry/DefaultTypeRegistry.php
```diff
@@ -0,0 +1,233 @@
...
+        $stringType = StringType::class,
+        $booleanType = BooleanType::class,
+        $idType = IDType::class
+    ) {
+        $this->standardTypes = [
+            Type::INT => is_object($intType) ? $intType : fn () => new $intType(),
+            Type::FLOAT => is_object($floatType) ? $floatType : fn () => new $floatType(),
+            Type::STRING => is_object($stringType) ? $stringType : fn () => new $stringType(),
+            Type::BOOLEAN => is_object($booleanType) ? $booleanType : fn () => new $booleanType(),
+            Type::ID => is_object($idType) ? $idType : fn () => new $idType(),
```
spawnia, line 65: ternary logic repeated 5 times, how about `function objectOrNew`?

src/Type/Registry/DefaultTypeRegistry.php
```diff
@@ -0,0 +1,233 @@
...
+        ];
+
+        $this->introspection = new Introspection($this);
+    }
+
+    public static function getInstance(): self
```
spawnia, line 71:
```suggestion
    /** Returns a singleton instance of itself. */
    public static function instance(): self
```

src/Type/Registry/DefaultTypeRegistry.php
```diff
@@ -0,0 +1,233 @@
...
+    {
+        return self::$instance ??= new self();
+    }
+
+    /**
+     * Register a default instance of the type registry.
+     *
+     * When using multiple schema's, you should pass a type registry to the schema constructor instead.
```
spawnia, line 79:
```suggestion
     * When using multiple schemas, you should pass a type registry to the schema constructor instead.
```

src/Type/Registry/DefaultTypeRegistry.php
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    public function getStandardType(string $type): ScalarType
+    {
+        if (is_callable($this->standardTypes[$type])) {
+            return $this->standardTypes[$type] = ($this->standardTypes[$type])();
+        }
+
+        return $this->standardTypes[$type];
+    }
```
spawnia, line 93:
```suggestion
    public function standardType(string $name): ScalarType
    {
        $type = $this->standardTypes[$name];
        if (is_callable($type) {
            return $this->standardTypes[$name] = $type();
        }

        return $type;
    }
```

src/Type/Registry/DefaultTypeRegistry.php
```diff
@@ -0,0 +1,233 @@
...
+
+    /**
+     * Returns all builtin scalar types.
+     *
+     * @throws InvariantViolation
+     *
+     * @return array<string, ScalarType>
+     */
+    public function getStandardTypes(): array
```
spawnia, line 132:
```suggestion
    public function standardTypes(): array
```

src/Type/Registry/DefaultTypeRegistry.php
```diff
@@ -0,0 +1,233 @@
...
+    {
+        return new NonNull($type);
+    }
+
+    /**
+     * @throws InvariantViolation
+     *
+     * @return array<string, Directive>
+     */
+    public function getInternalDirectives(): array
```
spawnia, line 158:
```suggestion
    public function internalDirectives(): array
```

src/Type/Registry/DefaultTypeRegistry.php
```diff
@@ -0,0 +1,233 @@
...
+                Directive::REASON_ARGUMENT_NAME => [
+                    'type' => $this->string(),
+                    'description' => 'Explains why this element was deprecated, usually also including a suggestion for how to access supported similar data. Formatted using the Markdown syntax, as specified by [CommonMark](https://commonmark.org/).',
+                    'defaultValue' => Directive::DEFAULT_DEPRECATION_REASON,
+                ],
+            ],
+        ]);
+    }
+
+    public function getIntrospection(): Introspection
```
spawnia, line 229:
```suggestion
    public function introspection(): Introspection
```

src/Type/Registry/DefaultTypeRegistry.php
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * A default implementation of the type registry.
+ *
+ * It will return the standard types by default, but can be initialized with custom types.
+ *
+ * When a type by class is requested, it will return a callable that will instantiate the type on first call.
+ * On subsequent calls, it will return the same instance immediately.
+ */
+class DefaultTypeRegistry implements StandardTypeRegistry
```
spawnia, line 27:
```suggestion
class DefaultStandardTypeRegistry implements StandardTypeRegistry
```

A bit unwieldy, but I think it is important to differentiate this from userland type registries.

src/Type/Registry/StandardTypeRegistry.php
```diff
@@ -0,0 +1,79 @@
...
+use GraphQL\Type\Definition\NonNull;
+use GraphQL\Type\Definition\NullableType;
+use GraphQL\Type\Definition\ScalarType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Introspection;
+
+/**
+ * A registry that returns standard types.
+ */
+interface StandardTypeRegistry
```
spawnia, line 17: This interface does more than just deal with standard types. Is it feasible to split:
- `StandardTypeRegistry`
- `TypeWrapper`
- `BuiltInDirectiveRegistry`
- `IntrospectionRegistry`

Or can we find a common term? I think the term type registry is already quite loaded, since we advise users to implement one to provide a `typeLoader` function. I guess most users would not need the whole extent of what this interface does though.

src/Type/SchemaConfig.php
```diff
@@ -39,6 +40,7 @@
  *   assumeValid?: bool|null,
  *   astNode?: SchemaDefinitionNode|null,
  *   extensionASTNodes?: array<SchemaExtensionNode>|null,
+ *   typeRegistry?: StandardTypeRegistry,
```
spawnia, line 43:
```suggestion
 *   typeRegistry?: StandardTypeRegistry|null,
```

src/Utils/BuildClientSchema.php
```diff
@@ -31,7 +32,8 @@
  * @phpstan-import-type UnnamedInputObjectFieldConfig from InputObjectField
  *
  * @phpstan-type Options array{
- *   assumeValid?: bool
+ *   assumeValid?: bool,
+ *   typeRegistry?: StandardTypeRegistry
```
spawnia, line 36:
```suggestion
 *   typeRegistry?: StandardTypeRegistry,
```

spawnia (commented):
I did not get through all files just yet, this is just some immediate feedback I can give you right now.

src/Type/Registry/DefaultTypeRegistry.php
ruudk: I renamed it. If you have a better name, let me know.

src/Type/Registry/StandardTypeRegistry.php
ruudk: Sure, we can split it up. But what will this bring? There are currently only 2 places where the `typeRegistry` is needed: `BuildClientSchema` and `SchemaConfig`. If we split up the interface, we would need to require 4 different types. 

We could remove the `TypeWrapper` functionality (`listOf` and `nonNull`) and keep those on `Type` instead?

ruudk: Removed the TypeWrapper functionality.

src/Type/Definition/Type.php
```diff
@@ -133,22 +114,20 @@ public static function builtInTypes(): array
...
-            self::FLOAT => static::float(),
-            self::STRING => static::string(),
-            self::BOOLEAN => static::boolean(),
-            self::ID => static::id(),
-        ];
+        return DefaultStandardTypeRegistry::instance()->standardTypes();
     }
 
     /**
+     * @deprecated use a custom instance of `DefaultStandardTypeRegistry` on the `Schema` instead
```
spawnia, line 121: Should we deprecate methods such as `getStandardTypes()` in this class too?

src/Type/Registry/DefaultStandardTypeRegistry.php
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /**
+     * @template T of object
+     *
+     * @param T|class-string<T> $objectOrClassName
+     *
+     * @return T|callable():T
+     */
+    private function objectOrLazyInitialize($objectOrClassName)
```
spawnia, line 77: https://github.com/webonyx/graphql-php/blob/master/CONTRIBUTING.md#protected-over-private

src/Type/Registry/StandardTypeRegistry.php
spawnia: We could also keep it in a single interface - or perhaps one interface that extends multiple smaller interfaces, but I don't like the name. This interface does many more things than what its name implies.

src/Utils/SchemaExtender.php
```diff
@@ -557,14 +557,7 @@ protected function extendInterfaceType(InterfaceType $type): InterfaceType
...
     {
-        return $type instanceof NamedType
-            && (
-                $type->name === Type::STRING
-                || $type->name === Type::INT
-                || $type->name === Type::FLOAT
-                || $type->name === Type::BOOLEAN
-                || $type->name === Type::ID
-            );
+        return $type instanceof NamedType && in_array($type->name, Type::STANDARD_TYPE_NAMES, true);
```
spawnia, line 560:
```suggestion
        return $type instanceof NamedType
            && in_array($type->name, Type::STANDARD_TYPE_NAMES, true);
```

src/Validator/Rules/QuerySecurityRule.php
```diff
@@ -116,9 +115,9 @@ protected function collectFieldASTsAndDefs(
 
                     $fieldDef = null;
                     if ($parentType instanceof HasFieldsType) {
-                        $schemaMetaFieldDef = Introspection::schemaMetaFieldDef();
-                        $typeMetaFieldDef = Introspection::typeMetaFieldDef();
-                        $typeNameMetaFieldDef = Introspection::typeNameMetaFieldDef();
+                        $schemaMetaFieldDef = $context->getSchema()->typeRegistry->introspection()->schemaMetaFieldDef();
+                        $typeMetaFieldDef = $context->getSchema()->typeRegistry->introspection()->typeMetaFieldDef();
+                        $typeNameMetaFieldDef = $context->getSchema()->typeRegistry->introspection()->typeNameMetaFieldDef();
```
spawnia, line 120: Extract `$introspection`

tests/DefaultStandardTypeRegistryTest.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    private DefaultStandardTypeRegistry $registry;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->registry = new DefaultStandardTypeRegistry();
+    }
+
+    /** @throws InvariantViolation */
```
spawnia, line 24: Not necessary, see https://github.com/webonyx/graphql-php/blob/0a917058620a197530b357c46d8d5943a054a37e/phpstan.neon.dist#L50-L53

src/Type/Definition/Type.php
ruudk: Done, also deprecated the other static helpers like `Type::string`. Do you agree with that?

src/Type/Registry/StandardTypeRegistry.php
ruudk: I extracted the directives into BuiltInDirectiveRegistry

But the Introspection one is weird.... 

It's more like a IntrospectionLocator, that points you towards the Introspection that is bound to the same type registry.

Because the Introspection object itself, is a registry for `_schema()` object type + also field definitions (`schemaMetaFieldDef`, `typeMetaFieldDef`, `typeNameMetaFieldDef`).

ruudk:
I removed the Introspection from the StandardTypeRegistry interface. It's now stored separately on the Schema and passed along the schema. Still not convinced. I think the Introspection class is weird. 

💡 I have an idea: Maybe, we can solve all these problems, by creating something completely different than what this PR does:

### `NamedTypeRegistry`

This is a registry, that holds type's (or callable's to the type) by their name.

Something like this:

```

include __DIR__ . '/vendor/autoload.php';

use GraphQL\Type\Definition\BooleanType;
use GraphQL\Type\Definition\FloatType;
use GraphQL\Type\Definition\IDType;
use GraphQL\Type\Definition\IntType;
use GraphQL\Type\Definition\ObjectType;
use GraphQL\Type\Definition\ScalarType;
use GraphQL\Type\Definition\StringType;
use GraphQL\Type\Definition\Type;

class CustomIntType extends IntType {}

class NamedTypeRegistry
{
    /** @var array<string, Type|(callable(): Type) */
    protected array $types;

    /** @param array<string, Type|(callable(): Type)> $types */
    public function __construct(array $types = [])
    {
        $standardTypes = [
            Type::INT => fn () => new IntType(),
            Type::FLOAT => fn () => new FloatType(),
            Type::BOOLEAN => fn () => new BooleanType(),
            Type::STRING => fn () => new StringType(),
            Type::ID => fn () => new IDType(),
        ];

        $introspectionTypes = [
            Introspection::SCHEMA_OBJECT_NAME => [Introspection::class, '_schema'],
            Introspection::TYPE_OBJECT_NAME => [Introspection::class, '_type'],
            Introspection::DIRECTIVE_OBJECT_NAME => [Introspection::class, '_directive'],
            Introspection::FIELD_OBJECT_NAME => [Introspection::class, '_field'],
            Introspection::INPUT_VALUE_OBJECT_NAME => [Introspection::class, '_inputValue'],
            Introspection::ENUM_VALUE_OBJECT_NAME => [Introspection::class, '_enumValue'],
            Introspection::TYPE_KIND_ENUM_NAME => [Introspection::class, '_typeKind'],
            Introspection::DIRECTIVE_LOCATION_ENUM_NAME => [Introspection::class, '_directiveLocation'],
        ];

        $this->types = array_merge(
            $standardTypes,
            $introspectionTypes,
            $types,
        );
    }

    /** @return Type|(callable(): Type) */
    public function byName(string $name)
    {
        if (! isset($this->types[$name])) {
            throw new \Exception(sprintf('Type "%s" not found', $name));
        }

        $type = $this->types[$name];
        if (is_callable($type)) {
            $this->types[$name] = $type = $type($this);
        }

        return $type;
    }

    /** @throws InvariantViolation */
    public function int(): ScalarType
    {
        return $this->byName(Type::INT);
    }

    /** @throws InvariantViolation */
    public function float(): ScalarType
    {
        return $this->byName(Type::FLOAT);
    }

    /** @throws InvariantViolation */
    public function string(): ScalarType
    {
        return $this->byName(Type::STRING);
    }

    /** @throws InvariantViolation */
    public function boolean(): ScalarType
    {
        return $this->byName(Type::BOOLEAN);
    }

    /** @throws InvariantViolation */
    public function id(): ScalarType
    {
        return $this->byName(Type::ID);
    }
}

$mutationType = fn (NamedTypeRegistry $registry) => new ObjectType([
    'name' => 'Mutation',
    'fields' => [
        'sum' => [
            'type' => $registry->int(),
            'args' => [
                'x' => ['type' => $registry->int()],
                'y' => ['type' => $registry->int()],
            ],
            'resolve' => static fn ($calc, array $args): int => $args['x'] + $args['y'],
        ],
    ],
]);

$registry = new NamedTypeRegistry([
    'Int' => fn () => new CustomIntType(),
    'Query' => fn (NamedTypeRegistry $registry) => new ObjectType([
        'name' => 'Query',
        'fields' => [
            'echo' => [
                'type' => $registry->string(),
                'args' => [
                    'message' => ['type' => $registry->string()],
                ],
                'resolve' => static fn ($rootValue, array $args): string => $rootValue['prefix'] . $args['message'],
            ],
        ],
    ]),
    'Mutation' => $mutationType
]);

assert($registry->string() === $registry->string());
assert(get_class($registry->string()) === StringType::class);

assert($registry->int() === $registry->int());
assert(get_class($registry->int()) === CustomIntType::class);

$introspectionTypeObject = $registry->byName(Introspection::TYPE_OBJECT_NAME);
assert($registry->string() === $introspectionTypeObject->getField('name')->getType());

assert($registry->string() === $registry->byName('Query')->getField('echo')->getType());
assert($registry->int() === $registry->byName('Mutation')->getField('sum')->getType());

echo 'All good.' . PHP_EOL;

```

The NamedTypeRegistry is an object, that holds all the (default) types of the schema. It can also be used to store additional types for the developer. But not required.

It initializes with default scalars, and introspection types. 

The constructor can override any of the default scalar types and introspection types (not sure why one would want that but ok). 

The Introspection types are lazy, and get a reference to the NamedTypeRegistry. Upon use, they will request the named scalar types. In fact, every lazy type, gets the NamedTypeRegistry as first argument to the callback. This solves the problem of "how to get the type registry".

This means we can remove the Type::overrideStandardTypes completely. The `Type::string()` methods etc can also be removed, as people should reference on the NamedTypeRegistry. 

For the field definitions that are currently stored on the Introspection type, we do the same. We make them static, and require a NamedTypeRegistry as argument. Then, the Schema queries them once, but caches them inside the Schema. 

We rename Introspection class to IntrospectionTypeBuilder.

To construct a Schema, one now needs to always pass a NamedTypeRegistry instance. 

The example above, is a working example.

src/Type/Registry/StandardTypeRegistry.php
spawnia: Looks fine now.

ruudk:
@spawnia what do you think about https://github.com/webonyx/graphql-php/pull/1426#issuecomment-1678562672

spawnia:
> @spawnia what do you think about [#1426 (comment)](https://github.com/webonyx/graphql-php/pull/1426#issuecomment-1678562672)

Sounds interesting, perhaps you can illustrate that approach in a second pull request and allow us to compare?

ruudk:
@spawnia Sounds good. Will work on it!

compwright:
> 💡 I have an idea: Maybe, we can solve all these problems, by creating something completely different than what this PR does:
> 
> ### `NamedTypeRegistry`
> This is a registry, that holds type's (or callable's to the type) by their name.

I could really use something like this, what's the status @ruudk?

TomHAnderson:
I'm commenting from an implementation perspective and not an internal one.  Why isn't this a container?  Am I just viewing it too shallow?

This is where I store my types and were there a standard type registry I suppose I would implement against it:
https://github.com/API-Skeletons/doctrine-orm-graphql/blob/10.2.x/src/Type/TypeContainer.php

This is nearly bare-bones PSR-11 and is "standard type" and not introspective.  I use many containers and allow overriding of their registry.  That way my application can use a shared type manager if I create two drivers.

spawnia:
I am trying a different approach in https://github.com/webonyx/graphql-php/pull/1868.

PR #1431, opened 2023-08-24
docs: remove DateTime scalar mention from GraphQL Utils

simPod:
It has been removed


spawnia:
Your fork seems outdated.

PR #1434, opened 2023-08-26
Added ability to control field visibility on schema definition

### LINKED ISSUE #782: Hide parts of the schema
So this already came up some times before: #649 & #405. But it never got anywhere

So I'm here to see if there is interest in getting this in even or that it's out of scope.

I'm really impressed by how GraphQL-Ruby has implemented this.

> With GraphQL-Ruby, it’s possible to hide parts of your schema from some users. This isn’t exactly part of the GraphQL spec, but it’s roughly within the bounds of the spec.

They list some great arguments about why this is useful:

> Here are some reasons you might want to hide parts of your schema:
>
> - You don’t want non-admin users to know about administration functions of the schema.
> - You’re developing a new feature and want to make a gradual release to only a few users first.

This is in part a security through obscurity functionality BUT it's also a really functional feature to be able to have feature switches and test out new parts of the schema in production without it being part of the public schema immediately.

That last part (feature flagging, incremental rollout) is what interests me most about this feature (see for example how [GitHub is using it](https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions) to hide their discussions API behind a feature flag.

---

I have been browsing around and I think a possible (from the user perspective) implementation could look something like this:

```diff
$userType = new ObjectType([
    'name' => 'User',
    'description' => 'Our blog visitor',
    'fields' => [
        'firstName' => [
            'type' => Type::string(),
            'description' => 'User first name'
        ],
        'email' => Type::string(),
        'internalId' => [
            'type' => Type::string(),
            'description' => 'Example field that could be hidden'
+           'visible' => function ($context): bool { /.../ },
        ],
    ],
+   'visible' => function ($context): bool { /.../ },
]);
```

GraphQL-Ruby implements the following rules for the `visible` callback result:

> These methods are called with the query context, based on the hash you pass as context:. If the method returns false, then that member of the schema will be treated as though it doesn’t exist for the entirety of the query. That is:
>
> - In introspection, the member will not be included in the result
> - In normal queries, if a query references that member, it will return a validation error, since that member doesn’t exist

These rules would as far as I'm concerned also apply to graphql-php since they seem reasonable and go a little further than simply hiding them from the introspection, they basically remove them from the schema entirely so they also cannot be called.

---

For a bit more detail check out the excellent GraphQL-Ruby docs page: https://graphql-ruby.org/authorization/visibility.

I'm mainly here to see IF this is even something that would be considered for inclusion in graphql-php so please give me some feedback on the idea (which is a shameless copy of the GraphQL-Ruby implementation and not a original idea) so I know if I should be moving forward or that it won't be accepted at all 😄

carlagouveia:
This PR aims to add the ability to control field visibility on the schema definition, using either a bool or a callable that has access to the context. This is not final in any way and I am open to other ideas on how to handle this.

Fixes #782


src/Type/Definition/FieldDefinition.php
```diff
@@ -181,6 +192,15 @@ public function getType(): Type
         return $this->type ??= Schema::resolveType($this->config['type']);
     }
 
+    public function isVisible(mixed $context): bool
+    {
+        if (is_callable($this->visible)) {
+            return call_user_func($this->visible, $context);
+        }
+
+        return true;
```
spawnia, line 201:
```suggestion
        return is_callable($this->visible)
            ? ($this->visible)($context)
            : $this->visible;
```

tests/Executor/ExecutorSchemaTest.php
```diff
@@ -30,6 +30,12 @@ public function testExecutesUsingASchema(): void
                 'url' => ['type' => Type::string()],
                 'width' => ['type' => Type::int()],
                 'height' => ['type' => Type::int()],
+                'mimetype' => [
+                    'type' => Type::string(),
+                    'visible' => function ($context): bool {
+                        return false;
+                    },
```
spawnia, line 37:
```suggestion
                    'visible' => fn ($context): bool => false,
```

tests/Executor/ExecutorSchemaTest.php
spawnia, line 1: Lacking test cases for:
- `false`
- `fn ($context): bool => true`

tests/Type/IntrospectionTest.php
```diff
@@ -1725,4 +1725,50 @@ public function testExecutesAnIntrospectionQueryWithoutCallingGlobalFieldResolve
         GraphQL::executeQuery($schema, $source, null, null, null, null, $fieldResolver);
         self::assertEmpty($calledForFields);
     }
+
+    /** @see it('does not expose invisible fields') */
```
spawnia, line 1729: Is this functionality copied from graphql-js? If not, remove this.

tests/Type/IntrospectionTest.php
```diff
@@ -1725,4 +1725,50 @@ public function testExecutesAnIntrospectionQueryWithoutCallingGlobalFieldResolve
...
+    public function testDoesNotExposeInvisibleFields(): void
+    {
+        $TestType = new ObjectType([
+            'name' => 'TestType',
+            'fields' => [
+                'nonVisible' => [
+                    'type' => Type::string(),
+                    'visible' => function ($context): bool {
+                        return false;
+                    },
```
spawnia, line 1739: Use a data provider and also test for `false`.

tests/Type/IntrospectionTest.php
```diff
@@ -1725,4 +1725,50 @@ public function testExecutesAnIntrospectionQueryWithoutCallingGlobalFieldResolve
...
+                    'fields' => [
+                        [
+                            'name' => 'visible',
+                        ],
+                    ],
+                ],
+            ],
+        ];
+
+        self::assertEquals($expected, GraphQL::executeQuery($schema, $request)->toArray());
```
spawnia, line 1772:
```suggestion
        self::assertSame($expected, GraphQL::executeQuery($schema, $request)->toArray());
```

src/Executor/ReferenceExecutor.php
```diff
@@ -601,6 +601,10 @@ protected function resolveField(ObjectType $parentType, $rootValue, \ArrayObject
             return static::$UNDEFINED;
         }
 
+        if (! $fieldDef->isVisible($contextValue)) {
+            return static::$UNDEFINED;
```
spawnia, line 605: So for clients, the field would simply not show up in the results? I am pretty sure this behaviour goes against the GraphQL specification. If anything, I would expect a validation error.

spawnia (changes requested):

src/Executor/ReferenceExecutor.php
carlagouveia: I wasn't really sure how to proceed here since the spec is not clear on this case. If we prefer validation instead, where would be the best place to add a new test case?

spawnia: Probably next to the test cases that ensure that an error is thrown when a client requests entirely unknown fields. I would expect hidden fields to behave exactly the same as a field that is defined at all, at least from a client perspective.

carlagouveia:
@spawnia just added the field visibility validation. Let me know if that is what you had in mind. 👍

spawnia:
I made some modifications to further align how a non-visible and a non-existent field would behave.

Due to the different call sites, we don't always have the same `$context` available. I think it would be of very limited use to provide whatever context is available, as the presented information can differ and the implementation of `visibility` would have to account for that. If you think something in particular would be useful or necessary to have available, we can discuss that further.

Please document this feature at https://github.com/webonyx/graphql-php/blob/master/docs/type-definitions/object-types.md#field-configuration-options. The note for `visible` should provide a short outline why this would be preferrable over conditional inclusion or exclusion of a field, and how exactly it behaves.

carlagouveia:
> I made some modifications to further align how a non-visible and a non-existent field would behave.
> 
> Due to the different call sites, we don't always have the same `$context` available. I think it would be of very limited use to provide whatever context is available, as the presented information can differ and the implementation of `visibility` would have to account for that. If you think something in particular would be useful or necessary to have available, we can discuss that further.
> 
> Please document this feature at https://github.com/webonyx/graphql-php/blob/master/docs/type-definitions/object-types.md#field-configuration-options. The note for `visible` should provide a short outline why this would be preferrable over conditional inclusion or exclusion of a field, and how exactly it behaves.

@spawnia thanks for all the modifications. I've updated the documentation, adding information about this feature. Here is the PR [#1456](https://github.com/webonyx/graphql-php/pull/1456).

carlagouveia:
> You can push the docs update directly to this pull request.

Done @spawnia.

docs/type-definitions/object-types.md
```diff
@@ -58,14 +58,15 @@ This example uses **inline** style for Object Type definitions, but you can also
...
+| resolveField | `callable`             | **function ($value, array $args, $context, [ResolveInfo](../class-reference.md#graphqltypedefinitionresolveinfo) $info): mixed**<br> Given the **$value** of this type, it is expected to return value for a field defined in **$info->fieldName**. A good place to define a type-specific strategy for field resolution. See section on [Data Fetching](../data-fetching.md) for details. |
+| visible      | `bool` or `callable`   | **function ($context): mixed**<br> You can customize the visibility of fields by defining them as visible or not visible using a `bool` or your own custom logic by sending a `callable` that receives the `$context`. In introspection, the field will not be included in the result and if a query references that field, it will return a validation error.                             |
```
spawnia, line 69: > that receives the `$context`

As mentioned in https://github.com/webonyx/graphql-php/pull/1434#issuecomment-1735062027, that is no longer the case.

```suggestion
| visible      | `bool` or `callable`   | Defaults to `true`. The given callable receives no arguments and is expected to return a `bool`, it is called once  when the field may be accessed. The field is treated as if it were not defined at all when this is `false`. |
```

PR #1439, opened 2023-09-04
chore(deps): lock cs fixer so it does not randomly break CI

simPod:


simPod:
https://github.com/webonyx/graphql-php/actions/runs/6074829878/job/16479717559#step:5:1

simPod:
Though not sure it will help since it's run on master only. At least it can be detected more easily since the first failed commit would be the fixer upgrade from Renovate.

spawnia:
Makes sense, thank you.

PR #1469, opened 2023-10-12
Fix PHP code block highlighting

shmax:
I just spent several days trying to get php highlighting working in mkdocs for another of my projects, and finally figured it out. Thought I'd share the wealth.


spawnia:
Very nice, thank you

Issue #1204, opened 2022-07-20
Ignore Introspection queries in QueryComplexity rule


ruudk:
Followup from https://github.com/webonyx/graphql-php/discussions/1203

The `QueryComplexity` rule should ignore introspection query. This way you can set the max query complexity very low and still allow introspections to happen.

Introspection has a complexity of around 181.

kinimodmeyer:
In regards to https://github.com/webonyx/graphql-php/issues/1207 i guess i would also like to calculate the introspection queries as they can produce a significant load.

Maybe we can implement this feature here in that why, that you have the option enable/disable the exclude.

Thanks!

PR #1475, opened 2023-11-01
Allow access to calculated query complexity

ruudk:
**Early return when complexity is disabled**
This allows other validation rules to depend on the `QueryComplexity` rule, and disable the query complexity completely on some condition.

```php
$rules[QueryComplexity::class] = $queryComplexity = new QueryComplexity($this->maxQueryComplexity);

// Always enable the query complexity rule, but ignore for introspection queries
$rules[WhenIntrospection::class] = new WhenIntrospection(function () use ($queryComplexity) : void {
    $queryComplexity->setMaxQueryComplexity(QueryComplexity::DISABLED);
});
```

**Allow access to calculated query complexity**
This allows you to read the calculated query complexity. A use case could be: add a rule that logs when queries are detected with a complexity higher than X.

This helps with #1204

```php
$rules[QueryComplexity::class] = $queryComplexity = new QueryComplexity($this->maxQueryComplexity);

$rules[WhenLeaveOperation::class] = new WhenLeaveOperation(function () use ($queryComplexity) : void {
    if ($queryComplexity->getQueryComplexity() >= 400) {
        $this->logger->info('Found a query with high complexity');
    }
});
```

### DIFF
--- a/src/Validator/Rules/QueryComplexity.php
+++ b/src/Validator/Rules/QueryComplexity.php
@@ -75,6 +75,10 @@ class QueryComplexity extends QuerySecurityRule
                             return;
                         }
 
+                        if ($this->maxQueryComplexity === self::DISABLED) {
+                            return;
+                        }
+
                         $complexity = $this->fieldComplexity($operationDefinition->selectionSet);
 
                         if ($complexity <= $this->maxQueryComplexity) {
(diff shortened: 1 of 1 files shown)


ruudk:
@spawnia this one is ready for review, whenever you have time.

PR #1481, opened 2023-11-18
Don't try parse value as string if it's already enum unit. PhpEnumType

danbka33:
Fixes a bug in subscriptions when an enum object is received instead of a string

I added a couple of tests to check how the value is parsed.

In addition, I fix a test that fails depending on the PHP version (`tests/Type/EnumTypeTest.php@testAllowsSimpleArrayAsValues`). It might be worth using a check that does not include things that may change depending on the php version.

All test passed. Tested with PHP 8.2.10 (cli) (built: Aug 31 2023 18:52:55) (NTS)

изображение

If we talk about the backward compatibility of these changes, then nothing should break.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
+- Fix PhpEnumType parse value if it's already enum value https://github.com/webonyx/graphql-php/pull/1481
+
 ## v15.8.0
 
 ### Added
--- a/src/Type/Definition/PhpEnumType.php
+++ b/src/Type/Definition/PhpEnumType.php
@@ -45,6 +45,15 @@ class PhpEnumType extends EnumType
         ]);
     }
 
+    public function parseValue($value)
+    {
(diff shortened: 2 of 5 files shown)


shmax:
Not sure I totally follow what you're up to, here, but if it makes any difference there's a PR pending that refactors the `PHPEnumType` constructor to take a config object instead of a string:

https://github.com/webonyx/graphql-php/pull/1375

danbka33:
I'm sure there won't be any problems with this.  I saw this PR.  Only one method from the parent class is added here.  I’m sure that the merger with your PR will take place without conflicts, but even if there are conflicts, it can be solved.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
+- Fix PhpEnumType parse value if it's already enum value https://github.com/webonyx/graphql-php/pull/1481
```
spawnia, line 14:
```suggestion
- Handle enum values in `PhpEnumType::parseValue()` https://github.com/webonyx/graphql-php/pull/1481
```

Issue #1478, opened 2023-11-11
Defining a fragment before the query results in an error


TheGlenn88:
Hi,

I've been implementing the Type Registry, following the docs:
https://webonyx.github.io/graphql-php/schema-definition/#lazy-loading-of-types

and using the example
https://github.com/webonyx/graphql-php/tree/master/examples/01-blog

What i've found is that if you define a fragment before the query like this:
```
fragment Name on User {
  firstName
}

{
  user(id: 1) {
    ...Name
  }
}
```

it throws an exception: `Exception: Unknown graphql type: User`

if you write the query the other way around

```
{
  user(id: 1) {
    ...Name
  }
}

fragment Name on User {
  firstName
}
```

It returns the user record correctly.

I've been having this problem in my own implementation and spent a long time trying to fix my code and accidentally found that removing the fragment removed the error, so I realised it was some sort of issue with the fragment.

Is this a bug of graphql-php or is this a problem with the Type Registry implementation?

For reference, I ran the same query against an express-graphql implementation and it doesn't matter which way around the fragment is defined.

Thanks
Mike

spawnia:
@shmax I believe the issue here is that the implementation of `Types::byTypeName()` introduced in https://github.com/webonyx/graphql-php/pull/1363 is insufficient. That function is used as `Schema.typeLoader`, it needs to be able to locate and instantiate types that were not yet loaded.

PR #1488, opened 2023-11-29
simplify byTypeName

shmax:
Fix for #1478 

The TypeLoader included with the documentation is sort of an odd melding of my own type loader, and the original one that it replaced, so I included "byTypeName" without really considering how it works. In this change, I replace it with the somewhat simpler implementation taken straight from my own production code.

Disclaimer: for the life of me, I can't seem to figure out how to hook up the debugger when running the example servers, so while I did test this fix in the context of the blog-01 sample, the fix here is somewhat faith-based.

### DIFF
--- a/docs/schema-definition.md
+++ b/docs/schema-definition.md
@@ -169,15 +169,13 @@ final class Types
         return static fn () => self::byClassName($classname);
     }
 
-    public static function byTypeName(string $typeName): Type&NamedType
+    public static function byTypeName(string $shortName): Type&NamedType
     {
-        return match ($typeName) {
-            'Boolean' => self::boolean(),
-            'Float' => self::float(),
-            'ID' => self::id(),
-            'Int' => self::int(),
-            default => self::$types[$typeName] ?? throw new \Exception("Unknown GraphQL type: {$typeName}."),
-        }
+        $method = lcfirst($shortName);
+        if(method_exists(__CLASS__, $shortName))
+            return self::{$method}()();
+
+        throw new \Exception("Unknown graphql type: {$shortName}");
     }
 
     private static function byClassName(string $classname): Type
--- a/examples/01-blog/Blog/Types.php
+++ b/examples/01-blog/Blog/Types.php
@@ -118,23 +118,9 @@ final class Types
      */
     public static function byTypeName(string $shortName): Type
     {
-        $cacheName = \strtolower($shortName);
-
-        if (isset(self::$types[$cacheName])) {
-            return self::$types[$cacheName];
-        }
-
-        $method = \lcfirst($shortName);
-        switch ($method) {
-            case 'boolean':
-                return self::boolean();
-            case 'float':
-                return self::float();
-            case 'id':
-                return self::id();
-            case 'int':
-                return self::int();
-        }
+        $method = lcfirst($shortName);
+        if(method_exists(__CLASS__, $method))
+            return self::{$method}()();
 
         throw new \Exception("Unknown graphql type: {$shortName}");
     }
--- a/examples/01-blog/Blog/Types.php
+++ b/examples/01-blog/Blog/Types.php
@@ -119,8 +119,9 @@ final class Types
     public static function byTypeName(string $shortName): Type
     {
         $method = lcfirst($shortName);
-        if(method_exists(__CLASS__, $method))
+        if (method_exists(__CLASS__, $method)) {
             return self::{$method}()();
+        }
 
         throw new \Exception("Unknown graphql type: {$shortName}");
     }


spawnia:
I think this example is incomplete and will not work properly for built-in scalars, also it does unnecessary work when `byTypeName()` is called repeatedly.

shmax:
Why do you say it won't work for built-in scalars?

I didn't try them all, but I tested `int` and `id` (in the bloc-01 sample) and didn't notice any problems.

And what unnecessary work are you talking about?

Please elaborate.

spawnia:
Ah, I guess it will in fact work because method names are case insensitive in PHP.

The unnecessary work would be in case of a repeated call to `byTypeName`. Instead of checking if the type is already loaded and returning it right away, your implementation will:
- build an array
- check if it is callable (which I believe to be somewhat expensive in PHP, since callables are so broad)
- call it (using `call_user_func`, which is unnecessary and slow compared to just calling the expression using `()`)
- create a lazy callback that does the instantiation by class name
- call said callback, which will in turn do string manipulation on the class name to get the type name
- use the calculated type name to check if it is already instantiated, then return that

I think the extra check is cheap enough and the likelyhood of having to load a type multiple times is high enough to warrant it.

shmax:
Did you confirm any of this in the debugger? I'm having trouble getting the debugger to work with the samples, but I stepped through the flow in my own codebase, and I'm not able to confirm many of your points:

> build an array

Where? What array? Are you talking about what happens in `Type::getStandardTypes`? That doesn't execute in the context of `findByName`.

> check if it is callable (which I believe to be somewhat expensive in PHP, since callables are so broad)

Okay, fair point. I've changed it to an assert (which have 0 cost when configured for production), and moved to your preferred syntax. 807147d99fe40daa24d3a3fdf7ac8aca843782ea

> create a lazy callback that does the instantiation by class name

Yes, this is how the lazy load system works.

> call said callback, which will in turn do string manipulation on the class name to get the type name
Well, yes, but only once. Once the lazy load is fulfilled, it doesn't repeat. And this is something that the implementer can adjust to taste. If the string manipulations are a real concern, then you can tailor your naming convention so that it's not needed. We can do that in the sample if you like.

So, to conclude:
* the built-in types don't come through `findByName`
* the lazy-loading is working as expected. If there is a fundamental problem with it we should figure it out, but the goal of this PR is to get the sample working (which I think it does, and with less code than we had before) and anything more profound seems a bit out of scope.

If you still think there is some kind of performance issue here let me know and I'll fire up the profiler and check this solution against the one in your own open PR.

spawnia:
> What array?

`[__CLASS__, lcfirst($shortName)]` is an array.

> I've changed it to an assert (which has 0 cost when configured for production)

Clients are able to control the type names passed to the type loader, simply by including whatever type name they want in a fragment or a variable. There is no guarantee the passed type names are correct, so unless we want this to crash with a cryptic error of `method not found`, a runtime check is preferrable.

> Well, yes, but only once. Once the lazy load is fulfilled, it doesn't repeat.

I am sure it does, given the control flow. It goes from type name to method name (`lcfirst`), then from method name to class name (since the methods reference the classes via `::class`), then back from class name to type name (`preg_replace` and stuff). Only then does the cache come into play (in your example, I access it immediately given the type name).

Changing the naming convention might not work if type names are equivalent to PHP keywords, so a suffix of `Type` is added. I will document this.

> the built-in types don't come through findByName

Ah yes,`Schema` shortcuts them. Still, I would rather have the method work for all types.

shmax:
I've been using basically this same code for years in production. It's been battle-tested with every possible use of fragments and weird edge case you can imagine. And since it does only activate for particular edge cases (eg fragment usage), the string manipulations you are fixated on don't even register on the profiler. And even if they did, I'm sure it could be mitigated with a little collaboration.

As a lead on this project, you have to develop something called "bedside manner". When someone takes time away from their busy schedule to contribute something that doesn't even affect their own usage, you should make an effort to be supportive. Instead of trying to see how fast you can find something to nitpick and hand-wave away the entire effort, make suggestions. Ask questions. If you're concerned about performance, then turn it into a discussion, not a chance to showboat, etc.

In short, lighten the fuck up. Be nice. Say "thank you" every now and then. Add the occasional smiley. You would be amazed how much it helps.

spawnia:
Forgive me I did not believe *it just works™*, given the previous example was broken. That is why I was poking holes into this solution. We seem to disagree on the optimal solution and approach to sample code, that is fine. I appreciate your contributions to this project anyway, which is why I asked you for feedback on https://github.com/webonyx/graphql-php/pull/1485.

PR #1491, opened 2023-12-01
Correctly print deprecated argument with default value

ruudk:
Today I deprecated an argument with a default value and noticed that the schema was printed wrong:

```graphql
type Query {
  singleField(id: ID @deprecated(reason: "this is deprecated") = 123): Int
}
```

The `@deprecated` directive should be put after the `= 123` default value.


ruudk:
@spawnia Ready for review, whenever you have time 💙

spawnia:
Good catch, also seems like a very natural progression to have this combination.

PR #1492, opened 2023-12-05
Allow callable in `args`

ruudk:
This makes it possible to dynamically alter the `args` via a callable.

### DIFF
--- a/src/Type/Definition/Argument.php
+++ b/src/Type/Definition/Argument.php
@@ -60,13 +60,17 @@ class Argument
     }
 
     /**
-     * @phpstan-param ArgumentListConfig $config
+     * @phpstan-param ArgumentListConfig|callable(): ArgumentListConfig $config
      *
      * @return array<int, self>
      */
-    public static function listFromConfig(iterable $config): array
+    public static function listFromConfig($config): array
     {
         $list = [];
+
+        if (\is_callable($config)) {
+            $config = $config();
+        }
 
         foreach ($config as $name => $argConfig) {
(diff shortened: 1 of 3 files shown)


ruudk:
@spawnia Ready for review, whenever you have time 💙

spawnia:
> This makes it possible to dynamically alter the `args` via a callable.

I don't understand the benefit for that, please elaborate.

On the performance side, I only see this hurting. Given that `Argument::listFromConfig()` is called eagerly, there is no bonus for lazyness to be had here.

ruudk:
I agree that there is no performance benefit of doing this lazily. 

But it allows for using `yield` to return arguments that are depending on conditions. 

Example:

```php
fn() {
  yield 'id' => ['type' => Type::int()];
  if ($this->someFeatureFlag()) {
    yield 'other' => ['type' => Type::int()];
  }
  if ($this->someSpecialFlag()) {
    yield 'special' => ['type' => Type::int()];
  }
}
```

Obviously, we can also generate this array before constructing the type, or we can use a method that returns the flat array.
But everywhere we allow for callable that can yield results, except here.

spawnia:
> But everywhere we allow for callable that can yield results, except here.

Disagree, we only do that when it brings performance benefits. Other config options don't allow it, e.g. `name`.

I think it is somewhat misleading to offer this option since in other cases lazyness helps performance. In this case, it hurts performance - whether it is used or not. Thank you for the suggestion, but I am going to pass on this.

PR #1506, opened 2024-01-05
ci: ensure GA run is triggered on fixers' push

simPod:
- We have to use PAT since events triggered by the GITHUB_TOKEN will not create a new workflow run
 
 https://github.com/stefanzweifel/git-auto-commit-action/blob/12f68633e45c72459cd040c868605f2471c7f63b/README.md#commits-made-by-this-action-do-not-trigger-new-workflow-runs  
 https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow


simPod:
I asked @vladar in DM for creating scoped PAT.

spawnia:
Thanks for digging in!

PR #1514, opened 2024-01-19
ci(prettier): remove dead settings

simPod:
This settings has no effect.

Resolves warning here https://github.com/webonyx/graphql-php/actions/runs/7530844894


spawnia:
Included in https://github.com/webonyx/graphql-php/pull/1506.

PR #1526, opened 2024-02-15
Added https://github.com/API-Skeletons/doctrine-orm-graphql

TomHAnderson:


spawnia:
You may propose changes to the docs via merge requests against the `master` branch.
Do not edit the generated HTML files in the `gh-pages` branch directly, they are automatically generated.

PR #1533, opened 2024-03-04
feat: use concrete exceptions instead of RequestError

simPod:
I propose implementing more detailed errors instead of generic RequestError. It allows consumers to better react to specific error cases.


spawnia:
Thanks, released with https://github.com/webonyx/graphql-php/releases/tag/v15.10.0.

PR #1536, opened 2024-03-07
fix: build client schema have custom scalar value mixed type

vuongxuongminh:
Currently, the serialize function of scalar type made by `GraphQL\Utils\BuildClientSchema` class will cast value to string but it should return raw value instead for consistent with `GraphQL\Utils\BuildSchema` and in GraphQL spec not have any notes said value of scalar types must be string https://graphql.org/learn/schema/#scalar-types

This PR to solve the problem above

### DIFF
--- a/tests/Utils/BuildClientSchemaTest.php
+++ b/tests/Utils/BuildClientSchemaTest.php
@@ -1013,4 +1013,28 @@ SDL;
         $this->expectExceptionMessage('Expected Foo to be a GraphQL Object type.');
         $clientSchema->assertValid();
     }
+
+    public function testCustomScalarValueMixedExecution(): void
+    {
+        $schema = BuildSchema::build('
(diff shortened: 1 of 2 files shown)


tests/Utils/BuildClientSchemaTest.php
```diff
@@ -1013,4 +1013,28 @@ public function testRecursiveUnion(): void
         $this->expectExceptionMessage('Expected Foo to be a GraphQL Object type.');
         $clientSchema->assertValid();
     }
+
+    public function testCustomScalarValueMixedExecution(): void
```
spawnia, line 1017: You are only testing `string` now which would have worked before, can you add a test that passes an `int`?

vuongxuongminh: Updated

PR #1538, opened 2024-03-08
ensure new lines are counted in multi line comments, fixes #1177

### LINKED ISSUE #1177: Lexer wrong location when multiline comments
When using the lexer with a schema that has a multiline comments the location is wrong.

Taken the following schema
```
"""
testing
"""
scalar something
```
It ends up saying that the name token `scalar` is on line 2.

thekonz:
See https://github.com/webonyx/graphql-php/issues/1177

Basically, the Lexer ignored any line breaks inside of multi line comments. This was pretty annoying when generating code coverage for graphql sdl files.
The test shows it pretty well. If you revert the change in the Lexer, you will get `2` as the line number for the single line comment instead of `5`.

### DIFF
--- a/tests/Language/SchemaParserTest.php
+++ b/tests/Language/SchemaParserTest.php
@@ -199,6 +199,15 @@ type Hello {
             'loc' => $loc(0, 85),
         ];
         self::assertEquals($expected, $doc->toArray());
+
+        // ensure the lexer does not treat multi line comments as one line
+        $tokenAfterMultiLineComment = $doc->loc?->startToken?->next?->next;
+        self::assertEquals('Even with comments between them', trim($tokenAfterMultiLineComment?->value ?? ''));
+        self::assertEquals(5, $tokenAfterMultiLineComment?->line);
+
+        $typeToken = $tokenAfterMultiLineComment?->next;
+        self::assertEquals('type', $typeToken?->value);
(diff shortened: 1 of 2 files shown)


tests/Language/SchemaParserTest.php
```diff
@@ -199,6 +199,15 @@ public function testParsesTypeWithDescriptionMultiLineString(): void
...
         self::assertEquals($expected, $doc->toArray());
+
+        // ensure the lexer does not treat multi line comments as one line
+        $tokenAfterMultiLineComment = $doc->loc?->startToken?->next?->next;
+        self::assertEquals('Even with comments between them', trim($tokenAfterMultiLineComment?->value ?? ''));
+        self::assertEquals(5, $tokenAfterMultiLineComment?->line);
+
+        $typeToken = $tokenAfterMultiLineComment?->next;
+        self::assertEquals('type', $typeToken?->value);
+        self::assertEquals(6, $typeToken?->line);
```
spawnia, line 210:
```suggestion
        self::assertSame('Even with comments between them', trim($tokenAfterMultiLineComment?->value ?? ''));
        self::assertSame(5, $tokenAfterMultiLineComment?->line);

        $typeToken = $tokenAfterMultiLineComment?->next;
        self::assertSame('type', $typeToken?->value);
        self::assertSame(6, $typeToken?->line);
```

PR #1548, opened 2024-03-24
Add unaliased path to resolve info

### LINKED ISSUE #1345: Add real path to `ResolveInfo`
I'd like to be able to get the real (trusted) path in `ResolveInfo` next to the untrusted aliased path.

Similar to what was written in https://github.com/webonyx/graphql-php/issues/412.

```graphql
query {
  myAlias: actor {
    myOtherAlias: movies
  }
}
```

`ResolveInfo->path` will contain `myAlias.myOtherAlias` instead of `actor.movies`.

Why is this useful? It can be used to quickly log the path of a requested field. For example, when somebody does not have access. By logging the real path, you never have to guess what it was. Also, it allows trusting the path because it cannot be spoofed.

What would be the best way to implement this? I tried adding an additional `safePath` to ResolveInfo, but that requires many changes in ReferenceExecutor. 

Another way could be to introduce a `PathAlias` value object that holds the alias + field name. For example:
```php
protected function executeFields(ObjectType $parentType, $rootValue, array $path, \ArrayObject $fields)
{
    // ...
    foreach ($fields as $responseName => $fieldNodes) {
            $fieldPath = $path;
            if ($responseName !== $fieldNodes[0]->name->value) {
                $fieldPath[] = new PathAlias($fieldNodes[0]->name->value,  $responseName);
            } else {
                $fieldPath[] = $responseName;
            }
```

So `ResolveInfo->path` could become `[PathAlias('myAlias', 'actor'), PathAlias('myOtherAlias', 'movie'), 'some', 'other', 'field']`.

We could put this behind a feature flag to not break existing projects that rely on the `ResolveInfo->path`.

Another simpler way could be to do this:
```php
protected function executeFields(ObjectType $parentType, $rootValue, array $path, \ArrayObject $fields)
{
    // ...
    foreach ($fields as $responseName => $fieldNodes) {
            $fieldPath = $path;
            if ($responseName !== $fieldNodes[0]->name->value) {
                $fieldPath[] = sprintf('%s: %s', $responseName, $fieldNodes[0]->name->value);
            } else {
                $fieldPath[] = $responseName;
            }
```

This would produce `['myAlias: actor', 'myOtherAlias: movies']`.

What do you think? I'd be happy to work on a PR to get this improved.

ruudk:
Fixes #1345

### DIFF
--- a/docs/class-reference.md
+++ b/docs/class-reference.md
@@ -308,7 +308,7 @@ Passed as 4th argument to every field resolver. See [docs on field resolving (da
 
 @phpstan-import-type QueryPlanOptions from QueryPlan
 
-@phpstan-type Path array<int, string|int>
+@phpstan-type Path list<string|int>
 
 ### GraphQL\Type\Definition\ResolveInfo Props
 
@@ -355,7 +355,7 @@ public $parentType;
  *
  * @api
  *
- * @var array<int, string|int>
+ * @var list<string|int>
  *
  * @phpstan-var Path
  */
@@ -1743,7 +1743,7 @@ function getLocations(): array
  * Returns an array describing the path from the root value to the field which produced this error.
  * Only included for execution errors. When fields are aliased, the path includes aliases.
  *
- * @return array<int, int|string>|null
+ * @return list<int|string>|null
  *
  * @api
  */
--- a/docs/class-reference.md
+++ b/docs/class-reference.md
@@ -309,7 +309,6 @@ Passed as 4th argument to every field resolver. See [docs on field resolving (da
 @phpstan-import-type QueryPlanOptions from QueryPlan
 
 @phpstan-type Path array<int, string|int>
-@phpstan-type UnaliasedPath list<string|int>
 
 ### GraphQL\Type\Definition\ResolveInfo Props
 
@@ -369,7 +368,7 @@ public $path;
  *
  * @var list<string|int>
  *
- * @phpstan-var UnaliasedPath
+ * @phpstan-var Path
  */
 public $unaliasedPath;
 
--- a/docs/class-reference.md
+++ b/docs/class-reference.md
@@ -309,6 +309,7 @@ Passed as 4th argument to every field resolver. See [docs on field resolving (da
 @phpstan-import-type QueryPlanOptions from QueryPlan
 
 @phpstan-type Path array<int, string|int>
+@phpstan-type UnaliasedPath list<string|int>
 
 ### GraphQL\Type\Definition\ResolveInfo Props
 
@@ -351,7 +352,7 @@ public $fieldNodes;
 public $parentType;
 
 /**
- * Path to this field from the very root value.
+ * Path to this field from the very root value. When fields are aliased, the path includes aliases.
  *
  * @api
  *
@@ -360,6 +361,17 @@ public $parentType;
  * @phpstan-var Path
  */
 public $path;
+
+/**
+ * Path to this field from the very root value. This will never include aliases.
+ *
+ * @api
+ *
+ * @var list<string|int>
+ *
+ * @phpstan-var UnaliasedPath
+ */
+public $unaliasedPath;
 
 /**
  * Instance of a schema used for execution.
@@ -1730,13 +1742,25 @@ function getLocations(): array
 ```php
 /**
  * Returns an array describing the path from the root value to the field which produced this error.
- * Only included for execution errors.
+ * Only included for execution errors. When fields are aliased, the path includes aliases.
  *
  * @return array<int, int|string>|null
  *
  * @api
  */
 function getPath(): ?array
+```
+
+```php
+/**
+ * Returns an array describing the path from the root value to the field which produced this error.
+ * Only included for execution errors. This will never include aliases.
+ *
+ * @return list<int|string>|null
+ *
+ * @api
+ */
+function getUnaliasedPath(): ?array
 ```
 
 ## GraphQL\Error\Warning
--- a/src/Error/Error.php
+++ b/src/Error/Error.php
@@ -36,7 +36,7 @@ class Error extends \Exception implements \JsonSerializable, ClientAware, Provid
      * corresponds to this error. Only included for errors during execution.
      * When fields are aliased, the path includes aliases.
      *
-     * @var array<int, int|string>|null
+     * @var list<int|string>|null
      */
     public ?array $path;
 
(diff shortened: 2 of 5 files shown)


src/Error/Error.php
```diff
@@ -75,7 +86,8 @@ public function __construct(
         ?array $positions = null,
         ?array $path = null,
         ?\Throwable $previous = null,
-        ?array $extensions = null
+        ?array $extensions = null,
+        ?array $unaliasedPath = null
```
ruudk, line 90: I had to put this at the end to break BC.

src/Executor/ReferenceExecutor.php
```diff
@@ -1307,14 +1345,24 @@ protected function collectSubFields(ObjectType $returnType, \ArrayObject $fieldN
...
+
             $fieldPath = $path;
             $fieldPath[] = $responseName;
-            $result = $this->resolveField($parentType, $rootValue, $fieldNodes, $fieldPath, $this->maybeScopeContext($contextValue));
+
+            $fieldNode = $fieldNodes[0];
+            assert($fieldNode instanceof FieldNode, '$fieldNodes is non-empty');
+
+            $fieldUnaliasedPath = $unaliasedPath;
+            $fieldUnaliasedPath[] = $fieldNode->name->value;
```
ruudk, line 1363: Same as `executeFieldsSerially`

src/Executor/ReferenceExecutor.php
```diff
@@ -576,11 +585,13 @@ function ($results, $responseName) use ($contextValue, $path, $parentType, $root
...
-     * @param array<int, string|int>      $path
-     * @param mixed                       $contextValue
+     * @param mixed                  $rootValue
+     * @param array<int, string|int> $path
+     * @param array<int, string|int> $unaliasedPath
+     * @param mixed                  $contextValue
      *
-     * @phpstan-param Path                $path
+     * @phpstan-param Path           $path
+     * @phpstan-param Path           $unaliasedPath
```
ruudk, line 594: It would be great if phpdoc_align was set to left, to not cause all these changes.

https://cs.symfony.com/doc/rules/phpdoc/phpdoc_align.html

src/Error/Error.php
```diff
@@ -262,6 +278,19 @@ public function getPath(): ?array
         return $this->path;
     }
 
+    /**
+     * Returns an array describing the path from the root value to the field which produced this error.
+     * Only included for execution errors. This will never include aliases.
+     *
+     * @return array<int, int|string>|null
```
ruudk, line 285: I kept it `array<int, int|string>` but personally I would make this a `list<int|string>`.

ruudk:
@spawnia This one is ready for review, when you have time.

ruudk:
@spawnia You are probably busy. Do you have time for a review? Or is there anybody else that can do it?

src/Executor/ReferenceExecutor.php
```diff
@@ -520,25 +521,33 @@ protected function doesFragmentConditionMatch(Node $fragment, ObjectType $type):
     /**
      * Implements the "Evaluating selection sets" section of the spec for "write" mode.
      *
-     * @param mixed             $rootValue
+     * @param mixed            $rootValue
      * @param array<string|int> $path
-     * @param mixed             $contextValue
+     * @param array<string|int> $unaliasedPath
+     * @param mixed            $contextValue
```
spawnia, line 527: Let's unalign them.

src/Executor/ReferenceExecutor.php
spawnia: Can you create a separate pull request for that?

src/Error/Error.php
spawnia: Let's change it to `list`, good idea!

ruudk: Changed it!

src/Executor/ReferenceExecutor.php
ruudk: Undid any unrelated changes.

ruudk:
@spawnia this is ready now. The errors are not related to this PR.

docs/class-reference.md
```diff
@@ -309,6 +309,7 @@ Passed as 4th argument to every field resolver. See [docs on field resolving (da
 @phpstan-import-type QueryPlanOptions from QueryPlan
 
 @phpstan-type Path array<int, string|int>
+@phpstan-type UnaliasedPath list<string|int>
```
spawnia, line 312: Let's change both!

spawnia:
Please merge `master`, I just fixed the PHPStan issues.

docs/class-reference.md
ruudk: Done!

ruudk:
It's green now.

spawnia (approved):

simPod (commented):
Somewhere in this change is performance issue 🤔 . v15.12.0 causes my app queries

> Error: Maximum execution time of 60 seconds exceeded

while 15.11.2 is fine

spawnia:
> Somewhere in this change is performance issue 🤔 . v15.12.0 causes my app queries
> 
> > Error: Maximum execution time of 60 seconds exceeded
> 
> while 15.11.2 is fine

Do they succeed after a while or do they loop forever?

simPod:
Seems like loop forever. I've waited 5m and nothing got finished. While on 15.11 it's done in ~10s (larger query).

ruudk:
Interesting. We run a very large GraphQL schema in production with a lot of traffic and did not notice anything different.

simPod:
My other app works fine. Also I cannot reproduce locally. I synced php patch versions and still the issue is only in prod env, not local one. I'm trying to investigate further.

simPod:
I have compared the two versions, you can see where the latest one is slower

https://blackfire.io/profiles/compare/07b25a5c-c984-49fb-a134-5b4026020104/graph

ruudk:
What kind of query did you use? Is this a big query that is using a big list or is it nested deep?

While looking at the diff in `completeListValue` I fail to see why this might be slower.

src/Executor/ReferenceExecutor.php
```diff
@@ -1307,14 +1345,24 @@ protected function collectSubFields(ObjectType $returnType, \ArrayObject $fieldN
      *
      * @return Promise|\stdClass|array<mixed>
      */
-    protected function executeFields(ObjectType $parentType, $rootValue, array $path, \ArrayObject $fields, $contextValue)
+    protected function executeFields(ObjectType $parentType, $rootValue, array $path, array $unaliasedPath, \ArrayObject $fields, $contextValue)
     {
         $containsPromise = false;
         $results = [];
         foreach ($fields as $responseName => $fieldNodes) {
+            $fieldNodes = $fields[$responseName];
```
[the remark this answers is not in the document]
spawnia: @simPod Can you check if https://github.com/webonyx/graphql-php/releases/tag/v15.12.1 fixes your issue?

simPod:
> Interesting. We run a very large GraphQL schema in production with a lot of traffic and did not notice anything different.

@ruudk the issue appears when you return a lot of entries. For each entry it computes the path and now unaliasedPath. That causes the slowdown. The query depth is not relevant. If you're returning small datasets you won't notice the slowdown.

ruudk:
I think it would be great to have a unit test that reproduces the issue. Then we can work on making this faster. Could you create it ?

PR #1549, opened 2024-03-24
Fix scoping of context for mutations

ruudk:
In #1385 we introduced the `ScopedContext` interface. When implemented on the Context, it would clone the context down to it's children.

This works great for queries. But mutations are executed serially. I forgot to call `maybeScopeContext` there.

This fixes the problem, and adds a test to guard it.it would clone the context down to it's children.

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -538,7 +538,7 @@ class ReferenceExecutor implements ExecutorImplementation
 
                 $fieldPath = $path;
                 $fieldPath[] = $responseName;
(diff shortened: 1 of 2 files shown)


ruudk:
@spawnia This one is ready for review, when you have time.

ruudk:
@spawnia You are probably busy. Do you have time for a review? Or is there anybody else that can do it?

spawnia (approved):

spawnia:
Thanks, released with https://github.com/webonyx/graphql-php/releases/tag/v15.11.2.

PR #1570, opened 2024-06-11
Remove redundant lines in ReferenceExecutor

ruudk:
Somehow this was added in #1548 but I fail to see why.

Let's remove it.

This might be the cause for the performance degradation reported in #1548.

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -1350,9 +1350,6 @@ class ReferenceExecutor implements ExecutorImplementation
         $containsPromise = false;
         $results = [];
         foreach ($fields as $responseName => $fieldNodes) {
-            $fieldNodes = $fields[$responseName];
-            assert($fieldNodes instanceof \ArrayObject, 'The keys of $fields populate $responseName');
-
             $fieldPath = $path;
             $fieldPath[] = $responseName;
 


spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.12.1, thank you.

simPod:
I just deployed it but my issue persists 🤔 I'll run the profiling again when I'm at the PC.

ruudk:
Would be great if you could create a reproducer.

simPod:
So far I was able to fetch perf comparison https://blackfire.io/profiles/compare/9748d2b2-a120-4bb8-9010-95985981e9d8/graph

I still can't reproduce. The issue appeared on another project in production env but locally it does not appear.

ruudk:
Could you provide more details about the schema, the query, and the query complexity.

simPod:
When I remove these lines it's suddenly fast 🤔 

https://github.com/simPod/graphql-php/blob/d8ea59bb02af10fd58a1192f7a781ae8aedac2d6/src/Executor/ReferenceExecutor.php#L1356-L1360

with

image

and without those lines

image

simPod:
I'm pretty confident this line is responsible https://github.com/simPod/graphql-php/blob/d8ea59bb02af10fd58a1192f7a781ae8aedac2d6/src/Executor/ReferenceExecutor.php#L1360

removing it restores the previous speed.

Removing even this line has the same performance impact https://github.com/simPod/graphql-php/blob/d8ea59bb02af10fd58a1192f7a781ae8aedac2d6/src/Executor/ReferenceExecutor.php#L1354

though both are required for correct functionality...

PR #1572, opened 2024-06-12
perf(reference-executor: optimize a little, dedupe code

simPod:
- make some lines a bit shorter

Still did not find a way to avoid performance slowdown introduced in https://github.com/webonyx/graphql-php/pull/1548. This is just an outcome of few hours I looked at the code.


spawnia (approved):
Looks good to me, at least it should make it easier to fix any issues.

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.12.2

PR #1576, opened 2024-06-17
test: use `addToAssertionCount()` instead of `assertTrue(true)`

simPod:
Replace with `addToAssertionCount()`

### DIFF
--- a/tests/TestCaseBase.php
+++ b/tests/TestCaseBase.php
@@ -15,10 +15,9 @@ abstract class TestCaseBase extends TestCase
      *
      * @see TestCase::expectNotToPerformAssertions()
      */
-    public static function assertDidNotCrash(): void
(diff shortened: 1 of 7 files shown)


tests/TestCaseBase.php
```diff
@@ -8,19 +8,6 @@
...
 abstract class TestCaseBase extends TestCase
 {
-    /**
-     * Useful to test code with no observable behaviour other than not crashing.
-     *
-     * In contrast to PHPUnit's native method, this lets the test case count towards coverage.
-     *
-     * @see TestCase::expectNotToPerformAssertions()
-     */
-    public static function assertDidNotCrash(): void
```
spawnia, line 18: I like having this named method better than the many repeated calls to `self::addToAssertionCount(1);`. We can replace the implementation, but I would like to keep the semantics.

PR #1577, opened 2024-06-17
fix(error): avoid phpunit crash by triggering side-effect via count on testcase

simPod:
@mfn this is a followup of https://github.com/sebastianbergmann/phpunit/issues/5866#issuecomment-2172429263

What do you think about the fix? I don't see any other option since we analyze the whole stacktrace.

### DIFF
--- a/src/Error/FormattedError.php
+++ b/src/Error/FormattedError.php
@@ -7,6 +7,7 @@ use GraphQL\Language\Source;
 use GraphQL\Language\SourceLocation;
 use GraphQL\Type\Definition\Type;
 use GraphQL\Utils\Utils;
+use PHPUnit\Framework\Test;
 
 /**
  * This class is used for [default error formatting](error-handling.md).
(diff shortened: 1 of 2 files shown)


src/Error/FormattedError.php
```diff
@@ -299,7 +300,11 @@ public static function printVar($var): string
         }
 
         if (\is_object($var)) {
-            return 'instance of ' . \get_class($var) . ($var instanceof \Countable ? '(' . \count($var) . ')' : '');
+            $count = ! $var instanceof Test && $var instanceof \Countable
```
spawnia, line 303:
```suggestion
            // Calling `count` on PHPUnit\Framework\Test triggers an unintended side effect - see https://github.com/sebastianbergmann/phpunit/issues/5866#issuecomment-2172429263
            $count = ! $var instanceof Test && $var instanceof \Countable
```

src/Error/FormattedError.php
```diff
@@ -299,7 +300,12 @@ public static function printVar($var): string
         }
 
         if (\is_object($var)) {
-            return 'instance of ' . \get_class($var) . ($var instanceof \Countable ? '(' . \count($var) . ')' : '');
+            // Calling `count` on instances of `PHPUnit\Framework\Test` triggers an unintended side effect - see https://github.com/sebastianbergmann/phpunit/issues/5866#issuecomment-2172429263
+            $count = ! $var instanceof Test && $var instanceof \Countable
```
spawnia, line 304: How does this behave when PHPUnit is not installed?

simPod: I've tested it and it seems to be just k https://3v4l.org/FFKFf

spawnia:
Thank you @simPod and @mfn! Released with https://github.com/webonyx/graphql-php/releases/tag/v15.12.3.

PR #1579, opened 2024-06-19
Ensure unaliasedPath does not grow for each list item

spawnia:
Fixes the performance issue @simPod reported in https://github.com/webonyx/graphql-php/pull/1548.

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -1012,13 +1012,13 @@ class ReferenceExecutor implements ExecutorImplementation
         $containsPromise = false;
         $completedItems = [];
         foreach ($results as $item) {
-            $fieldPath = [...$path, $i];
-            $info->path = $fieldPath;
-            $unaliasedPath = [...$unaliasedPath, $i];
-            $info->unaliasedPath = $unaliasedPath;
+            $itemPath = [...$path, $i];
+            $info->path = $itemPath;
+            $itemUnaliasedPath = [...$unaliasedPath, $i];
+            $info->unaliasedPath = $itemUnaliasedPath;
             ++$i;
 
-            $completedItem = $this->completeValueCatchingError($itemType, $fieldNodes, $info, $fieldPath, $unaliasedPath, $item, $contextValue);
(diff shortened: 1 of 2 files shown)


spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.12.4.

src/Executor/ReferenceExecutor.php
```diff
@@ -1012,13 +1012,13 @@ protected function completeListValue(
         $containsPromise = false;
         $completedItems = [];
         foreach ($results as $item) {
-            $fieldPath = [...$path, $i];
-            $info->path = $fieldPath;
-            $unaliasedPath = [...$unaliasedPath, $i];
```
spawnia, line 1017: Because this references itself (unlike `$fieldPath`, which always uses `$path`), this caused the `$unaliasedPath` to grow with each list item.

```php
['items', 0]
['items', 0, 1]
['items', 0, 1, 2]
...
```

This resulted in quite enormous memory usage and a big performance hit. One of our large lists previously used about 175MB of memory, with the faulty version it used over 1300MB.

@ruudk This was not only a performance issue but also logically incorrect, the `$unaliasedPath` in lists was not right. Please add a test that covers this - it should fail with the previous code and work with the current code.

ruudk: Interesting! I was not expecting this at all... And it also makes me wonder: Why did PHPStan not complain about this either?

I'll create a test for it! Thanks for fixing it!

spawnia: I would not expect PHPStan to complain about this. It could be totally fine to do something like this, depending on the use case. For example, if the goal was to keep track of all visited indices of the traversed array, it could make sense.

ruudk: https://github.com/webonyx/graphql-php/pull/1580

simPod:
_Sorry I still did not get to the issue reproducer, I'm on workation (meh word) so have limited time and so far got time only to few other PRs here._

spawnia:
No worries @simPod, your report was a great help for me to find that this issue was the root cause of the increased memory usage we saw.

PR #1581, opened 2024-06-22
Avoid infinite recursion in `QueryDepth` validator

k0ka:
The `QueryDepth` validator goes into infinite recursion on the following query:
```GraphQL
query MyQuery { 
  human { ...F1 }
}

fragment F1 {
  ...F1
}
```

I added a simple protection against recursive fragments. However, we might just break depth calculation when it reaches the max query depth.

### DIFF
--- a/src/Validator/Rules/QueryDepth.php
+++ b/src/Validator/Rules/QueryDepth.php
@@ -15,6 +15,9 @@ use GraphQL\Validator\QueryValidationContext;
 
 class QueryDepth extends QuerySecurityRule
 {
+    /** @var array<string, bool> hash of fragment names which are already calculated in recursion */
+    protected array $calculatedFragments = [];
+
     protected int $maxQueryDepth;
 
     /** @throws \InvalidArgumentException */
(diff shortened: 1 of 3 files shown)


src/Validator/Rules/QueryDepth.php
```diff
@@ -15,6 +15,9 @@
 
 class QueryDepth extends QuerySecurityRule
 {
+    /** @var array<string, bool> hash of fragment names which are already calculated in recursion */
```
spawnia, line 18:
```suggestion
    /** @var array<string, bool> Fragment names which are already calculated in recursion */
```

I guess you are referring to hash as in the datastructure HashMap, but there is no such thing in PHP. I assumed to find a usage of a hashing function somewhere in the code, but this is just a plain associative array.

spawnia (changes requested):

src/Validator/Rules/QueryDepth.php
k0ka: Yep, I meant "[Hash table](https://en.wikipedia.org/wiki/Hash_table)", not to confuse with just `array<string>`

spawnia:
Thank you @k0ka, released with https://github.com/webonyx/graphql-php/releases/tag/v15.12.5.

PR #1587, opened 2024-07-21
Create field args mapper and cache args resolution

Warxcell:
Small enhancement/performance optimization + new feature.

We already have [input parse value](https://webonyx.github.io/graphql-php/type-definitions/inputs/#converting-input-object-array-to-value-object) which turns array into something user defined, but same thing for field args' is missing. 

And because this could be something expensive (for example - user might validate if args are within defined boundaries) - it doesn't make sense to call that mapper (or even the code that collects the args) more than once for same input. (For example in sub fields of iterable - args won't change in between each iterable).

Did brief test on performance optimization - it was about 15ms on ~800 calls - not something big, but I believe its good to have it - it will be beneficial even to users that doesn't use that args mapper.


src/Type/Definition/FieldDefinition.php
```diff
@@ -56,6 +59,15 @@ class FieldDefinition
     /** @var array<int, Argument> */
     public array $args;
 
+    /**
+     * Callback for resolving field value given parent value.
```
spawnia, line 63: Inaccurate

tests/Executor/ExecutorTest.php
```diff
@@ -330,6 +330,79 @@ public function testCorrectlyThreadsArguments(): void
...
+                                    ],
+                                ],
+                            ])
+                        ),
+                        'resolve' => static fn (): array => [new \stdClass(), new \stdClass(), new \stdClass()],
+                    ],
+                ],
+            ]),
+        ]);
+        $result = Executor::execute($schema, $docAst, null, null, [], 'Example');
```
spawnia, line 400: Why use a named query?

tests/Executor/ResolveTest.php
```diff
@@ -33,7 +33,7 @@ public function testDefaultFunctionAccessesProperties(): void
     }
 
     /**
-     * @param UnnamedFieldDefinitionConfig $testField
+     * @phpstan-param UnnamedFieldDefinitionConfig $testField
```
spawnia, line 36: Why?

Warxcell: No idea, it was failing for no reason. + For consistency, everywhere else its used like `phpstan-param` not `param.

tests/Executor/ExecutorTest.php
Warxcell: Does it matter?

src/Type/Definition/FieldDefinition.php
Warxcell: Yeah, sorry. I just copied resolver fn logic and forgot to change descriptions. :) Fixed.

tests/Executor/ExecutorTest.php
spawnia: The test should be as simple as possible.

Warxcell: Ok. Simplified.

Warxcell:
Any idea why phpstan could be falling? I can't run it locally to test, because for some reason it requires php8.2 and I have php8.3 locally.

spawnia:
> Any idea why phpstan could be falling? I can't run it locally to test, because for some reason it requires php8.2 and I have php8.3 locally.

Can you not see the output in https://github.com/webonyx/graphql-php/actions/runs/10038021756/job/27739070572?pr=1587?

> Error: Method GraphQL\Tests\Executor\ResolveTest::buildSchema() has parameter $testField with no value type specified in iterable type array<string, mixed>.
> Error: PHPDoc tag @param for parameter $testField contains unresolvable type.

GitHub also inlines the errors for me:

image

Warxcell:
Locally:

```
 php -dmemory_limit=-1 ./vendor/bin/phpstan
Note: Using configuration file /home/bozhidar_hristov/projects/graphql-php/phpstan.neon.dist.
 414/414 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


                                                                                                                        
 [OK] No errors                                                                                                         
                                                                                                                        
```

:shrug:

Warxcell:
Sorry to bother you, can we merge this? :)

Warxcell:
Ping

spawnia:
@Warxcell Please merge the latest changes from `master`, let's see if tests and static analysis pass.

Warxcell:
Done. No idea why phpstan is failing:
Locally I have this error (which again, no idea why happens)


```
------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/Utils/SchemaExtender.php                                                                                                                                                                                                                                             
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  424    Method GraphQL\Utils\SchemaExtender::extendImplementedInterfaces() should return array<int, GraphQL\Type\Definition\InterfaceType> but returns array<int, GraphQL\Type\Definition\InterfaceType|(GraphQL\Type\Definition\NamedType&T of GraphQL\Type\Definition\Type)>.  
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 


                                                                                                                        
 [ERROR] Found 1 error                                                                                                  
                                               
```

src/Type/Definition/FieldDefinition.php
```diff
@@ -14,6 +15,7 @@
  * @phpstan-import-type FieldResolver from Executor
  * @phpstan-import-type ArgumentListConfig from Argument
  *
+ * @phpstan-type ArgsMapper callable(array<string, mixed>, FieldDefinition, FieldNode): mixed
```
spawnia, line 18: I believe this is the source of all PHPStan errors, since it causes a circular dependency which PHPStan seems to be unable to resolve. Thus, all code that uses a type import from `FieldDefinition` also fails to analyse.

spawnia: See https://github.com/phpstan/phpstan/issues/4637

spawnia: image

spawnia: Maybe we will just have to use a type `mixed` somewhere and kind of handwave the issue.

Warxcell: I already tried: @phpstan-type ArgsMapper callable(array<string, mixed>, mixed, mixed): mixed

but same.

  56     Circular definition detected in type alias FieldDefinitionConfig.                                                                                        
  56     Circular definition detected in type alias UnnamedFieldDefinitionConfig.    
  
  Very weird. Will research.

Warxcell: Moved it to executor. Maybe its better anyway - similar to ResolveFn.

Warxcell:
Making draft, as I might seen a bug.

src/Type/Definition/FieldDefinition.php
```diff
@@ -5,15 +5,18 @@
...
 use GraphQL\Utils\Utils;
 
 /**
  * @see Executor
  *
  * @phpstan-import-type FieldResolver from Executor
+ * @phpstan-import-type ArgsMapper from Executor
  * @phpstan-import-type ArgumentListConfig from Argument
  *
+ *
```
spawnia, line 19:
```suggestion
 *
```

spawnia:
Please add docs at https://webonyx.github.io/graphql-php/type-definitions/object-types/#field-configuration-options

Warxcell:
I'm thinking something - maybe its better to use both `$fieldDef` and `$fieldNode` as cache key. So if

```
field {
   field2 <- if this is from interface, and it returns two versions of that interface - it would be called twice.
} 
```

src/Executor/ReferenceExecutor.php
```diff
@@ -721,18 +733,22 @@ protected function resolveFieldValueOrError(
...
         $rootValue,
         ResolveInfo $info,
         $contextValue
     ) {
         try {
             // Build a map of arguments from the field.arguments AST, using the
             // variables scope to fulfill any variable references.
-            $args = Values::getArgumentValues(
+            /** @phpstan-ignore-next-line */
+            $this->fieldArgsCache[$fieldDef] ??= new \SplObjectStorage();
```
spawnia, line 745: Why ignore this?

Warxcell: phpstan complains that `SplObjectStorage<object, mixed>` is not assignable. I guess it defaults to template in SplObjectStorage. Seems impossible to tell that newly created `SplObjectStorage` will be `SplObjectStorage<FieldNode, mixed>`

```
            /** @var \SplObjectStorage<FieldNode, mixed> */
            $this->fieldArgsCache[$fieldDef] ??= new \SplObjectStorage();
```

 `745    PHPDoc tag @var does not specify variable name. `

Warxcell: That way works, but its a bit long for typing.

```
            if (!$this->fieldArgsCache->offsetExists($fieldDef)) {
                /** @var \SplObjectStorage<FieldNode, mixed> $cache2 */
                $cache2 = new \SplObjectStorage();
                $this->fieldArgsCache->offsetSet($fieldDef, $cache2);
            }
```

I don't see big difference of either way. Its local assignment anyway - no chance of something to broke.

spawnia: Ok, you can leave the code as is - but please add a short comment to the `@phpstan-ignore` to explain that the error comes from a lack of generic inference.

Warxcell: Got you.

spawnia:
Thank your @Warxcell for your hard work and persistence on this!

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.16.0.

PR #1597, opened 2024-08-19
feat: narrow errors type to list

simPod:


spawnia:
Thank you, released with https://github.com/webonyx/graphql-php/releases/tag/v15.14.1.

shadowhand:
TIL about `list` as a type hint. 🙇🏽‍♂️

simPod:
@shadowhand check this out https://phpstan.org/writing-php-code/phpdoc-types

shadowhand:
@simPod ah, because `list` has only existed since phpstan 1.9.0 and later. I knew of `Foo[]` but not `list<Foo>`. 👍🏼

PR #1599, opened 2024-08-21
Add support for native php enums to be serialized to proper value

kloining:
With native PHP enums being introduced in PHP 8.1, we have bumped into cases where we would like to provide an underlying data like: ['field'=>SomethingEnum::ENUM_VALUE], where the ENUM value corresponds with the ENUM value in graphql.

Now, we need to provide the following data: ['field'=>SomethingEnum::ENUM_VALUE->value].

This solves the following issue: GraphQL\Error\SerializationError: Cannot serialize value as enum: instance of App\SomethingEnum


spawnia:
I don't understand what this is supposed to achieve. Please add a test case that illustrates the use case and fails without this change.

kloining:
@spawnia I will provide a test asap to prove the case. PR description is updated

kloining:
@spawnia I have now added two simple tests to prove that native enums can be provided as resolver values.

This might not seem logical, but we are using webonyx with graphqls-files (not php types), and when we provide with custom resolver classes returning arrays with respons, we sometime use arrays with native enums. This will fail in our case, but with this simple tweak, it works.

I can add tests for int and UnitEnum as well if wanted

tests/Type/EnumTypeNativeEnumTest.php
```diff
@@ -0,0 +1,69 @@
...
+namespace GraphQL\Tests\Type;
+
+use GraphQL\GraphQL;
+use GraphQL\Tests\Type\PhpEnumType\StringPhpEnum;
+use GraphQL\Type\Definition\EnumType;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Schema;
+use PHPUnit\Framework\TestCase;
+
+final class EnumTypeNativeEnumTest extends TestCase
```
spawnia, line 12: Please integrate this into `EnumTypeTest`

kloining: Done @spawnia

spawnia:
Please mark as ready when you get CI to pass.

kloining:
Closing this PR, as @spawnia has done his own implementation

spawnia:
Oh, so https://github.com/webonyx/graphql-php/pull/1604 solves the issue you are having? It does not necessarily concern every possible usage of `EnumType`, so there might still be some edge cases left.

PR #1608, opened 2024-09-04
Serializes `\UnitEnum` and `\BackedEnum` in `GraphQL\Type\Definition\EnumType` when resolving return values from a Query or Mutation

Dahl99:
Since PHP has had native enums since version 8.1, there is an annoyance of not being able to return a native enum as a result from a query or mutation. If this is the case, it would have to be handled to return a string either by doing `MyEnum::A->name` or `MyEnum::A->value` depending on if it's a backed or unit enum.

This is also a bit of a nuisance if one uses DTOs for returning the data, as the type would have to be a string instead of being an enum which mirrors the enum in the GraphQL schema.

This pull request should address these issues by serializing native PHP enums if they were to be returned from a query or mutation.


spawnia:
Have you seen https://github.com/webonyx/graphql-php/pull/1604?

Dahl99:
> Have you seen #1604?

Yes, but returning a native PHP enum from a custom resolver with the schema first approach still fails. I've updated the tests to use SDL for creating the schema and they both fail without the changes to `GraphQL\Type\Definition\EnumType`.

Dahl99:
@spawnia mind having another look at this pull request?

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.14.0. Please consider sponsoring.

PR #1624, opened 2024-10-30
feat: do not override ClientSafe exception with CoercionError

simPod:
When `ClientSafe` exception is thrown during `parseValue`, it is overriden by CoercionError. Therefore, the actual error message is not propagated to the caller. This allows propagating not only `GraphQL\Error\Error` but also ClientSafe exceptions.

### DIFF
--- a/src/Utils/Value.php
+++ b/src/Utils/Value.php
@@ -2,6 +2,7 @@
 
 namespace GraphQL\Utils;
 
+use GraphQL\Error\ClientAware;
 use GraphQL\Error\CoercionError;
 use GraphQL\Error\Error;
 use GraphQL\Error\InvariantViolation;
@@ -61,7 +62,10 @@ class Value
             try {
                 return self::ofValue($type->parseValue($value));
             } catch (\Throwable $error) {
-                if ($error instanceof Error) {
+                if (
+                    $error instanceof Error
+                    || ($error instanceof ClientAware && $error->isClientSafe())
+                ) {
                     return self::ofErrors([
                         CoercionError::make($error->getMessage(), $path, $value, $error),
                     ]);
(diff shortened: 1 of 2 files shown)


src/Utils/Value.php
```diff
@@ -61,7 +62,7 @@ public static function coerceInputValue($value, InputType $type, ?array $path =
             try {
                 return self::ofValue($type->parseValue($value));
             } catch (\Throwable $error) {
-                if ($error instanceof Error) {
+                if ($error instanceof Error || ($error instanceof ClientAware && $error->isClientSafe())) {
```
spawnia, line 65:
```suggestion
                if ($error instanceof Error
                    || ($error instanceof ClientAware && $error->isClientSafe())
                ) {
```

tests/Utils/CoerceInputValueTest.php
```diff
@@ -177,6 +178,31 @@ public function testReturnsErrorForIncorrectValueType(): void
...
+                throw $clientSafeException;
+            },
+            'parseLiteral' => fn () => null,
+        ]);
+
+        $result = Value::coerceInputValue(['value' => 1], $scalar);
+        $this->expectGraphQLError($result, [CoercionError::make(
+            'message',
+            null,
+            ['value' => 1],
```
spawnia, line 202: The expected values line up with values that are previously set up in the test. Please use variables to make the association clear.

tests/Utils/CoerceInputValueTest.php
```diff
@@ -177,6 +178,33 @@ public function testReturnsErrorForIncorrectValueType(): void
...
+                throw $clientSafeException;
+            },
+            'parseLiteral' => fn () => null,
+        ]);
+
+        $value = ['value' => 1];
+
+        $result = Value::coerceInputValue($value, $scalar);
+        $this->expectGraphQLError($result, [CoercionError::make(
+            'message',
```
spawnia, line 202: also duplicated

spawnia:
Thanks, released with https://github.com/webonyx/graphql-php/releases/tag/v15.18.0

PR #1628, opened 2024-11-08
Fix the native Enum serialization

martio:
During the serialization of a native Enum, we do not normalize its name to the GraphQL standard, which requires all uppercase letters.

The full discussion on this topic can be found here: https://github.com/nuwave/lighthouse/issues/2629

### DIFF
--- a/src/Type/Definition/EnumType.php
+++ b/src/Type/Definition/EnumType.php
@@ -137,11 +137,11 @@ class EnumType extends Type implements InputType, OutputType, LeafType, Nullable
         }
 
         if (is_a($value, \BackedEnum::class)) {
-            return $value->name;
+            return mb_strtoupper($value->name);
         }
 
         if (is_a($value, \UnitEnum::class)) {
-            return $value->name;
+            return mb_strtoupper($value->name);
         }
 
         $safeValue = Utils::printSafe($value);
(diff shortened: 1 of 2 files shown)


spawnia:
There is no requirement for uppercase enum cases in the GraphQL specification, and this library is not going to be opinionated about it.

martio:
@spawnia It’s true that the GraphQL specification doesn’t require uppercase for Enums, but there is a recommendation for it. I described it in this [comment](https://github.com/nuwave/lighthouse/issues/2629#issuecomment-2464572441).

Here, the issue arises only when using casts in Laravel Eloquent. When you return a string, everything works fine.

Perhaps it’s worth adding a comparison between the key defined in schema.graphql and the one we receive as an argument? :smirk:

spawnia:
This is a breaking change for users that rely on the existing behaviour of unaltered enum adoption.

martio:
@spawnia The solution I proposed is not perfect, but I think it’s worth adding a `strict` flag for the Enum format recommended for GraphQL, along with proper formatting. I can prepare these changes if you want.

spawnia:
I want to keep this library simple. You can do a wrapper for `PhpEnumType` or extend it to achieve your desired outcome.

PR #1632, opened 2024-11-11
Automatically reset static caches in `Type`, `Introspection` and `Directive` when overriding standard types

ruudk:
See #1424

This library uses a static cache for standard types and the internal directives.

While not ideal, this works fine. As long as you don't have a scenario where you want to change them.

In my situation, we have 2 schema's with a custom ID type. They are not the same instance. This still works fine, but as soon as we run our whole test suite at once, the static cache is initialized in some test, and another test later expects something else.

I think the better solution would be to remove these static caches completely, and store them on an instance instead of statically. But that will be a much bigger task given the current architecture.

When you use `overrideStandardTypes`, the static caches will automatically be reset.

It could also be used for people that run their GraphQL server in tools like RoadRunner and FrankenPHP.


ruudk:
CI is broken because of Renovate. See https://github.com/webonyx/graphql-php/pull/1630#issuecomment-2467904626

src/Type/Definition/Directive.php
```diff
@@ -162,4 +166,9 @@ public static function isSpecifiedDirective(Directive $directive): bool
     {
         return \array_key_exists($directive->name, self::getInternalDirectives());
     }
+
+    public static function reset(): void
+    {
+        self::$internalDirectives = [];
```
spawnia, line 172: Why not `unset`? Seems simpler and keeps the rest of the code simple as well.

src/Type/Definition/Type.php
```diff
@@ -261,4 +266,10 @@ public function jsonSerialize(): string
     {
         return $this->toString();
     }
+
+    public static function reset(): void
+    {
+        static::$standardTypes = [];
+        static::$builtInTypes = [];
```
spawnia, line 273: Also simpler with `unset`.

ruudk: It's not possible to unset static variables. I got that error.

ruudk:
@spawnia Added a test that covers all 3 reset methods, since they are related. If you clear the Type's cache, the other's also have to be reset to prevent duplicate type instances.

src/Type/Definition/Type.php
spawnia: I do not like making `[]` a special case that means the variable is unset. Let's make them nullable and use `null` instead then.

ruudk: Good idea. Applied your feedback.

src/Type/Definition/Type.php
```diff
@@ -31,7 +31,10 @@ abstract class Type implements \JsonSerializable
     ];
 
     /** @var array<string, ScalarType> */
-    protected static array $standardTypes;
+    protected static ?array $standardTypes = [];
+
+    /** @var array<string, Type&NamedType>|null */
+    protected static ?array $builtInTypes = null;
```
spawnia, line 37:
```suggestion
    protected static ?array $builtInTypes;
```

src/Type/Definition/Type.php
```diff
@@ -31,7 +31,10 @@ abstract class Type implements \JsonSerializable
     ];
 
     /** @var array<string, ScalarType> */
```
spawnia, line 33:
```suggestion
    /** @var array<string, ScalarType>|null */
```

src/Type/Definition/Type.php
```diff
@@ -31,7 +31,10 @@ abstract class Type implements \JsonSerializable
...
-    protected static array $standardTypes;
+    protected static ?array $standardTypes = [];
```
spawnia, line 34:
```suggestion
    protected static ?array $standardTypes;
```

src/Type/Definition/Type.php
```diff
@@ -261,4 +262,10 @@ public function jsonSerialize(): string
     {
         return $this->toString();
     }
+
+    public static function reset(): void
+    {
+        static::$standardTypes = [];
```
spawnia, line 268:
```suggestion
        static::$standardTypes = null;
```

spawnia:
> Another idea could be to introduce a single TypeCache object, where all the types are cached.
> 
> Then there is only 1 place to reset the cache.

I like the idea of centralizing this into a single method that library users are expected to call. From the perspective of a user of this library, isn't the need to reset anything tied to using `Type::overrideStandardTypes()`? So how about we just reset all dependent caches when it is called. Or are there other situations where it would be necessary?

ruudk:
Smart! I'll test if that is sufficient.

ruudk:
Alright. When I do it like this, it works perfectly for my use case. But there is 1 test that fails: `testPreservesOriginalStandardTypes`.

I think this test does not make much sense to be honest. When you are overriding the standard types, why would you expect the _other_ standard types to be the same instance? Is it expected that the standard types are already initialized (used) before calling overrideStandardTypes? I don't think so.

What do you think @spawnia ?

src/Type/Definition/Type.php
```diff
@@ -267,7 +267,6 @@ public function jsonSerialize(): string
 
     public static function reset(): void
     {
-        static::$standardTypes = null;
```
ruudk: This needs to be reset, in my use case (running phpunit tests that once set this)

ruudk: What if we add a second argument to the overrideStandardTypes method that resets it completely? By default it's off, to mimic the existing logic. But it allows to reset and override fully.

ruudk: Pushed a commit to explain what I mean

ruudk: @spawnia Pushed a commit to explain what I mean

src/GraphQL.php
```diff
@@ -211,14 +211,15 @@ public static function getStandardTypes(): array
      * Standard types not listed here remain untouched.
      *
      * @param array<string, ScalarType> $types
+     * @param bool $partial Whether to replace only types listed in $types or all standard types
```
spawnia, line 214: This makes little sense to me. If this method were called with `types: ['ID' => $newIDType], partial: false`, what would that even mean? How could a subset of the required standard types fully replace all standard types?

src/Type/Definition/Type.php
```diff
@@ -144,11 +145,18 @@ public static function getStandardTypes(): array
...
      * @throws InvariantViolation
      */
-    public static function overrideStandardTypes(array $types): void
+    public static function overrideStandardTypes(array $types, bool $partial = true): void
     {
+        self::reset();
+
+        if (! $partial) {
+            static::$standardTypes = null;
+        }
```
spawnia, line 158: I don't see why it would be necessary to force reinstantiation of standard types that are not being overridden.

src/GraphQL.php
ruudk: Imagine you have 2 schema's.

In schema 1 you override ID type.

In schema 2 you override String type.

Schema 1 doesn't want to get schema's 2 string type at any point.
Schema 2 doesn't want to get schema's 1 ID type at any point.

With normal PHP FPM this is not an issue as only one schema is active within a request and destroyed afterwards.
But when doing unit tests, or even worse, functional tests, you it could be that one test used schema 1, that initialized the standard types. Then later a test uses schema 2, and boom. It breaks.

So the only way to be able to do this, is to wipe the cache.

Or, to explicitly reinitialize the types that are not overridden, so that the array always contains the full standard types.

ruudk: I guess I can also solve it like this:
```php
Type::overrideStandardTypes([
            Type::INT => isset($types[Type::INT]) ? $loader(Type::INT) : new IntType(),
            Type::FLOAT => isset($types[Type::FLOAT]) ? $loader(Type::FLOAT) : new FloatType(),
            Type::STRING => isset($types[Type::STRING]) ? $loader(Type::STRING) : new StringType(),
            Type::BOOLEAN => isset($types[Type::BOOLEAN]) ? $loader(Type::BOOLEAN) : new BooleanType(),
            Type::ID => isset($types[Type::ID]) ? $loader(Type::ID) : new IDType(),
        ]);
```

ruudk: @spawnia Just confirmed, that this also works. So I removed the partial thing from the PR.

It's ready for review again!

spawnia:
I modified the test to be more explicit about what it is trying to achieve and also make it more understandable why we would want that. @ruudk Are you ok with the current state?

ruudk:
@spawnia Yeah this makes sense. I tested it on our CI and it works 🎉  Thanks for the help!

spawnia:
Thank you @ruudk, released with https://github.com/webonyx/graphql-php/releases/tag/v15.18.1.

PR #1639, opened 2024-11-12
Add PHP 8.4 to CI matrix

simPod:


ruudk:
@simPod The auto format workflow is not ideal. Whenever it pushes a change, it doesn't trigger the workflows, because it's not a PAT. If the contributor wants their PR green again, they have to pull and rebase/force push, to get it green again.

simPod:
I did some work around that in the past

https://github.com/webonyx/graphql-php/commit/9dbc47ac233564ed1d57048c86de8e9ed4bb4c2a it should use PAT or user token.

also this https://github.com/webonyx/graphql-php/pull/1506

-------

so are you saying it is not working for MRs created from fork now?

ruudk:
Just had it on my latest PR. The checks all are orange and I have to pull and force push

Issue #1072, opened 2022-02-16
Include aliases in query plan


crissi:
Running lookAhead with queryplan with this query: 
```   
$resolveInfo->lookAhead(['group-implementor-fields'])->queryPlan();
```
```   
query  {
  caseById(id: 7317) {
    default: schemesCount
    closed: schemesCount(status: CLOSED)
    approved: schemesCount(status: APPROVED)
  }
}
```
Gives us:
```
{
  "fields": {
    "schemesCount": {
      "type": "Int",
      "fields": [],
      "args": {
        "status": "APPROVED"
      }
    }
  }
}

```
I suspect all fields should be included somehow in the result!

spawnia:
I think the purpose of the query plan is to allow the server to make decisions on how to resolve the query efficiently. Given aliases are given by the client and transparently handled by the execution engine of this library, they do not seem useful to include.

crissi:
yes, but since each alias can have different arguments like in my example. 

Here it just returns the argument of the last **schemesCount** property.

spawnia:
Right, that is misleading. `args` should really be a list with multiple args passed to the different aliases - same with `fields` actually. Perhaps a better structure would be:

```json
{
  "fields": {
    "schemesCount": {
      "type": "Int",
      "aliases": [
        {
          "fields": [],
          "args": {}
        },
        {
          "fields": [],
          "args": {
            "status": "APPROVED"
          }
        },
        {
          "fields": [],
          "args": {
            "status": "CLOSED"
          }
        },
      ]
    }
  }
}
```

crissi:
maybe include the alias name in there  for good measure

jeroenschieven:
We are having the same issue. We use the queryPlan to dynamically build an ElasticSearch query. Some fields have quite some possible arguments, therefore we want to be able to alias those fields.

Unfortunately, the only alternative for now is duplicating these fields.

zephyx:
I needed this too, so I have modified the lib to get this feature.
I don't know if it will be merged or no, but if needed for your application: https://github.com/webonyx/graphql-php/pull/1648

I use it as follow:

=> In My Query Resolver : 

$aliasArgs = $info->lookAhead()->aliasArgs();
//I can then create multiple results for each of the aliases like this:
$res[$aliasName] = $myResult;

=> In My Type Resolver : 
//I have access to all my results with $root and I have just to access to them with a 
return $root[$info->fieldNodes[0]->alias->value];

[...]

PR #1648, opened 2024-11-29
Add method `getFieldSelectionWithAliases` to class `ResolveInfo`

zephyx:
@see https://github.com/webonyx/graphql-php/issues/1072

That was the simplest way to solve this issue with minimal breaking change due to the presence of $this->arrayMergeDeep(...).

If you have a better idea, no problem, I just needed this feature in my application.

NOTE: composer fix / check have updated several files but I have not modified them, so they are not included here.

### DIFF
--- a/src/Type/Definition/ResolveInfo.php
+++ b/src/Type/Definition/ResolveInfo.php
@@ -128,11 +128,14 @@ class ResolveInfo
     /**
      * @param \ArrayObject<int, FieldNode> $fieldNodes
      * @param list<string|int> $path
+     *
      * @phpstan-param Path $path
+     *
      * @param array<string, FragmentDefinitionNode> $fragments
      * @param mixed|null $rootValue
      * @param array<string, mixed> $variableValues
      * @param list<string|int> $unaliasedPath
+     *
      * @phpstan-param Path $unaliasedPath
      */
     public function __construct(
--- a/src/Type/Definition/ResolveInfo.php
+++ b/src/Type/Definition/ResolveInfo.php
@@ -219,7 +219,7 @@ class ResolveInfo
     /**
      * Returns names of all fields selected in query for `$this->fieldName` up to `$depth` levels.
      *
-     * For each field there is a key "aliases" that lists all aliases of this field,
+     * For each field we create an array containing all aliases of this field,
      * or the original field name if a field is not aliased.
      *
      * For each of those aliases, you can find the following keys:
@@ -244,48 +244,40 @@ class ResolveInfo
      * this method will return:
      * [
      *     'id' => [
-     *          'aliases' => [
-     *              'id' => [
-     *                  'args' => []
-     *              ]
+     *          'id' => [
+     *               'args' => []
      *          ]
      *      ],
      *      'nested' => [
-     *           'aliases' => [
-     *               'nested' => [
-     *                  'args' => [],
-     *                  'fields' => [
+     *          'nested' => [
+     *              'args' => [],
+     *              'fields' => [
+     *                  'nested1' => [
      *                      'nested1' => [
-     *                          'aliases' => [
-     *                              'nested1' => [
-     *                                  'args' => [
-     *                                      'myArg' => 1
-     *                                  ]
-     *                              ],
-     *                              'nested1Bis' => [
-     *                                  'args' => []
-     *                              ]
-     *                          ]
-     *                      ]
-     *                  ]
-     *               ],
-     *               'alias1' => [
-     *                  'args' => [],
-     *                  'fields' => [
-     *                       'nested1' => [
-     *                           'aliases' => [
-     *                               'nested1' => [
-     *                                   'args' => [
-     *                                       'myArg' => 2,
-     *                                       'mySecondAg' => "test"
-     *                                   ]
-     *                               ]
+     *                           'args' => [
+     *                               'myArg' => 1
      *                           ]
+     *                       ],
+     *                       'nested1Bis' => [
+     *                           'args' => []
+     *                       ]
+     *                  ]
+     *              ]
+     *           ],
+     *           'alias1' => [
+     *              'args' => [],
+     *              'fields' => [
+     *                   'nested1' => [
+     *                       'nested1' => [
+     *                            'args' => [
+     *                                'myArg' => 2,
+     *                                'mySecondAg' => "test"
+     *                            ]
      *                       ]
      *                   ]
      *               ]
-     *           ]
-     *       ]
+     *          ]
+     *      ]
      * ]
      *
      * Warning: this method it is a naive implementation which does not take into account
@@ -419,10 +411,10 @@ class ResolveInfo
                 if ($fieldType instanceof WrappingType) {
                     $fieldType = $fieldType->getInnermostType();
                 }
-                $fields[$fieldName]['aliases'][$aliasName]['args'] = Values::getArgumentValues($fieldDef, $selectionNode, $this->variableValues);
+                $fields[$fieldName][$aliasName]['args'] = Values::getArgumentValues($fieldDef, $selectionNode, $this->variableValues);
 
                 if ($descend > 0 && $selectionNode->selectionSet !== null) {
-                    $fields[$fieldName]['aliases'][$aliasName]['fields'] = $this->foldSelectionWithAlias($selectionNode->selectionSet, $descend - 1, $fieldType);
+                    $fields[$fieldName][$aliasName]['fields'] = $this->foldSelectionWithAlias($selectionNode->selectionSet, $descend - 1, $fieldType);
                 }
             } elseif ($selectionNode instanceof FragmentSpreadNode) {
                 $spreadName = $selectionNode->name->value;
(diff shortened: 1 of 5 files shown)


zephyx:
I've made a few changes in order to manage the case where we have a field which is not aliased and the same one aliased.
Before we were loosing the args of the first one, not now with those modifications.

spawnia:
Please add tests.

zephyx:
After writing some tests I have seen an edge case and completely refactored the code and moved it somewhere else.
It's late, I will finish the tests tomorrow or sometime this week :)

src/Type/Definition/ResolveInfo.php
```diff
@@ -214,6 +216,126 @@ public function getFieldSelection(int $depth = 0): array
...
+     *          'aliases' => [
+     *              'id' => [
+     *                  [args] => []
+     *              ]
+     *          ]
+     *      ],
+     *      'nested' => [
+     *           'aliases' => [
+     *               'nested' => [
+     *                  ['args'] => [],
```
spawnia, line 255: This does not look like a valid array, what is going on with the keys?

src/Type/Definition/ResolveInfo.php
```diff
@@ -214,6 +216,126 @@ public function getFieldSelection(int $depth = 0): array
...
+     * For each field there is an "aliases" key regrouping all aliases of this field
+     * (if a field is not aliased, its name is present here as a key)
+     * For each of those "alias" you can find "args" key containing the arguments of the alias and the "fields" key
+     * containing the subfield of this field/alias. Each of those field have the same structure as described above.
+     *
+     * Example:
+     * query MyQuery{
+     * {
+     *   root {
+     *     id,
```
spawnia, line 231:
```suggestion
     *     id
```

src/Type/Definition/ResolveInfo.php
```diff
@@ -214,6 +216,126 @@ public function getFieldSelection(int $depth = 0): array
...
+    /**
+     * Helper method that returns names of all fields selected in query for
+     * $this->fieldName up to $depth levels.
+     * For each field there is an "aliases" key regrouping all aliases of this field
+     * (if a field is not aliased, its name is present here as a key)
+     * For each of those "alias" you can find "args" key containing the arguments of the alias and the "fields" key
+     * containing the subfield of this field/alias. Each of those field have the same structure as described above.
+     *
+     * Example:
+     * query MyQuery{
```
spawnia, line 228:
```suggestion
```

src/Type/Definition/ResolveInfo.php
```diff
@@ -214,6 +216,126 @@ public function getFieldSelection(int $depth = 0): array
...
+     *           ]
+     *       ]
+     * ]
+     *
+     * Warning: this method it is a naive implementation which does not take into account
+     * conditional typed fragments. So use it with care for fields of interface and union types.
+     * You still can alias the union type fields with the same name in order to extract their corresponding args.
+     *
+     * Example:
+     *  query MyQuery{
```
spawnia, line 295:
```suggestion
```

src/Type/Definition/ResolveInfo.php
```diff
@@ -214,6 +216,126 @@ public function getFieldSelection(int $depth = 0): array
...
+     *  {
+     *    root {
+     *      id,
```
spawnia, line 298:
```suggestion
     *      id
```

src/Type/Definition/ResolveInfo.php
```diff
@@ -266,4 +388,61 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+        foreach ($selectionSet->selections as $selectionNode) {
+            if ($selectionNode instanceof FieldNode) {
+                $fieldName = $selectionNode->name->value;
+                $aliasName = $selectionNode->alias->value ?? $fieldName;
+
+                if ($fieldName === Introspection::TYPE_NAME_FIELD_NAME) {
+                    continue;
+                }
+
+                assert($parentType instanceof HasFieldsType, 'ensured by query validation and the check above which excludes union types');
```
spawnia, line 411: Is this copy-pasted from somewhere? The comment makes little sense to me in this context.

tests/Type/ResolveInfoTest.php
```diff
@@ -648,4 +648,263 @@ public function testPathAndUnaliasedPathForList(): void
             ],
         ], $result);
     }
+
+    private int $aliasArgsNbTests = 0;
```
spawnia, line 652: Use a local variable that is included in the closure by reference (`use (&$var)`).

spawnia (changes requested):

spawnia:
The PHPStan error should be fixable with an `assert` call to ensure non-null.

src/Type/Definition/ResolveInfo.php
zephyx: Should I also update the one from the function getFieldSelection?

src/Type/Definition/ResolveInfo.php
zephyx: I was using the same logical as QueryPlan, we don't want stuff like __typename and we need to have fieldsType.
But I can remove this if you want.

src/Type/Definition/ResolveInfo.php
```diff
@@ -214,6 +216,126 @@ public function getFieldSelection(int $depth = 0): array
...
+     *
+     * @throws \Exception
+     * @throws Error
+     * @throws InvariantViolation
+     *
+     * @return array<string, mixed>
+     *
+     * @api
+     */
+    public function getFieldSelectionWithAlias(int $depth = 0): array
```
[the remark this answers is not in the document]
zephyx: 1 goes down one level, 0 stays at the root level.
This is the same logic as getFieldSelection(), for reasons of consistency.

I can go further 1 if you want, I am not expecting any side effect further 1, I have already tested this at 2 when writing the code.

-X depth = 0 depth
getFieldSelection() run like this, I don't think I should throw an error here if there are no errors in getFieldSelection().
I tried to keep the same logic.

zephyx:
Updated

spawnia:
I know I originally proposed it like this in https://github.com/webonyx/graphql-php/issues/1072#issuecomment-1041345447, but there we also included the type name per field. Given we no longer need that, is there any reason why we should nest aliases under a key `aliases`, as in this example:

```php
[
    'level2' => [
        'aliases' => [
            'level2000' => [
                'args' => [
                    'width' => 1,
                    'height' => 1,
                ],
            ],
            'level2' => [
                'args' => [
                    'width' => 2,
                    'height' => 20,
                ],
            ],
        ],
    ],
]
```

Or can we simplify like this?

```php
[
    'level2' => [
        'level2000' => [
            'args' => [
                'width' => 1,
                'height' => 1,
            ],
        ],
        'level2' => [
            'args' => [
                'width' => 2,
                'height' => 20,
            ],
        ],
    ],
]
```

zephyx:
Yes, in current state, we can simplify this
But when I started writing this code, I imagined adding other information like the type you mention.
I forgot this point during my refactoring, but for example adding the type is just one line of code.
Leaving this level “aliases” would make it possible to add other information to the return if we want to evolve the function, no?

If it's ok with you, I can add this line to the code: 
$fields[$fieldName]['type'] = $fieldType->name();

This will return if you keep the types described in tests : 
```
[
    'level2' => [
        'type' => 'Int'
        'aliases' => [
            'level2000' => [
                'args' => [
                    'width' => 1,
                    'height' => 1,
                ],
            ],
            'level2' => [
                'args' => [
                    'width' => 2,
                    'height' => 20,
                ],
            ],
        ],
    ],
     'level2bis' => [
        'type' => 'level2bis'
        'aliases' => [
            'level2bis' => [
                'args' => [],
                'fields' => [
                     'level3' => [
                         'type' => 'Int',
                         'aliases' => [
                             'level3' => [
                                 'args' => [
                                     'length' => 10
                                 ]
                             ]
                         ]
                     ]
               ]
            ]
        ],
    ],
]
```

src/Type/Definition/ResolveInfo.php
```diff
@@ -203,10 +204,135 @@ public function getFieldSelection(int $depth = 0): array
         $fields = [];
 
         foreach ($this->fieldNodes as $fieldNode) {
-            if (isset($fieldNode->selectionSet)) {
+            $selectionSet = $fieldNode->selectionSet;
```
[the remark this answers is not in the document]
spawnia: Yup, it is just a refactoring. Given much of the code was copy-pasted, I wanted to improve the original too.

zephyx:
> Adding the type name is kind of an arbitrary extra information anyways. Nullability or being in a list could also matter - as could other information about the field or the schema at that point.

Yes, that was my point (evolution): if latter we need more information, it would have been easy to add it alongside the "aliases" level. Once removed, we won't be able to add field related informations without introducing breaking changes.
But nevermind, I can remove it if you want.

spawnia:
We can always add a new API if something more complex is necessary. Let's keep the simple things simple and only add complexity when it is really necessary.

spawnia:
I renamed the nested `fields` to `selectionSet`. This terminology follows the language of the GraphQL specification when talking about queries, see https://spec.graphql.org/draft/#Field. It also clearly distinguishes this from the type level, it is about the current execution and not about what is in the schema.

spawnia:
Thank you @zephyx, released with https://github.com/webonyx/graphql-php/releases/tag/v15.19.0.

PR #1660, opened 2024-12-02
:package: Drop `@throws Error` from protected method `ReferenceExecutor::executeOperation`

andrew-demb:
This method catches `Error` and converts them to the `return null` and adds error to the context (https://github.com/webonyx/graphql-php/blob/f7b2ee082e32be5bbd14baf2c4fcf7aa2bd450c8/src/Executor/ReferenceExecutor.php#L328-L332)


spawnia:
Thank you @andrew-demb

PR #1664, opened 2024-12-13
ResolveInfo->getFieldSelectionWithAliases(): Fix edge case query with ListOf Type

zephyx:
I have found a problem with the function getFieldSelectionWithAliases() and Type::listOf() during some new tests.
I fix that here.


spawnia:
Thank you @zephyx, released with https://github.com/webonyx/graphql-php/releases/tag/v15.19.1.

PR #1668, opened 2025-01-21
Calculate query complexity with fragments correctly independent of order

### LINKED ISSUE #785: Query complexity is missing on fields included by fragments

spawnia:
Fixes https://github.com/webonyx/graphql-php/issues/785


spawnia:
Closing — after merging master, this PR has an empty diff. The changes are already in master.

PR #1672, opened 2025-02-17
Transfer config in schema extender

### LINKED ISSUE #1671: Feature: ability to attach custom attributes to field definitions.
Hey, I'm writing custom Executor, optimized for my needs (basically I need per coordinate cache) - so my idea is to attach custom cache resolver which will be called prior original one to check from cache. The issue is that it's very hard to do that - because I need to fork basically whole lib - it would be awesome if it's possible to attach custom attributes to field definition so I can check them later in my custom Executor. 

What do you think?

Warxcell:
Fixes https://github.com/webonyx/graphql-php/issues/1671


spawnia:
I just applied the codestyle fixes through https://github.com/webonyx/graphql-php/pull/1674, please merge the latest `master` before I review.

PR #1675, opened 2025-02-26
Make NodeVisitor Node argument covariant

ruudk:
When creating a node visitor that uses more specific types, we get this error:

```
Parameter #2 $visitor of static method GraphQL\Language\Visitor::visit() expects array<string, array<string, callable(GraphQL\Language\AST\Node): (GraphQL\Language\VisitorOperation|void|false|null)>|(callable(GraphQL\Language\AST\Node): (GraphQL\Language\VisitorOperation|void|false|null))>, array{OperationDefinition: array{enter: Closure(GraphQL\Language\AST\OperationDefinitionNode): void, leave: Closure(): void}, Field: array{enter: Closure(GraphQL\Language\AST\FieldNode): void, leave: Closure(GraphQL\Language\AST\FieldNode): void}, InlineFragment: array{enter: Closure(GraphQL\Language\AST\InlineFragmentNode): void, leave: Closure(GraphQL\Language\AST\InlineFragmentNode): void}, FragmentDefinition: array{enter: Closure(GraphQL\Language\AST\FragmentDefinitionNode): void, leave: Closure(GraphQL\Language\AST\FragmentDefinitionNode): void}, SelectionSet: Closure(GraphQL\Language\AST\SelectionSetNode): void} given.
```

For the following code:
```php
$ast = Visitor::visit($ast, [
                NodeKind::OPERATION_DEFINITION => [
                    'enter' => function (OperationDefinitionNode $node) : void {
                        $this->parents[] = $node->operation;

                        // Remove the operation name as we don't need this
                        $node->name = null;
                    },
                    'leave' => function () : void {
                        array_pop($this->parents);
                    },
                ],
                NodeKind::FIELD => [
                    'enter' => function (FieldNode $node) : void {
                        $this->parents[] = $node->name->value;
                    },
                    'leave' => function (FieldNode $node) : void {
                        array_pop($this->parents);
                    },
                ],
 // ...
```

By changing the signature to `covariant` the problem goes away.

### DIFF
--- a/src/Language/Visitor.php
+++ b/src/Language/Visitor.php
@@ -89,7 +89,7 @@ use GraphQL\Utils\Utils;
  *       ]
  *     ]);
  *
- * @phpstan-type NodeVisitor callable(Node): (VisitorOperation|null|false|void)
+ * @phpstan-type NodeVisitor callable(covariant Node): (VisitorOperation|null|false|void)
  * @phpstan-type VisitorArray array<string, NodeVisitor>|array<string, array<string, NodeVisitor>>
  *
  * @see \GraphQL\Tests\Language\VisitorTest
--- a/docs/class-reference.md
+++ b/docs/class-reference.md
@@ -1298,7 +1298,7 @@ visitor API:
    ]
    ]);
 
-@phpstan-type NodeVisitor callable(Node): (VisitorOperation|null|false|void)
+@phpstan-type NodeVisitor callable(covariant Node): (VisitorOperation|null|false|void)
 @phpstan-type VisitorArray array<string, NodeVisitor>|array<string, array<string, NodeVisitor>>
 
 @see \GraphQL\Tests\Language\VisitorTest


ruudk:
I guess this doesn't work. When I tested it, it produced no error, but that's probably because PHPStan skipped the error completely.

spawnia:
In order to make this work, we would have to define an array shape for the visitor. Something like this:

```
@phpstan-type VisitorArray array{
    OperationDefinition?: array{
        enter?: callable(OperationDefinitionNode): (VisitorOperation|null|false|void),
        leave?: callable(OperationDefinitionNode): (VisitorOperation|null|false|void),
    }|callable(OperationDefinitionNode): (VisitorOperation|null|false|void),
    Field?: ...
```

How badly do you want this? 😉

spawnia:
Yeah, I am fine with it. It should allow use to get rid of some `@phpstan-ignore` we have around `Visitor::visit` as well, and offer a better typesafe API for users. I just hope we don't run into some kind of limit PHPStan has around type complexity. Let's see!

src/Language/Visitor.php
```diff
@@ -89,8 +112,102 @@
  *       ]
  *     ]);
  *
- * @phpstan-type NodeVisitor callable(Node): (VisitorOperation|null|false|void)
- * @phpstan-type VisitorArray array<string, NodeVisitor>|array<string, array<string, NodeVisitor>>
+ * @phpstan-type VisitorArray array{
+ *      enter?: callable(Node): (VisitorOperation|null|false|void),
```
spawnia, line 116: Let's add a type alias for `VisitorOperation|null|false|void`

ruudk:
These docs / cs fixer pushes are handy and annoying at the same time... they don't trigger CI to run.. So I have to pull them , force push it.

spawnia:
CI does run for the previously pushed commit, you can check the results there.

image

ruudk:
After this change, PHPStan time shoots from 30 seconds to multiple minutes on CI 😭

spawnia:
> After this change, PHPStan time shoots from 30 seconds to multiple minutes on CI 😭

I am not sure if it is worth the added complexity and worse development experience then. I imagine it would also slow down PHPStan for consumers of the library significantly?

ruudk:
Maybe @ondrejmirtes has some insights on how to solve this type of problem. I tried searching for this, but I probably use the wrong terminology so couldn't find similar things.

ondrejmirtes:
Please formulate the question by providing full context, I have no idea what's going on here.

ruudk:
You're right, I feel bad for not giving you the proper context when pinging you. My apologies.

This library has an AST visitor that works like this: https://github.com/webonyx/graphql-php/blob/ed6b5cc7848103cd7ebbd794c91307372e48983a/src/Language/Visitor.php#L30-L90

For example:
```php
$ast = Visitor::visit($ast, [
    NodeKind:: OPERATION_DEFINITION => [
        'enter' => function ($node) : void {
            // $node is OperationDefinitionNode
        },
    ],
    NodeKind::FIELD => [
        'enter' => function ($node) : void {
            // $node is FieldNode
        },
    ],
]);
```

Depending on the node kind you add the visitor for, you get the corresponding Node type.

I want to write my visitor like this:
```php
$ast = Visitor::visit($ast, [
    NodeKind:: OPERATION_DEFINITION => [
        'enter' => function (OperationDefinitionNode  $node) : void {},
    ],
    NodeKind::FIELD => [
        'enter' => function (FieldNode  $node) : void {},
    ],
]);
```

But that doesn't work because you get an error like this:
> Parameter #2 $visitor of static method GraphQL\Language\Visitor::visit() expects array<string, array<string, callable(GraphQL\Language\AST\Node): (GraphQL\Language\VisitorOperation|void|false|null)>|(callable(GraphQL\Language\AST\Node): (GraphQL\Language\VisitorOperation|void|false|null))>, array{OperationDefinition: array{enter: Closure(GraphQL\Language\AST\OperationDefinitionNode): void, leave: Closure(): void}, Field: array{enter: Closure(GraphQL\Language\AST\FieldNode): void, leave: Closure(GraphQL\Language\AST\FieldNode): void}, InlineFragment: array{enter: Closure(GraphQL\Language\AST\InlineFragmentNode): void, leave: Closure(GraphQL\Language\AST\InlineFragmentNode): void}, FragmentDefinition: array{enter: Closure(GraphQL\Language\AST\FragmentDefinitionNode): void, leave: Closure(GraphQL\Language\AST\FragmentDefinitionNode): void}, SelectionSet: Closure(GraphQL\Language\AST\SelectionSetNode): void} given.

In this PR I changed the type of the `VisitorArray` (2nd param to Visitor::visit) and made it a fully defined array shape:
https://github.com/webonyx/graphql-php/blob/606c093b2e83c4cbbe3caf063888150f85a389ce/src/Language/Visitor.php#L115-L210

It's not nice, but at least the developer now has full type safety.

It [works](https://github.com/webonyx/graphql-php/actions/runs/13569334654/job/37930026286), except it now takes ±4 minutes to analyze. [Before](https://github.com/webonyx/graphql-php/actions/runs/13567542744/job/37923949375) it would take 30 seconds.

So this is a performance issue, and also something that might be expected when creating such a monstrous array shape. 

But that leads me to the question: can this be done simpler? 

Maybe it could be done once we have support for [class-string-map](https://github.com/phpstan/phpstan/issues/9521) and then leverage the info defined here: https://github.com/webonyx/graphql-php/blob/ed6b5cc7848103cd7ebbd794c91307372e48983a/src/Language/AST/NodeKind.php#L76-L137

ruudk:
> Alright, so you want to report a performance PHPStan issue. Please create an isolated gist.github.com and open an issue on PHPStan's GitHub.

Done here: https://github.com/phpstan/phpstan/issues/12661

> Also you can try creating an example on phpstan.org/try with two or three different NodeKinds (so it finishes in time) and I can try rewriting it using https://phpstan.org/writing-php-code/phpdoc-types#offset-access or something like that.

This is the small variant that runs on the playground: https://phpstan.org/r/a3fde113-c526-4b87-be65-ee8c2586217c

ondrejmirtes:
I'm sorry, I wasn't able to figure out a different syntax with the "offset access type", it's too complicated.

ruudk:
Thanks for checking anyway 🙏 

Closing this .

PR #1679, opened 2025-03-12
ci: setup autofix

simPod:


simPod:
for some reason the workflow is not triggered even though PAT is set 🤔

.github/workflows/autoformat.yml
```diff
@@ -71,6 +67,4 @@ jobs:
 
       - run: vendor/bin/php-cs-fixer fix
 
-      - uses: stefanzweifel/git-auto-commit-action@v5
-        with:
-          commit_message: Apply php-cs-fixer changes
+      - uses: autofix-ci/action@v1
```
simPod, line 70: Requested app installation with @vladar 

It works smoooth https://github.com/simPod/graphql-php/pull/4

.github/workflows/autoformat.yml
```diff
@@ -52,14 +52,10 @@ jobs:
     runs-on: ubuntu-latest
 
     permissions:
-      # Give the default GITHUB_TOKEN write permission to commit and push the changed files back to the repository.
-      contents: write
+      contents: read
```
spawnia, line 55: Can we omit this?

.github/workflows/autoformat.yml
spawnia, line 70:
```suggestion
      - uses: autofix-ci/action@v1
        with:
          commit-message: Apply php-cs-fixer changes
```

spawnia (commented):
Great find! I guess I will have to update like a dozen repos with this when it proves to work well 😅

spawnia:
Can we make it so Static Analysis and Test checks run on the fix commit?

image

simPod:
@spawnia it's running

image

simPod:
It actually has permissions for other users as well 🤔 https://github.com/webonyx/graphql-php/pull/1694

simPod:
The only problem I see is that actual checks are not visible in jobs

image

image

the solution is to somewhat dupe em I guess

autofix-troubleshooter:
Hi! I'm the <a href="https://autofix.ci">autofix logo<strong>autofix.ci</strong></a> troubleshooter bot.

It looks like you correctly set up a CI job that uses the autofix.ci GitHub Action, but the [**autofix.ci GitHub App**](https://github.com/apps/autofix-ci/) has not been installed for this repository. This means that autofix.ci unfortunately does not have the permissions to fix this pull request. If you are the repository owner, please install the app and then restart the CI workflow! 😃

.github/workflows/autoformat.yml
spawnia, line 1: How about we rename this workflow file to `autofix.yml`?

.github/workflows/autoformat.yml
spawnia: I guess we can keep it around.

simPod:
@spawnia the app expects a single artifact `autofix.ci.zip`, therefore I had to merge it all into a single job to support multiple style violations but I'm fine with it.

.github/workflows/autoformat.yml
spawnia: Looks pretty sleek now. Is the installation ready?

spawnia:
Thank you @simPod for digging into this! I hope it is going to work well in the future.

PR #1681, opened 2025-03-17
ResolveInfo::getFieldSelectionWithAliases() => now add instance types and the folded union types to the returned schema.

zephyx:
Hi, 

After using the ResolveInfo::getFieldSelectionWithAliases() method I needed some additionnal data in order to go through the all schema properly.
So I have updated the method to add: 
- Type instance for each alias level (not set at field level in order to keep the schema simple and avoid breaking changes)
- Union type management: This allows us to properly go through unions types and retrieve all their respective subfields/args

Before this update, we were able to go through union types but it was unreliable, some fields with the same name in different unioned types could be substituted for others.
Now it's consistent and it allows me in my application to handle a proper infinite level eager loading across the entire schema with unions.


src/Type/Definition/ResolveInfo.php
```diff
@@ -235,79 +237,107 @@ public function getFieldSelection(int $depth = 0): array
...
+     *       }
+     *       ...on MyCustomObject {
+     *         nested3
+     *       }
+     *     }
      *   }
      * }
      *
-     * Given this ResolveInfo instance is a part of "root" field resolution, and $depth === 1,
+     * Given this ResolveInfo instance is a part of "root" field resolution, $depth === 1, and nested represent an ObjectType with a configured name "Nested",
```
spawnia, line 251: What exactly is meant by `nested` here? A field?

src/Type/Definition/ResolveInfo.php
```diff
@@ -235,79 +237,107 @@ public function getFieldSelection(int $depth = 0): array
...
-     *                      'nested1' => [
-     *                           'args' => [
-     *                               'myArg' => 2,
-     *                               'mySecondAg' => "test,
-     *                           ],
+     *                 'nested1' => [
+     *                     'nested1' => [
+     *                          'args' => [
+     *                              'myArg' => 2,
+     *                              'mySecondAg' => "test,
```
spawnia, line 289: Looks broken

src/Type/Definition/ResolveInfo.php
```diff
@@ -235,79 +237,107 @@ public function getFieldSelection(int $depth = 0): array
...
-     *                  ],
-     *              ],
+     *                 ],
+     *             ],
      *         ],
      *     ],
+     *     'myUnion' => [
+     *         'myUnion' => [
+     *              'args' => [
+     *                  'myArg' => 3
```
spawnia, line 300: Always add trailing commas

src/Type/Definition/ResolveInfo.php
```diff
@@ -235,79 +237,107 @@ public function getFieldSelection(int $depth = 0): array
...
      *
-     * @throws \Exception
+     * @return array<string, mixed>
+     *
      * @throws Error
      * @throws InvariantViolation
      *
-     * @return array<string, mixed>
-     *
+     * @throws \Exception
```
spawnia, line 340: I liked the previous order better

src/Type/Definition/ResolveInfo.php
```diff
@@ -410,12 +440,17 @@ private function foldSelectionWithAlias(SelectionSetNode $selectionSet, int $des
                     continue;
                 }
 
-                assert($parentType instanceof HasFieldsType, 'ensured by query validation');
-
                 $fieldDef = $parentType->getField($fieldName);
```
spawnia, line 443: Removing the `assert` causes PHPStan failures.

tests/Type/ResolveInfoTest.php
```diff
@@ -855,6 +1000,29 @@ public function testGetFieldSelectionWithAliases(): void
             GRAPHQL
         );
 
+        $result10 = GraphQL::executeQuery(
+            new Schema(['query' => $query]),
+            <<<GRAPHQL
+            query {
```
spawnia, line 1006:
```suggestion
            {
```

tests/Type/TestClasses/CustomWithObject.php
```diff
@@ -0,0 +1,19 @@
...
+{
+    public function __construct()
+    {
+        $config = [
+            'fields' => [
+                'customA' => new MyCustomType(),
+                'customB' => new OtherCustom(),
+            ],
+        ];
+        parent::__construct($config);
```
spawnia, line 17:
```suggestion
        parent::__construct([
            'fields' => [
                'customA' => new MyCustomType(),
                'customB' => new OtherCustom(),
            ],
        ]);
```

spawnia (changes requested):
Please fix the PHPStan issues, then mark as ready.

zephyx:
Hi, thank you for the review.

How should I fix this PHPStan error ?

`Cannot access offset 'customA' on (callable(): iterable)|iterable.`

On [tests/Type/ResolveInfoTest.php](https://github.com/webonyx/graphql-php/pull/1681/files#diff-c211bcea5142301bd0a1f3b5069a86e2519291729f691d8c1d976f90ff47f814)

Line 741

src/Type/Definition/ResolveInfo.php
zephyx: Yes, I mean that the field described on line 233 represents an ObjectType with a property 'name' (or the result of the ->name() method) = "Nested".
I wrote this in order to emphasize the link between this field and the first ObjectType field in "myUnion" which contains the same field type.
It's probably not clear enough without the ObjectType + UnionType class descriptions :/

spawnia:
```suggestion
     * Given this ResolveInfo instance is a part of root field resolution,
     * $depth === 1,
     * and fields "nested" represents an ObjectType named "Nested",
```

spawnia:
Thank you, released with https://github.com/webonyx/graphql-php/releases/tag/v15.20.0.

PR #1684, opened 2025-03-18
chore(deps): bump thecodingmachine/safe for PHP 8.4

jorgsowa:
Bumps version of the thecodingmachine/safe dependency.


composer.json
```diff
@@ -33,7 +33,7 @@
     "rector/rector": "^2.0",
     "symfony/polyfill-php81": "^1.23",
     "symfony/var-exporter": "^5 || ^6 || ^7",
-    "thecodingmachine/safe": "^1.3 || ^2"
```
spawnia, line 36: We need to allow `^1.3` for PHP 7.4 compatibility.

PR #1697, opened 2025-04-30
Remember initial types instead of lazy-loading

rudiedirkx:
Types that are given in schema config `'types'` don't have to be `'typeLoader'` loaded later:

```php
new Schema([
	'typeLoader' => function(string $name) {
		return GraphQLFactory::type(sprintf('App\Services\GraphQL\Schema\%sType', $name));
	},
	'query' => GraphQLFactory::type(ClubQueryType::class),
	'mutation' => GraphQLFactory::type(ClubMutationType::class),
	'types' => [
		GraphQLFactory::type(ReservationInterfaceType::class),
		GraphQLFactory::type(ClassReservationType::class),
		GraphQLFactory::type(CourtReservationType::class),
	],
]);
```

The 3 provided types don't have to go through the `typeLoader`. In my case the type loader doesn't work for those 3, but everything else should be lazy loader by `typeLoader`.

This might break something else, but I'm curious to see all the tests.


rudiedirkx:
Current workaround:

```php
$types = [
	'Datetime' => DatetimeType::class,
	'ReservationInterface' => ReservationInterfaceType::class,
];
return new Schema([
	'typeLoader' => function(string $name) use ($types) {
		$typeClass = $types[$name] ?? sprintf('App\Services\GraphQL\Schema\%sType', $name);
		return GraphQLFactory::type($typeClass); // @phpstan-ignore argument.templateType
	},
	'query' => GraphQLFactory::type(ClubQueryType::class),
	'mutation' => GraphQLFactory::type(ClubMutationType::class),
]);
```

but I rather put those `$types` types in `Schema`'s `'types'`.

src/Type/Schema.php
```diff
@@ -99,6 +99,10 @@ public function __construct($config)
         $this->extensionASTNodes = $config->extensionASTNodes;
 
         $this->config = $config;
+
+        foreach ($this->config->types as $type) {
```
spawnia, line 103: Users may specify `types` in a multitude of ways, its type is `iterable<Type&NamedType>|(callable(): iterable<Type&NamedType>)|iterable<(callable(): Type&NamedType)>|(callable(): iterable<(callable(): Type&NamedType)>)`. The code you wrote does not work with some of those options.

rudiedirkx:
I could fix this by maybe calling `->types` and `$type`, BUT that negates the lazyness of the callables, so maybe my approach is wrong? Maybe Schema should only do this if `types` AND `typeLoader` are given? Or not, since without `typeLoader`, the entire schema is loaded anyway. Maybe my 'workaround' is the right way to do conservative type-loading? You tell me :)

rudiedirkx:
I don't like this fix at all. Copied from `getTypeMap`. That's a lot of callback and `@var` and `assert()`. Maybe my approach is just wrong. How to pre-load `types` better?

PR #1707, opened 2025-05-28
Fix known type names validation for fragments with experimental variables

simstern:
This fixes building a schema with `experimentalFragmentVariables`.

Currenty, the `DocumentValidator` returns `Unknown type "String"`.

Example:
```
$typeDefs = '
            type Hello {
                world(phrase: String): String
            }
            fragment hello($phrase: String = "world") on Hello {
                world(phrase: $phrase)
            }
            type RootQuery {
                hello: Hello
            }
            schema {
                query: RootQuery
            }
        ';

$astDocument = Parser::parse($typeDefs, ['experimentalFragmentVariables' => true]);

return BuildSchema::buildAST($astDocument);
```

Fragments should also implement `TypeSystemDefinitionNode` to have the type names validated against the `Type::BUILT_IN_TYPE_NAMES` by the `GraphQL\Validator\Rules\KnownTypeNames` rule:

```
$isSDL = $definitionNode instanceof TypeSystemDefinitionNode || $definitionNode instanceof TypeSystemExtensionNode;
if ($isSDL && in_array($typeName, Type::BUILT_IN_TYPE_NAMES, true)) {
    return;
}
```

Sorry, i struggled to add a test for this case with the current mock data in the `KnownTypeNamesTest`. Please advise if it is required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
+- KnownTypeNames validation for fragments with variables (experimentalFragmentVariables option)
+
 ## v15.20.0
 
 ### Added
--- a/src/Language/AST/FragmentDefinitionNode.php
+++ b/src/Language/AST/FragmentDefinitionNode.php
@@ -2,7 +2,7 @@
 
 namespace GraphQL\Language\AST;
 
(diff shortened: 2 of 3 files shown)


spawnia:
Please add a test for this.

simstern:
@spawnia Thanks for reviewing. 

I added a test now, but then realised that even with my change, the test is successful.

I was upgrading another package that is built on top of this. As the tests were working prior v15.x, I blindly trusted the example I posted above. However, as I understand fragments should not be part of type definitions and therefore it's probably valid, that validation fails for this case.

PR #1709, opened 2025-06-06
Recommend using assumeValidSDL when caching a parsed schema

### LINKED ISSUE #1708: Recommend using assumeValidSDL when caching a parsed schema?
Should [the documentation for caching the parsed schema](https://webonyx.github.io/graphql-php/schema-definition-language/#performance-considerations) also recommend passing `assumeValidSDL` to avoid re-validating the schema on every request?

We do some manual operations to merge multiple `.graphql` files and modifications based on schema-level directives (authorization control based on session type) so we wouldn't want to opt out entirely. Instead we moved to manually call `DocumentValidator::assertValidSDL` as part of the build process and then pass `assumeValidSDL` to avoid it on every request. We saw a hefty reduction in resource usage after doing so.

```php
$cacheFilename = 'cached_schema.php';

if (!file_exists($cacheFilename)) {
    $document = Parser::parse(file_get_contents('./schema.graphql'));
    DocumentValidator::assertValidSDL($document);
    file_put_contents($cacheFilename, "<?php\nreturn " . var_export(AST::toArray($document), true) . ";\n");
} else {
    $document = AST::fromArray(require $cacheFilename); // fromArray() is a lazy operation as well
}

$typeConfigDecorator = function () {};
$schema = BuildSchema::build($document, $typeConfigDecorator, ['assumeValidSDL' => true]);
```

Do you see any reason why this could be a problem, or could this be made the default recommendation in the documentation?

zimzat:
Doing the check once manually during the build step mitigates the potentially hefty cost of doing it on every request.

resolves #1708

### DIFF
--- a/docs/schema-definition-language.md
+++ b/docs/schema-definition-language.md
@@ -65,6 +65,7 @@ It is recommended to cache the intermediate parsed representation of the schema 
 
 ```php
 use GraphQL\Language\Parser;
+use GraphQL\Validator\DocumentValidator;
 use GraphQL\Utils\BuildSchema;
 use GraphQL\Utils\AST;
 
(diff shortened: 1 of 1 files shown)


spawnia:
Thank you, nicely done.

PR #1718, opened 2025-07-05
feat: remove SchemaConfig::assertMaybeLazyObjectType()

simPod:
The methods are typed so I was wondering that we don't need the runtime validation. Also can remove `@throws InvariantViolation`.

WDYT?


simPod:
We actually have some required validation there e.g. see test testRejectsASchemaWhoseRootTypeIsAnInputType

So it cannot be simply removed right now.

Issue #1493, opened 2023-12-07
Feature request: disable return/resolved values validation


Warxcell:
What do you think about this one? The use-case is to gain a little bit of performance, by disabling resolver's returned value validation - and just trusting it.

No idea how much will gain - just throwing it in the wild and see what people think. :)

spawnia:
My instinct says that the additional check for whichever setting controls this behaviour might eat up some of the benefits afforded by it. Making this as fast as possible would require having a code path that does no checks at all, but also does not checks if that behaviour is enabled. This would require duplicating/rewriting substantial parts of the executor, as well as possibly swapping the default field resolver and other parts that have sanity/safety checks.

Feel free to explore this in a pull request, along with benchmarks to test the effectiveness. You can also use a custom executor in your own project, see https://github.com/webonyx/graphql-php/blob/3ae418579b555253eeb3afbf2ad69dc1a770867f/src/Executor/Executor.php#L78. I don't think this is going to be a worthwhile endeavours and am thus closing this issue, but I am also open to being convinced otherwise with working code and the benchmarks to prove it.

spawnia:
FTR?

Warxcell:
btw https://github.com/zalando-incubator/graphql-jit have thing called

```
disableLeafSerialization {boolean, default: false} - disables leaf node serializers. The serializers validate the content of the field at runtime so this option should only be set to true if there are strong assurances that the values are valid.
```

I will try to make test in few days to see what's the gain.

Warxcell:
@spawnia another thing that could improve performance is skipping

```
            $validationErrors = DocumentValidator::validate($schema, $documentNode, $validationRules);

            if ($validationErrors !== []) {
                return $promiseAdapter->createFulfilled(
                    new ExecutionResult(null, $validationErrors)
                );
            }
```

in case if document is cached. If document is cached - it was already validated somewhere in the past.

Warxcell:
before
![image](https://github.com/webonyx/graphql-php/assets/3340882/1e9a1366-08b0-4848-b320-54c2b6c69f47)

after:
![image](https://github.com/webonyx/graphql-php/assets/3340882/cc3c2809-7feb-45f9-bba6-c7a3a57f0729)

basically 100-200ms improvement on one specific query on our side.


```
 $cacheItem = $this->queryCache->getItem(md5($op->query));

            if ($cacheItem->isHit()) {
                $documentNode = AST::fromArray($cacheItem->get());
            } else {
[...]

PR #1730, opened 2025-07-11
Cache result of validation

shmax:
Heyo, I remember there was some [discussion about possibly disabling document validation checks](https://github.com/webonyx/graphql-php/issues/1493) a while back. I finally got around to looking into it and realized to my horror that it was indeed gobbling up a lot of resources in my app. 

I don't know if you folks ever really agreed on a plan, but I thought I'd float this simple caching solution that leaves it up to the user. 

Let me know what you think. If you like the general direction I can do a little polish and write better tests.


composer.json
```diff
@@ -11,7 +11,8 @@
   "require": {
     "php": "^7.4 || ^8",
     "ext-json": "*",
-    "ext-mbstring": "*"
+    "ext-mbstring": "*",
+    "psr/simple-cache": "^1.0"
```
spawnia, line 15: I would like to keep this library dependency-free. Let's define our own interface or implement this in terms of a callable instead.

spawnia (changes requested):

composer.json
shmax: Sure, no prob. Will update.

shmax: Updated.

src/Validator/ValidationCache.php
spawnia, line 1: I like this solution!

src/Validator/DocumentValidator.php
```diff
@@ -99,16 +99,22 @@ public static function validate(
         Schema $schema,
         DocumentNode $ast,
         ?array $rules = null,
-        ?TypeInfo $typeInfo = null
+        ?TypeInfo $typeInfo = null,
+        ?ValidationCache $cache = null
     ): array {
-        $rules ??= static::allRules();
+        if (isset($cache)) {
+            $cached = $cache->isValidated($schema, $ast);
```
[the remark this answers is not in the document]
shmax: I opted to cache just the custom rules that are passed in. The idea is that the default rules are static and can thus be assumed to be part of the normal validation process.

spawnia (commented):
I think we are good with the implementation in its current form. Can you add documentation to `docs/executing-queries.md`? I think a section after `Custom Validation Rules` would work fine.

src/Validator/DocumentValidator.php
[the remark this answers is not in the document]
shmax: I was sort of tempted to add a `buildKey` stubby to the interface so we'd have somewhere to discuss this kind of thing, but I opted instead to add some notes to the class header. Let me know what ya think.

spawnia (commented):
There are some tricky details to figure out when implementing this. I really value putting in the time now to properly document them now to validate the design. I am going to try implenting the validation cache in https://github.com/nuwave/lighthouse and use that to battle-test it in one of my projects to see if anything else comes up.

spawnia:
I just found that Lighthouse already implements validation caching, see https://github.com/nuwave/lighthouse/pull/2603. In general, the approach to hashing to schema and hashing the query string is the same, but there are some differences:
1. Lighthouse offers special consideration for the `QueryComplexity` rule, which requires variables to work. Generally, validation rules do not have access to variable values, but this one has special treatment even in `webonyx/graphql-php`. The solution in Lighthouse splits validation in two phases, one where the cacheable rules (everything except `QueryComplexity`) are run/cached, then `QueryComplexity` maybe runs afterwards.
2. Lighthouse has no considerations for the used library version of `webonyx/graphql-php` or `nuwave/lighthouse` itself, but probably should consider that.
3. Lighthouse does not allow dynamically passing `$rules` ad-hoc, but users may overwrite the interface `ProvidesCacheableValidationRules` and change which validation rules are used. However, I found that using just the keys or class names of the used validation rules array is not sufficient. The validation rules for security such as `QueryDepth` are configurable and behave differently based on the configuration.

To resolve 1., perhaps we should add multi-phase validation to this library or some kind of special treatment for `QueryComplexity`.
Points 2. and 3. seem like something where Lighthouse is currently lacking but could be improved.

I am going to try adding a pull request to Lighthouse soon that implements its validation cache feature using the facilities offered through this pull request and perhaps extend it where its lacking.

src/Validator/ValidationCache.php
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Some things to keep in mind when generating keys:
+ * - PHP's `serialize` method is fast, but can't handle certain structures such as closures.
+ * - If your `schema` includes closures or is too large or complex to serialize,
+ *   consider using a build-time version number or environment-based fingerprint instead.
+ * - Keep in mind that there are internal `rules` that are applied in addition to any you pass in,
+ *   and it's possible these may shift or expand as the library evolves, so it might make sense
+ *   to include the library version number in your keys.
+ *
+ * @see PsrValidationCacheAdapter for a simple reference implementation.
```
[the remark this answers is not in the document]
shmax: I figure devs will be savvy enough to either clone the repo or look on github, but I can remove it for now if you like.

shmax:
Not sure I follow. I don't use Lighthouse, and I'm not familiar with it. What does it have to do with what I'm doing in this PR?

spawnia:
> Not sure I follow. I don't use Lighthouse, and I'm not familiar with it. What does it have to do with what I'm doing in this PR?

Lighthouse is built atop this library. It already implements a validation cache that has proven to be effective. I would like to take the lessons from the implementation effort we made there to inform the implementation in this library. This is to ensure that I can replace the custom implementation in Lighthouse with what we provide here, and we are not missing anything.

shmax:
I see. Lighthouse implements its own high-level query flow, and uses `DocumentValidator:validate` directly. In the meantime, I modified my own code to just disable validation on production (I use persisted queries), so I'm not blocked.

I'll leave you to it.

shmax:
I see you're working on stuff around this, which is great. Holler if I can be of any help.

shmax:
@spawnia are you still planning on merging this at some point? Did you want me to address all these copilot comments?

spawnia:
I still see it as a useful feature addition, but I don't want to make promises that I will merge. I did a big mechanical sweep over all open PRs (merge master, fix mechanical issues, request automated reviews) to try and refine them to a point where my judgement is needed again.

shmax:
Well, let me know if you decide to move forward (I don't personally need it, as I'm halfway through migrating my entire codebase to Node.js, anyway).

PR #1737, opened 2025-07-24
Make querying `isOneOf` optional in Introspection query

spawnia:
Including it by default makes introspection fail on servers that do not support it:

```
Cannot query field "isOneOf" on type "__Type".
```

Thus, I added a configuration option `typeIsOneOf` which works analogous to `directiveIsRepeatable`.


spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.21.3. @ruudk @jasonbahl

PR #1747, opened 2025-08-12
Remove private constructor for Printer

ruudk:
The Printer is not stateful. So enforcing a singleton serves no purpose (anymore?). 

We can easily use `new Printer()->printAST($node)`.


spawnia:
Why instantiate the Printer in the first place?

ruudk:
Well, it is instantiated, and then statically cached. To me, there is no point in doing this. What was the reason for that?

spawnia:
I would rather make the class completely static then to invite superfluous instantiations.

ruudk:
Alright, not worth it as it's a BC break.

PR #1748, opened 2025-08-12
Do not escape forward slashes in Printer

ruudk:


spawnia:
Thank you @ruudk, released with https://github.com/webonyx/graphql-php/releases/tag/v15.22.2.

PR #1749, opened 2025-08-13
Make class Printer fully static

spawnia:
Follow up to https://github.com/webonyx/graphql-php/pull/1747.

@ruudk I believe there is no real breaking change here. Yes, users who extended the class may have some breakage, but otherwise not. Due to the protected constructor, it was not possible to obtain an instance of Printer, so `public function printAST` could not be called from anywhere. The only public API was `Printer::doPrint` anyways, which remains unchanged, other than a more accurate PhpDoc.

### DIFF
--- a/docs/class-reference.md
+++ b/docs/class-reference.md
@@ -1260,6 +1260,8 @@ $printed = GraphQL\Language\Printer::doPrint($ast);
  *
  * Handles both executable definitions and schema definitions.
  *
+ * @throws \JsonException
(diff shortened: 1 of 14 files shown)


ruudk:
Oh wow, I didn't realize that. Thanks for improving it 🎉

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.22.3.

PR #1751, opened 2025-08-15
Allow disabling query short form in Printer

ruudk:
In our project, I'd like to print operations explicitly, disabling the short form.

```graphql
query {
    id
    name
}
```

instead of

```graphql
{
    id
    name
}
```

I guess this is the downside of having this class static. I now have to pass this options along. Ideally, this would have been a property on the Printer object.

### DIFF
--- a/docs/class-reference.md
+++ b/docs/class-reference.md
@@ -1250,6 +1250,10 @@ $ast = GraphQL\Language\Parser::parse($query);
 $printed = GraphQL\Language\Printer::doPrint($ast);
 ```
 
+@phpstan-type Options array{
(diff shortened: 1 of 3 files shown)


spawnia:
I don't like the extra complexity and added choice. I think it is a slippery slope to go down. Once we start having any stylistic option at all, why not add an option for always adding braces after `query` as in `query () { ...`? How about an option for trailing commas? Making variables multiline?

Thanks for your continued effort on this library, but for the sake of maintainability, I am going to decline this one.

ruudk:
Yeah I can see that. Thanks.

PR #1752, opened 2025-08-15
Fix NodeVisitor return type

ruudk:
It also allows returning a Node.


ruudk:
@spawnia Finally got it green 😅

spawnia:
Thank you, released with https://github.com/webonyx/graphql-php/releases/tag/v15.22.4.

PR #1754, opened 2025-08-17
Allow replacing a Node with a NodeList in Visitor

ruudk:
The first 2 commits are actually from this branch: https://github.com/webonyx/graphql-php/pull/1752

With this change, it's now possible to do something like this:

```php
$editedAst = Visitor::visit(
    $ast,
    [
        NodeKind::INLINE_FRAGMENT => function ($node) {
            return $node->selectionSet->selections;
        },
    ]
);
```

It replaces an inline fragment (Node) with the selections (NodeList).

This makes it possible to create specific optimizer Visitors that cleanup operations.

For example, remove useless inline fragments on the same type:
```php
$wrapped = Visitor::visitWithTypeInfo($this->typeInfo, [
    NodeKind::INLINE_FRAGMENT => [
        'leave' => function (Node $node) : ?NodeList {
            Assert::isInstanceOf($node, InlineFragmentNode::class);

            $type = Type::getNamedType($this->typeInfo->getParentType());

            if ( ! $type instanceof \GraphQL\Type\Definition\ObjectType) {
                return null;
            }

            if ($type->name() !== $node->typeCondition?->name->value) {
                return null;
            }

            return $node->selectionSet->selections;
        },
    ],
]);

return Visitor::visit($node, $wrapped);
```

Or inline fragments that don't have a type selection and also don't use directives.

### DIFF
--- a/src/Language/Visitor.php
+++ b/src/Language/Visitor.php
@@ -218,21 +218,13 @@ class Visitor
                             ++$editOffset;
                         } elseif ($node instanceof NodeList) {
                             if ($editValue instanceof NodeList) {
-                                // Replace a single node with multiple nodes from a NodeList
-                                // We need to splice the NodeList's contents into the parent
-                                $replacementNodes = [];
-                                foreach ($editValue as $replacementNode) {
-                                    $replacementNodes[] = $replacementNode;
-                                }
-                                // Remove the original node and insert all replacement nodes
-                                $node->splice($editKey, 1, $replacementNodes);
-                                // Adjust the offset for subsequent edits
-                                $editOffset -= count($replacementNodes) - 1;
+                                $node->splice($editKey, 1, $editValue);
+                                $editOffset -= count($editValue) - 1;
                             } elseif ($editValue instanceof Node) {
                                 $node[$editKey] = $editValue;
                             } else {
-                                $notNode = Utils::printSafe($editValue);
-                                throw new \Exception("Can only add Node or NodeList to NodeList, got: {$notNode}.");
+                                $notNodeOrNodeList = Utils::printSafe($editValue);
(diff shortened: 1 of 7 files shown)


src/Language/Visitor.php
```diff
@@ -217,12 +217,23 @@ public static function visit(object $root, array $visitor, ?array $keyMap = null
...
-                            if (! $editValue instanceof Node) {
+                            if ($editValue instanceof NodeList) {
+                                // Replace a single node with multiple nodes from a NodeList
+                                // We need to splice the NodeList's contents into the parent
+                                $replacementNodes = [];
+                                foreach ($editValue as $replacementNode) {
+                                    $replacementNodes[] = $replacementNode;
+                                }
+                                // Remove the original node and insert all replacement nodes
+                                $node->splice($editKey, 1, $replacementNodes);
```
spawnia, line 228: Let's make `NodeList::splice` more flexible and have it accept any `iterable` instead of `array`.

tests/Language/VisitorTest.php
```diff
@@ -1732,4 +1732,70 @@ public function testAncestorsForObjectTypeDefinition(): void
...
+                }
+                field4
+            }
+        ', ['noLocation' => true]);
+
+        $editedAst = Visitor::visit(
+            $ast,
+            [
+                NodeKind::INLINE_FRAGMENT => [
+                    'leave' => fn ($node) => $node->selectionSet->selections,
```
spawnia, line 1753: Let's add strict types here

ruudk:
@spawnia Thanks, applied your feedback. It's cleaner now!

ruudk:
Not sure if the CI is an issue?

spawnia:
Thank you, released with https://github.com/webonyx/graphql-php/releases/tag/v15.23.0.

PR #1756, opened 2025-08-19
Improve Type::getNamedType return type

ruudk:
Only when null is passed, it will return null.

Otherwise it will always be Type&NamedType.


spawnia:
Thank you, released with https://github.com/webonyx/graphql-php/releases/tag/v15.23.1.

PR #1757, opened 2025-08-20
Early return when passing NonNull to Type::nonNull

ruudk:
This improves developer experience. If the intent was non null, and it already is non null, just return it.


spawnia:
Thanks, released with https://github.com/webonyx/graphql-php/releases/tag/v15.24.0.

PR #1758, opened 2025-08-22
chore: do not distribute Makefile

simPod:


ruudk (approved):

ruudk:
Should we also exclude docs, CHANGELOG.md and UPGRADE.md?

simPod:
I was thinking they may be intentionally included so I'm not sure about
those.

ruudk:
FWIW: I rarely see these type of things included in other projects. It's only extra stuff that gets pulled and pushed over and over again 🥲

spawnia:
> Should we also exclude docs, CHANGELOG.md and UPGRADE.md?

Yeah, I think we probably should. There is some convenience to having them right there in `vendor`, but it is probably very slim. Most users are going to head to the GitHub or docs page to find out about that stuff.

mfn:
I for one very much value having files like changelog in vendor.

It's much easier/faster to check them in vendor then checking each projects source remotely

spawnia:
> I for one very much value having files like changelog in vendor.
> 
> It's much easier/faster to check them in vendor then checking each projects source remotely

Just the changelog?

mfn:
changelog/upgrade, those files most relevant when doing incremental updates (which is the majority of actions taken)

simPod:
BTW did you consider Renovate? It automates most of the stuff for you.

PR #1774, opened 2025-10-03
[RFC] Pass `$objectValue` by reference when calling `resolveType`

ruudk:
This allows the Interface or Union type resolver to modify the object value.

This is useful when you have a single entity that needs different representations. For example, when your data layer returns a generic `PetEntity` with a type discriminator, but your GraphQL schema has separate `Dog` and `Cat` types.

Example:

```php
$PetType = new InterfaceType([
    'name' => 'Pet',
    'resolveType' => static function (PetEntity &$objectValue): string {
        if ($objectValue->type === 'dog') {
            $objectValue = new Dog($objectValue->name, $objectValue->woofs);
            return 'Dog';
        }

        $objectValue = new Cat($objectValue->name, $objectValue->meows);
        return 'Cat';
    },
    'fields' => ['name' => ['type' => Type::string()]],
]);
```

Now field resolvers receive the properly typed Dog or Cat object instead of the generic PetEntity,
allowing for type-safe resolution without needing to transform objects before they reach the type
resolver.

Common use cases:
- Database polymorphism (single table with type column)
- External APIs returning generic objects with type discriminators

> [!IMPORTANT]
> Unfortunately, this breaks the InterfaceType and UnionType contract. So this can only be done in a new major
> But this serves as a RFC. Maybe there is a better non-breaking way to do this.

/cc @spawnia @simPod

### DIFF
--- a/src/Type/Definition/AbstractType.php
+++ b/src/Type/Definition/AbstractType.php
@@ -20,5 +20,5 @@ interface AbstractType
      *
      * @phpstan-return ResolveTypeReturn
      */
-    public function resolveType($objectValue, $context, ResolveInfo $info);
+    public function resolveType(&$objectValue, $context, ResolveInfo $info);
 }
--- a/src/Type/Definition/InterfaceType.php
+++ b/src/Type/Definition/InterfaceType.php
@@ -67,7 +67,7 @@ class InterfaceType extends Type implements AbstractType, OutputType, CompositeT
         return $type;
     }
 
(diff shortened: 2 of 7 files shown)


ruudk:
I think this is a better solution to the problem:
- https://github.com/webonyx/graphql-php/pull/1776

ruudk:
Closing this one, as the other is a better solution that doesn't break.

PR #1776, opened 2025-10-03
Introduce `resolveValue` method to Interface and Union types

ruudk:
> [!NOTE]
> This is an alternative (and I think better) approach to:
> - https://github.com/webonyx/graphql-php/pull/1774

This allows transforming the object value after type resolution.

This is useful when you have a single entity that needs different representations. For example, when your data layer returns a generic `PetEntity` with a type discriminator, but your GraphQL schema has separate `Dog` and `Cat` types.

Example:

```php
$PetType = new InterfaceType([
    'name' => 'Pet',
    'resolveType' => static function (PetEntity $objectValue): string {
        if ($objectValue->type === 'dog') {
            return 'Dog';
        }
        return 'Cat';
    },
    'resolveValue' => static function (PetEntity $objectValue) {
        if ($objectValue->type === 'dog') {
            return new Dog($objectValue->name, $objectValue->woofs);
        }
        return new Cat($objectValue->name, $objectValue->meows);
    },
    'fields' => ['name' => ['type' => Type::string()]],
]);
```

Now field resolvers receive the properly typed Dog or Cat object instead of the generic PetEntity,
allowing for type-safe resolution without needing to transform objects before they reach the type
resolver.

Common use cases:
- Database polymorphism (single table with type column)
- External APIs returning generic objects with type discriminators

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -120,19 +120,10 @@ class ReferenceExecutor implements ExecutorImplementation
         );
 
         if (is_array($exeContext)) {
-            return new class($promiseAdapter->createFulfilled(new ExecutionResult(null, $exeContext))) implements ExecutorImplementation {
-                private Promise $result;
+            $executionResult = new ExecutionResult(null, $exeContext);
+            $fulfilledPromise = $promiseAdapter->createFulfilled($executionResult);
 
-                public function __construct(Promise $result)
-                {
-                    $this->result = $result;
-                }
-
-                public function doExecute(): Promise
-                {
-                    return $this->result;
-                }
-            };
+            return new PromiseExecutor($fulfilledPromise);
         }
 
         return new static($exeContext);
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -1092,6 +1092,7 @@ class ReferenceExecutor implements ExecutorImplementation
         $contextValue
     ) {
         $typeCandidate = $returnType->resolveType($result, $contextValue, $info);
+        $result = $returnType->resolveValue($result, $contextValue, $info);
 
         if ($typeCandidate === null) {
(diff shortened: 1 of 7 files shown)


ruudk:
Instead of calling it `resolveValue` we could also go with `transformValue` indicating that this is about transforming the value from A to B.

tests/Executor/TestClasses/PetEntity.php
```diff
@@ -0,0 +1,21 @@
...
+namespace GraphQL\Tests\Executor\TestClasses;
+
+final class PetEntity
+{
+    /** @var 'dog'|'cat' */
+    public string $type;
+
+    public string $name;
+
+    public bool $woofs;
```
spawnia, line 12: `vocalizes`

src/Executor/ReferenceExecutor.php
```diff
@@ -1092,6 +1092,7 @@ protected function completeAbstractValue(
         $contextValue
     ) {
         $typeCandidate = $returnType->resolveType($result, $contextValue, $info);
+        $result = $returnType->resolveValue($result, $contextValue, $info);
```
spawnia, line 1095: What's going on with the pass-by-reference of `$result`? What happens if we remove `&`?

Also, what is going on with its type - we seem to know it is `array<mixed>` here, but is it really? What if me make it a hard type hint?

ruudk: This seems like a bigger issue to investigate, ideally, separately from this PR. 

If we ship this as is, I'll make sure to follow up on it.

ruudk: the existing resolveType function also hints for `mixed`: https://github.com/ruudk/graphql-php/blob/a753f1c56df97dfe105c1c2ac8abc6d72af2af28/src/Type/Definition/AbstractType.php#L14-L24

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.25.0.

PR #1780, opened 2025-10-08
Remove pass-by-reference for `$result` in `ReferenceExecutor`

spawnia:
Follows up on https://github.com/webonyx/graphql-php/pull/1776#discussion_r2411090574

### DIFF
--- a/src/Executor/ReferenceExecutor.php
+++ b/src/Executor/ReferenceExecutor.php
@@ -864,7 +864,7 @@ class ReferenceExecutor implements ExecutorImplementation
         ResolveInfo $info,
         array $path,
         array $unaliasedPath,
-        &$result,
+        $result,
         $contextValue
     ) {
         // If result is an Error, throw a located error.
@@ -1007,7 +1007,7 @@ class ReferenceExecutor implements ExecutorImplementation
         ResolveInfo $info,
         array $path,
         array $unaliasedPath,
-        iterable &$results,
+        iterable $results,
         $contextValue
     ) {
         $itemType = $returnType->getWrappedType();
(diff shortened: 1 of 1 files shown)


spawnia:
Tests pass, I don't see how this change could be problematic.

ruudk:
Since the `ReferenceExecutor` is not final, one could be extending it.

For example:
```php
final class TestReferenceExecutor extends ReferenceExecutor {
    protected function completeAbstractValue(
        AbstractType $returnType,
        \ArrayObject $fieldNodes,
        ResolveInfo $info,
        array $path,
        array $unaliasedPath,
        &$result,
        $contextValue)
    {
        return parent::completeAbstractValue(
            $returnType,
            $fieldNodes,
            $info,
            $path,
            $unaliasedPath,
            $result,
            $contextValue,
        );
    }
}
```

This would fail now:
```
Declaration should be compatible with ReferenceExecutor->completeAbstractValue(returnType: \GraphQL\Type\Definition\AbstractType|\GraphQL\Type\Definition\Type&\GraphQL\Type\Definition\AbstractType, fieldNodes: \ArrayObject, info: \GraphQL\Type\Definition\ResolveInfo, path: int[]|string[], unaliasedPath: int[]|string[], result: mixed, contextValue: mixed) 
```

spawnia:
We do mark public elements of our API with `@api`, so I would categorize this as internal and non-breaking.

ruudk:
In that case, let's ship this! Way better indeed.

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.29.3.

PR #1781, opened 2025-10-08
Call `resolveValue` before calling `resolveType`

ruudk:
In hindsight it makes more sense to first resolve the value to the desired type before calling `resolveType`.

I figured this out while implementing this new feature in our project. 

Related to #1776


ruudk:
Haha, I actually also did do that initially, but figured you wouldn't like that big diff so removed it and kept it simpler.

But way better, to have the same order as it's called. Thanks!

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.25.1.

PR #1787, opened 2025-10-25
Document the release process

spawnia:


spawnia:
After successful review, I plan to make the same change in https://github.com/nuwave/lighthouse/pull/2721.

ruudk:
I also see some projects use this https://github.com/changesets/changesets to automate this further.

spawnia:
I have considered using semantic-release, as it is done in https://github.com/mll-lab/php-utils. What do you think about that?

ruudk:
I've seen that too. Since you have experience with that one, let's pick that one. We only need to make sure that we have a CI action to enforce proper PR title.

ruudk:
We use this in one of our internal projects:

```
      - uses: amannn/action-semantic-pull-request@v4
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          types: |
            fix
            feat
            chore
```

https://github.com/amannn/action-semantic-pull-request

PR #1790, opened 2025-11-06
Optimize Deferred memory usage with incremental queue processing

### LINKED ISSUE #972: Using Deferred produces memory leak?

spawnia:
## Summary

This PR optimizes memory usage when using Deferred objects by implementing incremental queue processing. This reduces peak memory usage by ~70% when handling thousands of Deferred objects.

## Problem

Issue #972 reported excessive memory usage with Deferred objects. Investigation revealed:
- 4000 Deferred objects consumed ~54MB peak memory (6.75x baseline of 8MB)
- This was NOT a memory leak - memory was properly released between requests
- The issue was **peak memory usage during execution**

## Root Cause

The static `SplQueue` in `SyncPromise` accumulated ALL closures before processing. With 4000 Deferred objects, thousands of closures were queued simultaneously, causing high memory usage.

## Solution

Implemented incremental queue processing with a batch threshold:
- Added `QUEUE_BATCH_SIZE` constant (500 items)
- When queue exceeds threshold, immediately process a batch
- This keeps queue size bounded and reduces peak memory

## Results

**Peak memory usage:**
- Before optimization: ~54MB (6.75x baseline)
- After optimization: ~16MB (2x baseline)
- **Reduction: 70%**

## Testing

Added `testDeferredMemoryEfficiency()` test that:
- ✅ PASSES with optimization (16MB < 25MB threshold)
- ❌ FAILS without optimization (48MB > 25MB threshold)
- Prevents future regressions
- All 240 existing tests pass

## Verification

Tested against original reproduction case from https://github.com/Danbka/php-graphql-deffred showing identical improvements.

Fixes #972

### DIFF
--- a/phpstan/php-below-8.2.neon
+++ b/phpstan/php-below-8.2.neon
@@ -10,4 +10,4 @@ parameters:
             count: 2
             path: ../src/Utils/TypeInfo.php
 
-        - message: "~PHPDoc tag `@throws` with type .*Random\\RandomException is not subtype of Throwable~"
+        - message: "~PHPDoc tag .+ with type .*Random\\\\RandomException.* is not subtype of Throwable~"


zimzat:
The 500 limit is going to impact how downstream batching is handled. e.g. If I'm returning a list of 500 items and each item has two Deferred items on it then it's going to trigger the batch loading protocol to run four+ partial batches instead of just two full batches. The multiplicative nature of that would be worse for queries that return multiple lists of deferred batches (e.g. 2+ sets 100 items that have 6 defers would trigger 18 batches of 84, 84, and 32 items).

> We may offer an option to adjust this value in the future.

I would find this useful; I'd probably set it to 1000 or 2000. 16MB of memory usage is really tiny in comparison to the amount of memory that the application itself is likely to hold for the data being returned.

Unless I'm making some incorrect assumptions on how those mechanisms work together, in which case this isn't an actual problem? :shrug:

spawnia:
Thank you @zimzat and @shadowhand for your feedback. I am continuing work in https://github.com/webonyx/graphql-php/pull/1805 and will take your comments into consideration.

spawnia:
Followed up in https://github.com/webonyx/graphql-php/releases/tag/v15.29.0

PR #1797, opened 2025-11-23
Improve performance by disabling runtime value validation

mithilP007:
Commented out expensive runtime validation checks to improve query execution performance:
- Removed null check for non-nullable fields (lines 887-889)
- Removed is_iterable check for list types (lines 900-904)

Based on benchmarks showing 100-200ms improvements for queries.

Addresses issue #1493


spawnia:
Fine experiment, but I can not merge this pull request in this state.

Issue #1803, opened 2025-12-05
Getting null instead of a list of items for v15.27.0


joshbmarshall:
Hi @spawnia I am having issues since version 15.27.0 was released, and I confirmed going back to 15.26.0 is working.

I'm using the overblog/dataloader-php in the resolve functions. I'm not sure how I can make a minimal reproduction for it, as it is connecting to a database.

I have a query that pulls many subitems and so is processing a batch.

The query is something like:

```
{
  eventDays {
    event {
      eventDays {
         date
      }
    }
  }
}
```

Instead of a list of the nested eventDays being returned, it is now returning null.

I comment out:

https://github.com/webonyx/graphql-php/blob/3402c86315e5ae88837c4f090b4b3516a55e7059/src/Executor/Promise/Adapter/SyncPromise.php#L188

and I get the results again.

I have been made aware of this because my schema is Type::nonNull(Type::listOf(...)) (which it should be)

Where an event has days, I'm getting a list of days in a range and then all the events for that range, and the details of each one inside that. I understand this might not be ideal but it's making it hit the 500 limit to process the batch.

The one that's getting the null list is the nested one, the first 40 are ok, then after that there is a null result.

Restoring the line 188 code above, and changing the batch size to 400, the null item is at position 20 (ie earlier).

Changing batch size to 530, there is no errors and no null items. Comparing response size, it's the same number of bytes as v15.26.0 was responding. So looks like it is fetching everything from the database correctly.

Putting some debug in, the queue size escalates to over 1000, indicating that the processBatch() is already processing an item.

I checked memory_get_peak_usage after the graphql query had executed, and the difference is minimal:

v15.26.0 - 50609984
v15.27.0 - 50136648 (with batch size set to 530 so the query works)

This is probably due to processBatch() being locked?

Additionally, I see around a 5% reduction in performance 15.26.0 to 15.27.0 - could that be due to the constant counting of the queue?

spawnia:
I have decided to revert the `Deferred` batching for now, as this is a serious bug and I don't have the time to look into this in detail right now. https://github.com/webonyx/graphql-php/releases/tag/v15.27.1 fixes the issue for now.

spawnia:
https://github.com/nuwave/lighthouse/issues/2728#issuecomment-3616095491 was also reportedly an error that came from https://github.com/webonyx/graphql-php/releases/tag/v15.27.0.

spawnia:
[...]

PR #1805, opened 2025-12-05
Optimize `Deferred` execution

spawnia:
Resolves https://github.com/webonyx/graphql-php/issues/1803

Follow up to https://github.com/webonyx/graphql-php/pull/1790

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "amphp/http-server": "^2.1",
     "dms/phpunit-arraysubset-asserts": "dev-master",
     "ergebnis/composer-normalize": "^2.28",
-    "friendsofphp/php-cs-fixer": "3.92.2",
+    "friendsofphp/php-cs-fixer": "3.92.3",
     "mll-lab/php-cs-fixer-config": "5.11.0",
     "nyholm/psr7": "^1.5",
     "phpbench/phpbench": "^1.2",


spawnia:
## Summary

This PR optimizes memory usage for deferred execution while preserving the DataLoader pattern that was broken in #1790.

### Problem
PR #1790 attempted to optimize memory but broke DataLoader batching - batch loads executed incrementally before all IDs were collected, causing null values with 600+ items.

### Solution
Reduces memory usage by replacing closure-based queue storage with direct object/array storage:

1. **Store executor on promise instance** - `$executor` property instead of closure context
2. **Queue promise objects directly** - `enqueue($this)` instead of `enqueue(function() {...})`
3. **Use arrays for waiting handlers** - `[$promise, $onFulfilled, $onRejected, $state, $result]` instead of closures
4. **Explicit cleanup** - `$task->executor = null` and `unset($task)` enable GC during long queues

### Tests Added
- `testDataLoaderPatternWithManyDeferredObjects()` - 600 items, verifies single batch load
- `testNestedDataLoaderPattern()` - 750 items, reproduces #1803 scenario

### Verification
- [x] Tests pass with optimization
- [x] Tests fail when #1790 changes are cherry-picked (confirms they catch the regression)
- [x] DataLoader pattern preserved - all IDs collected before first batch executes

spawnia:
@shadowhand @zimzat @joshbmarshall Can you give this a review?

joshbmarshall:
I loaded it up on the system that was having problems with v15.27.0

It still has issues with v15.27.0 tag, but this is working for me at 1954f81fe437d910c0283307d8f17d125b4a3a11

For my query, the memory usage is identical for this pull request as for master branch. But I think it's because I have nested resolvers:

event date > event > event dates

I think it's waiting for all the event dates to be queued before resolving, which is how I'd expect it to be. I don't think that can be managed!

zimzat:
@spawnia This looks promising; I don't see any issues with load order or breaking batches.

Below are some extra thoughts based on my own trial-error:

The different signatures in the queue feels concerning. I had started experimenting with the idea of switching out the anonymous functions with invokable classes. Your implementation of `$this->executor`  wins out over the cost of instantiating a new class, while a new class for the waiting queue item as a replacement for the array is a little more efficient there, so the two patterns combined reduces peak usage slightly further.

Using `SplFixedArray` for `$this->waiting[] = SplFixedArray` would also get half way to the same peak memory reduction.

https://gist.github.com/zimzat/310e9121225ef18640fad2f4ab675ad4

spawnia:
@zimzat I did follow up on your suggestion to use `SplFixedArray`, seems like a useful optimization.

Can you elaborate on what specifically you find concerning? The methods in `SyncPromise` are not marked with `@api`, so they are not supposed to be part of the public API.

zimzat:
> Can you elaborate on what specifically you find concerning?

The implicit array shape versus an explicit type check of a named and structured class. It's not a big deal, especially here and internally to the class (like you've said, this isn't technically a public API). The `if (x) elseif (y) elseif (z)` felt verbose compared to the simplicity of the queue always being `callable()`. ¯\\\_(ツ)\_/¯

spawnia:
@zimzat I did another round which should hopefully alleviate your concern. Benchmarks look good too, what do you think?

zimzat:
@spawnia That looks nice; a simple signature, self-contained handling, and still a massive drop in memory usage in the simple many deferred (no then) scenario. Thanks!

src/Executor/Promise/Adapter/SyncPromise.php
```diff
@@ -143,34 +179,54 @@ private function enqueueWaitingPromises(): void
...
+
+        // Clear references to allow GC
+        $task->waitingOnFulfilled = null;
+        $task->waitingOnRejected = null;
+        $task->waitingState = null;
+        $task->waitingResult = null;
+
+        try {
+            if ($state === self::FULFILLED) {
+                $task->resolve($onFulfilled === null ? $result : $onFulfilled($result));
```
shadowhand: I think this would be easier to understand as:


```suggestion
                if ($onFulfilled) {
                    $result = $onFulfilled($result);
                }
                $task->resolve($result);
```

More literal code is fine here, since PHP will optimize the loop when it compiles.

spawnia:
## Memory Benchmark Results

Ran memory benchmarks comparing `master` against the optimized implementation using a [benchmark script adapted from @zimzat's test](https://gist.github.com/zimzat/310e9121225ef18640fad2f4ab675ad4#file-test-php).

### Results

| Metric | Master | Optimized | Change |
|--------|--------|-----------|--------|
| Single Deferred | 888 bytes | 144 bytes | **-84%** |
| Nested Deferred (with `then`) | 1,272 bytes | 528 bytes | **-58%** |
| 1,000 Deferreds | 896 KB | 144 KB | **-84%** |
| 5,000 nested Deferreds | 6.5 MB | 2.8 MB | **-57%** |
| Peak memory | 13.9 MB | 9.3 MB | **-33%** |
| Final memory | 517 KB | 263 KB | **-49%** |

### Key Optimization: Garbage Collection

During benchmarking, we found that `RootSyncPromise::runQueuedTask()` must null out `$executor` after invocation to allow immediate garbage collection of the closure and its captured variables:

```php
public function runQueuedTask(): void
{
    // Clear reference to allow immediate garbage collection
    $executor = $this->executor;
    $this->executor = null;
    // ...
}
```

Without this, closures stay alive until the promise object is freed, causing significant memory accumulation when many deferred operations are queued. This pattern is now documented in the code with an explanatory comment.

spawnia:
## Benchmark Results Summary

Ran benchmarks comparing this branch against `master` on PHP 8.3.6.

### Key Changes
- Promise state changed from strings (`"pending"`, `"fulfilled"`, `"rejected"`) to integers (`0`, `1`, `2`)
- Extracted `SyncPromiseQueue` to a dedicated class
- Moved `$executor` handling from `SyncPromise` constructor to `Deferred`
- Added comprehensive `DeferredBench` for tracking deferred execution performance

### Memory Improvements

| Benchmark | Master | Branch | Change |
|-----------|--------|--------|--------|
| benchManyNestedDeferreds | 19.39 MB | 17.67 MB | **-8.9%** |
| bench1000Chains | 10.08 MB | 9.72 MB | **-3.6%** |
| benchManyDeferreds | 6.45 MB | 6.09 MB | **-5.6%** |

### Execution Time

| Benchmark | Master | Branch | Change |
|-----------|--------|--------|--------|
| benchManyNestedDeferreds | 17,801 μs | 16,651 μs | **-6.5%** |
| bench1000Chains | 4,042 μs | 3,919 μs | **-3.0%** |
| benchManyDeferreds | 673 μs | 922 μs | +37% |
| benchChain100 | 97 μs | 145 μs | +50% |
| benchChain5 | 6.0 μs | 10.1 μs | +68% |
| benchSingleDeferred | 0.9 μs | 1.2 μs | +38% |

### Summary

The branch achieves **significant memory savings** (up to 9%) for heavy deferred workloads while also **improving execution time** for the most demanding benchmarks (`benchManyNestedDeferreds`, `bench1000Chains`).

The microbenchmarks with simple promise chains show some CPU overhead, likely due to structural changes that benefit memory at a small cost to simple cases. For real-world GraphQL queries with many nested deferreds, the branch performs better overall.

spawnia:
Next steps for me:
- try this in my real-world projects
- try this in nuwave/lighthouse

shmax:
Lookin' good!

mfn:
I tested this in a private project with a couple of thousand GraphQL tests, no issues. Can't say anything about real life impact though.

spawnia:
Lighthouse and a large project of mine work.

spawnia:
Released with https://github.com/webonyx/graphql-php/releases/tag/v15.29.0.

joshbmarshall:
very early at the moment, but a number of my sites have gone down because of this update. I've reverted back to v15.28.0 for now which gets them going. I'll have to get you details but this is causing issues

joshbmarshall:
My sentry is reporting:

Call to undefined method GraphQL\Deferred::runQueue()

vendor/overblog/dataloader-php/lib/promise-adapter/src/Adapter/WebonyxGraphQLSyncPromiseAdapter.php in Overblog\PromiseAdapter\Adapter\WebonyxGraphQLSyncPromiseAdapter::await at line 115 

```php
    public function await($promise = null, $unwrap = false)
    {
        if (null === $promise) {
            Deferred::runQueue();
            SyncPromise::runQueue();
            $this->cancellers = [];
            return null;
        }
        $promiseAdapter = $this->getWebonyxPromiseAdapter();
```

PR #1813, opened 2025-12-10
Optimize Lexer::readName

kasparsklavins:
Finding the name's length using [`strspn(...)`](https://www.php.net/manual/en/function.strspn.php), and doing a single string copy with [`substr(...)`](https://www.php.net/manual/en/function.substr.php) is faster than looping through one character at a time. As [names must be [A-Za-z0-9_]+](https://spec.graphql.org/September2025/#Name), it's safe to skip UTF8 checks.

# Benchmark

Benchmark for introspection query shows promising results on my modest potato running php 8.3 with opcache enabled

<details>
<summary>bench.php</summary>

```php
use GraphQL\Language\Lexer;
use GraphQL\Language\Source;
use GraphQL\Language\Token;

require __DIR__ . '/vendor/autoload.php';

$source = new Source(Introspection::getIntrospectionQuery());

$run = function() use ($source) {
    $lexer = new Lexer($source);
    do {
        $token = $lexer->advance();
    } while ($token->kind !== Token::EOF);
};

$run(); // warmup
$run(); // warmup

$times = [];
for ($i = 0; $i < 100; ++$i) {
    $start = hrtime(true);
    $run();
    $times[] = hrtime(true) - $start;
}

var_dump([
    'min' => min($times) / 1_000,
    'max' => max($times) / 1_000,
    'avg' => array_sum($times) / count($times) / 1_000,
]);
```
</details>

|  | Min | Max | Average |
| --- | --- | --- | --- |
| Before | 243.5μs | 658.6μss | 268.1μs |
| After | 168.5μs (-30%) | 492.2μs (-25%) | 188.4μs (-30%) |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ### Changed
 
-- Optimize `Lexer` name tokenization  https://github.com/webonyx/graphql-php/pull/1813
+- Optimize `Lexer` name tokenization https://github.com/webonyx/graphql-php/pull/1813
 
 ## v15.27.2
 


spawnia:
Thank you @kasparsklavins, released with https://github.com/webonyx/graphql-php/releases/tag/v15.28.0.

Issue #1820, opened 2025-12-19
BC Break: SyncPromise::runQueue() and SyncPromise::getQueue() removed in v15.29.0


grossmannmartin:
### Description

Version 15.29.0 introduced a breaking change by removing the public static methods SyncPromise::runQueue() and SyncPromise::getQueue() from GraphQL\Executor\Promise\Adapter\SyncPromise.

### Cause

PR #1805 ("Optimize Deferred execution") refactored the promise queue handling for memory optimization. 
That included the removal of several public methods, namely `runQueue(): void`, `getQueue(): \SplQueue`, and maybe others.

### Impact

The removed `runQueue()` method is public and is used by external libraries, notably https://github.com/overblog/dataloader-php.

See: https://github.com/overblog/dataloader-php/blob/master/lib/promise-adapter/src/Adapter/WebonyxGraphQLSyncPromiseAdapter.php

### Expected Behavior

  According to semantic versioning, a minor version (15.28.0 → 15.29.0) should not remove public API methods. Either:
  1. These methods should be deprecated first and removed in a major version, or
  2. The new class should be introduced alongside the old methods with deprecation notices

spawnia:
From https://github.com/webonyx/graphql-php?tab=readme-ov-file#versioning:

> Elements that belong to the public API of this package are marked with the @api PHPDoc tag. Those elements are thus guaranteed to be stable within major versions. All other elements are not part of this backwards compatibility guarantee and may change between minor or patch versions.

I am open to restoring backwards compatibility given there is a valid use case. Your recommendation from https://github.com/webonyx/graphql-php/pull/1821#issue-3746499947 is sound, can you provide a pull request that does just that? Also, let's mark `SyncPromiseQueue` with `@api` and include it in the class-reference docs.

spawnia:
As this potentially affects more users, I am implementing the fix myself and plan to release it soon.

spawnia:
Fixed with https://github.com/webonyx/graphql-php/releases/tag/v15.29.1.

PR #1821, opened 2025-12-19
Revert "Optimize `Deferred` execution (#1805)"

grossmannmartin:
### Description

This reverts commit that was merged in #1805.

### Reason for Revert

PR #1805 introduced a breaking change in a minor version by removing public API methods from SyncPromise:

- SyncPromise::runQueue() - removed
- SyncPromise::getQueue() - removed

These methods were replaced by a new SyncPromiseQueue class, but the original public methods were not preserved for backward compatibility.

### Impact

The runQueue() method is used by external libraries, notably https://github.com/overblog/dataloader-php, in their WebonyxGraphQLSyncPromiseAdapter

After upgrading to v15.29.0, any code relying on these public methods fails.

### Recommendation

The memory optimization from #1805 is valuable and should be re-introduced, but with proper backward compatibility:

1. Keep SyncPromise::runQueue() and SyncPromise::getQueue() as deprecated wrappers
2. Add deprecation notices pointing to SyncPromiseQueue
3. Remove the deprecated methods in the next major version (v16.0.0)

### Related

- Original PR: #1805
- Affected library: https://github.com/overblog/dataloader-php


spawnia:
We can discuss a resolution for the issue you are seeing in https://github.com/webonyx/graphql-php/issues/1820, I do not plan to revert though.

PR #1824, opened 2025-12-29
Do not escape forward slashes in `Printer::printDescription`

ruudk:
Similar to #1748.


spawnia (approved):

spawnia:
Thanks, released with https://github.com/webonyx/graphql-php/releases/tag/v15.29.2.

PR #1833, opened 2026-01-16
Fix/valuesofcorrecttype parentinput null 14

vardumper:
- **Utils::printSafe / printSafeJson** Special-case floats to return 'NAN', 'INF', '-INF' instead of casting. PHP 8.5 emits warnings when NaN/INF are coerced to string, which turned into visible warnings/errors and broke tests/behavior
- **Utils::AST::astFromValue()** Don’t treat floats as ints when they are NaN/INF. Prevents an incorrect cast and PHP notices like “The float 1.0E+40 is not representable as an int, cast occurred
- **SchemaValidationContext** Defensive checks around $memberType->name and filtering AST nodes (use isset(...) and use '' fallback)
- **ValuesOfCrrectType.php** In some circumstances the TypeInfo may not have a parent input type available (e.g. schema type info missing or a previously encountered validation problem). Calling getNullableType() with null resulted in a  TypeError in older versions. In those cases fall back to scalar validation and skip trave r sing the node. 

Only ValuesOfCorrectType fix was needed, the rest is PHP 8.5 improvements in order to being able to run unit tests.
This is supposed to fix an error I stumbled upon in custom Asset Types in Craft 5.8.x that extend Crafts' craftcms/cms/src/gql/interfaces/elements/Asset.php Asset class.

### DIFF
--- a/src/Validator/Rules/ValuesOfCorrectType.php
+++ b/src/Validator/Rules/ValuesOfCorrectType.php
@@ -74,11 +74,11 @@ class ValuesOfCorrectType extends ValidationRule
                 // parent input type to check if it is a list.
                 $parentInputType = $context->getParentInputType();
 
-                // In some circumstances the TypeInfo may not have a parent input
-                // type available (e.g. schema type info missing or a previously
-                // encountered validation problem). Calling getNullableType() with
-                // null resulted in a TypeError in older versions. In those cases
-                // fall back to scalar validation and skip traversing the node.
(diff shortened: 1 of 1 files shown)


spawnia:
Please read https://github.com/webonyx/graphql-php?tab=readme-ov-file#versioning and the contribution guidelines - development only happens on `master`. Also, please split into focused pull requests that each do one thing.

Issue #1027, opened 2021-12-07
Add support for adding description to schema


spawnia:
I am happy to accept a PR that adds this feature.  Since we mirror `graphql-js` when possible, the easiest route to implementing this is to follow https://github.com/graphql/graphql-js/pull/2419

acao:
just need to add `description` to `__Schema` field in introspection

spawnia:
Support was added with https://github.com/webonyx/graphql-php/releases/tag/v15.30.0.

Issue #1140, opened 2022-05-05
Add `@specifiedBy` directive


spawnia:
See https://github.com/graphql/graphql-js/pull/2276

spawnia:
Implemented in [v15.33.0](https://github.com/webonyx/graphql-php/releases/tag/v15.33.0) via https://github.com/webonyx/graphql-php/pull/1913.

PR #1834, opened 2026-01-17
Schema description

kasparsklavins:
- Implements https://github.com/graphql/graphql-spec/pull/466
- Reference implementation https://github.com/graphql/graphql-js/pull/2419
- Resolves https://github.com/webonyx/graphql-php/issues/1027

Schema description can be provided either using SDL:

```graphql
"Schema description"
schema {
  query: Query
}
```

or, by providing a `"description"` option to `\GraphQL\Type\Schema`:

```php
$schema = new \GraphQL\Type\Schema([
    'description' => 'Schema description',
    'query' => $queryType,
]);
```

### DIFF
--- a/tests/Type/SchemaConfigTest.php
+++ b/tests/Type/SchemaConfigTest.php
@@ -1,5 +1,4 @@
-<?php
-declare(strict_types=1);
+<?php declare(strict_types=1);
 
 namespace GraphQL\Tests\Type;
 


src/Type/SchemaConfig.php
```diff
@@ -91,6 +94,9 @@ public static function create(array $options = []): self
         $config = new static();
 
         if ($options !== []) {
+            if (isset($options['description'])) {
+                $config->setDescription($options['description']);
+            }
```
kasparsklavins: Codecov doesnt like missing tests for this, but there's no existing tests for `SchemaConfig::create(...)` at all... should I add one specifically for this code path?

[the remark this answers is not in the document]
spawnia: That would be great! I guess the existing tests are using the setters?

tests/Language/SchemaParserTest.php
```diff
@@ -212,6 +212,49 @@ public function testParsesTypeWithDescriptionMultiLineString(): void
         self::assertSame(6, $typeToken->line);
     }
 
+    /** @see it('parses schema with description string') */
```
spawnia: Is there such a test in graphql-js? The same question goes for all other tests with the `@see it(...` pattern - we only do that when referencing tests from there.

kasparsklavins: It was added in the reference implementation's PR https://github.com/graphql/graphql-js/pull/2419/changes#diff-eef872a01555d73bc0661ab39a979824b9f9dddbae6356144e3c16143c7acd35R144-R161

src/Type/SchemaConfig.php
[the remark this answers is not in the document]
kasparsklavins: I had misread the codecov report. The error was about getters, instead of setters. 

Anyway, added a test case for it.

spawnia:
Thank you, released with https://github.com/webonyx/graphql-php/releases/tag/v15.30.0

kasparsklavins:
Afaik, with this graphql-php should be fully compliant with October 2021 specification. Or is anything still missing?

spawnia:
@kasparsklavins Support for the `@specifiedBy` directive is still missing, see https://github.com/webonyx/graphql-php/issues/1140.

PR #1849, opened 2026-02-06
Make deprecated introspection optional for legacy servers

johnxy84:
- Add `includeDeprecated` option to `Introspection::getIntrospectionQuery()` to omit deprecated args/fields for older spec compatibility.
- Omit `isDeprecated`/`deprecationReason` and `includeDeprecated` arguments when disabled.
- Harden `BuildClientSchema` to treat missing `deprecationReason` as null.
- Add regression tests for the new behavior.


tests/Type/IntrospectionTest.php
```diff
@@ -2020,4 +2020,27 @@ public function testIncludeDescriptionFieldOnSchema(): void
         preg_match_all('/\bdescription\b/', Introspection::getIntrospectionQuery(['descriptions' => false, 'schemaDescription' => true]), $matches);
         self::assertCount(0, $matches[0]);
     }
+
+    /** @see it('excludes deprecated fields and indicators when disabled') */
```
spawnia: Are there tests in graphql-js called exactly like this? If so, please provide links to them. If not, we can omit this comment.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Changed
+
+- Allow omitting deprecated introspection arguments/fields for legacy servers
```
spawnia:
```suggestion
- Allow omitting deprecated introspection arguments/fields for legacy servers https://github.com/webonyx/graphql-php/pull/1849
```

tests/Type/IntrospectionTest.php
johnxy84: Ohh, did not know that's how it worked, but no, this is not covered [there](https://github.com/graphql/graphql-js/blob/16.x.x/src/type/__tests__/introspection-test.ts). I'll take out the comment.

PR #1857, opened 2026-02-14
Add code-first applied directives support for schema printing

spawnia:
Adds support for printing applied directives from code-first schema definitions in the SchemaPrinter.


src/Utils/SchemaPrinter.php
```diff
@@ -559,16 +604,132 @@ protected static function printInputObject(InputObjectType $type, array $options
...
+    /**
+     * @param array<string> $excludedDirectiveNames
+     *
+     * @phpstan-param AppliedDirectiveDefinition $definition
+     *
+     * @return array<DirectiveNode>
+     */
+    private static function collectAppliedDirectives(object $definition, array $excludedDirectiveNames = []): array
+    {
+        $directives = [];
```
spawnia: I don't love the pass-by-reference API of filling this, might redesign this to be functional.

spawnia:
@ruudk @simPod @LastDragon-ru First draft, what do you think? I have added some concrete discussion points. Feel free to comment and/or suggest anything else.

LastDragon-ru:
> ```
> /** @var array<GraphQL\Language\AST\DirectiveNode> */
> public array $directives;
> ```

I'm not sure about suggested `GraphQL\Language\AST\DirectiveNode` because seems the only way to create it is `Parser::directive("@directive(arg: 123)")` (+ requires args escaping...).

If I understand correctly, now we have only `\GraphQL\Type\Definition\Directive` - the definition of the directive (and [example in the documentation](https://webonyx.github.io/graphql-php/type-definitions/directives/#custom-directives)). So maybe a new type for "usage" will be better? (but we will also need a way to get the definition)

spawnia:
Responding to https://github.com/webonyx/graphql-php/pull/1857#issuecomment-3904549573 by @LastDragon-ru:

> I'm not sure about suggested GraphQL\Language\AST\DirectiveNode because seems the only way to create it is Parser::directive("@directive(arg: 123)") (+ requires args escaping...).

Direct instantiation is possible, albeit a bit cumbersome:

```php
use GraphQL\Language\AST\ArgumentNode;
use GraphQL\Language\AST\DirectiveNode;
use GraphQL\Language\AST\IntValueNode;
use GraphQL\Language\AST\NameNode;
use GraphQL\Language\AST\NodeList;

$node = new DirectiveNode([
    'name' => new NameNode([
        'value' => 'directive',
    ]),
    'arguments' => new NodeList([
        new ArgumentNode([
            'name' => new NameNode([
                'value' => 'arg',
            ]),
            'value' => new IntValueNode([
                'value' => '123', // IntValueNode stores numeric literal as string
            ]),
        ]),
    ]),
]);
```

I guess this could be simplified through a dedicated type, but that makes it harder to handle directive applications in the printer, given we want to support printing from SDL-first and code-first schemas. A simplified version might look something like this:

```php
$applied = new AppliedDirective([
    'name' => 'directive',
    'arguments' => [
        'arg' => 123,
    ],
]);
```

However, that runs into an issue of distinguishing between string and enum literals. `'arg' => 'FOO'` might print as `arg: "FOO"` or `arg: FOO`, depending on the type of the directive argument. We could add an explicit marker, or require directive type information to print this correctly.

LastDragon-ru:
> However, that runs into an issue of distinguishing between string and enum literals

Good point. Another point - `args: array[]` is not strict, so we probably will need some checks to make phpstan happy :) I just personally prefer to avoid `mixed`. Although directives args never properly validated even in graphal-js... (https://github.com/webonyx/graphql-php/discussions/1539 https://github.com/graphql/graphql-js/issues/1389) 

> I guess this could be simplified through a dedicated type, but that makes it harder to handle directive applications in the printer, given we want to support printing from SDL-first and code-first schemas. 

Yep. But the current approach also makes it harder. If we have `astNode` and `directives` what should be used?

- `astNode->directives`
- `directives`
- both

One more question - will it (`$directives`) conflict with https://github.com/graphql/graphql-js/issues/1343 ?

ruudk:
@spawnia Today I needed exactly this functionality. Would it be possible to rebase this to see if we can ship it?

PR #1917, opened 2026-05-28
chore: upgrade phpstan to 2.1.56

simPod:
Supersedes #1916 with a single squashed commit.

Includes the PHPStan 2.1.56 dependency update and the static-analysis fixes needed for the upgraded inference.

Additionally:
- Optimizes `Visitor::extractVisitFn()` to use `is_array()` instead of `is_callable()` as the type discriminator on the hot path — `is_array()` is a trivial zval type-tag check while `is_callable()` must resolve the callable
- Adds `VisitorBench` covering all four visitor API patterns to detect future performance regressions
- Adds `.phpbench` to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 vendor
 benchmarks/results
 .php-cs-fixer.cache
+.phpbench
 .phpunit.result.cache
 composer.lock
 composer.phar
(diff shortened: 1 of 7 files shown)


spawnia (approved):
Thanks @simPod for your continued effort on this. I refactored Visitor to ensure optimal performance - `is_callable` is not cheap.

PR #1927, opened 2026-06-11
Allow setting per-schema scalar overrides explicitly via SchemaConfig

tmoitie:
Fixes the silent loss of lazy type loading introduced with per-schema scalar overrides (#1869, v15.31.0), by allowing overrides to be set explicitly on `SchemaConfig` so the discovery scan of `types` can be skipped. Related to #1874.

## The problem

The executor and validator resolve built-in scalars (`Boolean`, `String`, `Int`, `Float`, `ID`) via `Schema::getType()` on essentially every operation — including a bare `{ __typename }`.

Since #1869 (and after #1884 moved scalar overrides exclusively to `types`), the first such lookup goes:

1. `getType()` calls `loadType()`, which returns `null` for built-in scalar names when a `typeLoader` is set (the `TODO` referencing #1874).
2. `getType()` falls through to `getScalarOverrides()`, which calls `materializeTypes()` — fully resolving the `types` config to scan it for `ScalarType` instances shadowing built-in names.

When `types` is a lazy callable, that scan resolves it entirely. This silently turns the documented lazy schema pattern — `typeLoader` for per-name loading plus a lazy `types` callable for types not reachable from the root types — eager on the first executed query, for every schema, including the overwhelming majority that define no scalar overrides at all. The cost lives inside the user's callable, so the library cannot be lazy *after* invoking it; the only fix is to not invoke it on the hot path.

In PHP-FPM the `Schema` is rebuilt per request, so this is a per-request cost. Downstream impact observed in a Lighthouse (`nuwave/lighthouse`) app: every GraphQL operation became 1.5–3.4× slower wall-clock between 15.30.2 and 15.32.3 (~6× more PHP work, xdebug cachegrind 9.1 MB → 57 MB for an authed `{ __typename }` request). Lighthouse's `SchemaBuilder::build()` is exactly this pattern:

```php
$config->setTypeLoader(fn ($name) => $this->typeRegistry->search($name)); // lazy
$config->setTypes(fn (): array => $this->typeRegistry->possibleTypes());  // builds ALL types from the AST when invoked
```

### Repro

```php
$config = SchemaConfig::create()
    ->setQuery($query)
    ->setTypeLoader($lazyLoader)
    ->setTypes($countingCallableBuilding500Types);

$schema = new Schema($config);
$schema->getType('Boolean'); // per-request hot path
```

| graphql-php | types materialised by one `getType('Boolean')` |
|---|---|
| v15.30.2 | 0 / 500 |
| v15.31.0 … v15.32.3 | 500 / 500 |
| this PR, default config | 500 / 500 (unchanged) |
| this PR, `scalarOverrides` set (e.g. `[]`) | **0 / 500** |

## The fix

Overrides cannot be discovered from a lazy `types` callable without resolving it — that is fundamental. So this PR makes explicitness possible instead: a new `SchemaConfig` option `scalarOverrides`.

- `null` (default): discover by scanning `types`, exactly as before — zero behaviour change for existing users.
- An array (including `[]` for "none"): use exactly these overrides, never scan. Explicit overrides take precedence in `getTypeMap()` the same way discovered ones do.

```php
$schema = new Schema([
    'query' => $queryType,
    'typeLoader' => $myTypeLoader,
    'types' => $myLazyTypes,
    'scalarOverrides' => [$uppercaseString], // or [] for none
]);
```

The setter validates entries are `ScalarType` instances named after built-in scalars.

This gives lazy-schema users a one-line opt-out while keeping the scalar-overrides feature fully intact. If this lands, Lighthouse's `SchemaBuilder::build()` could pass its known overrides (usually `[]`) explicitly and restore its pre-15.31 per-request performance — happy to file the companion PR there.

Downstream context: we currently pin 15.30.2 to avoid the regression, which re-exposes GHSA-r7cg-qjjm-xhqq, GHSA-fc86-6rv6-2jpm and the 15.31.5 quadratic-validation fix — this change is what would let us (and others in the same spot) unpin.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


spawnia:
Hey, tried to update this branch against master but ran into a GitHub quirk: deskpro/graphql-php is an organization-owned fork, and "Allow edits from maintainers" doesn't actually grant push access for org-owned forks (only personal-account ones), even though the API reports it as enabled.

I already merged master in and validated locally (fix/stan/test all green), so it's ready - could you or someone with access to the org fork pull master in and push? Happy to share the exact diff if useful.

tmoitie:
Hi @spawnia, looks like your merge made it through. I checked on my end as well and I'm getting "Already up to date"

src/Type/Schema.php
```diff
@@ -373,6 +373,13 @@ private function loadType(string $typeName): ?Type
     private function getScalarOverrides(): array
     {
         if ($this->scalarOverrides === null) {
+            // When overrides are given explicitly, the scan of types is unnecessary.
+            // This keeps a lazily provided types callable unresolved, see https://github.com/webonyx/graphql-php/issues/1874.
+            $explicitScalarOverrides = $this->config->scalarOverrides;
+            if ($explicitScalarOverrides !== null) {
+                return $this->scalarOverrides = $explicitScalarOverrides;
+            }
```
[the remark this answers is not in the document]
tmoitie: Hi @spawnia, would you like me to action this one?

Feedback from my claude was that as written, it matches existing convention. Docblock documents name-keyed contract.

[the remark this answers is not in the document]
spawnia: Yeah fine - as Claude would say "no belt and suspenders" needed for this.

src/Type/Schema.php
```diff
@@ -373,6 +373,13 @@ private function loadType(string $typeName): ?Type
     private function getScalarOverrides(): array
     {
         if ($this->scalarOverrides === null) {
+            // When overrides are given explicitly, the scan of types is unnecessary.
+            // This keeps a lazily provided types callable unresolved, see https://github.com/webonyx/graphql-php/issues/1874.
```
spawnia:
```suggestion
            // This keeps a lazily provided types callable unresolved, see https://github.com/nuwave/lighthouse/issues/2771.
```

This would be the stronger case for this, right?

src/Type/SchemaConfig.php
```diff
@@ -252,6 +271,51 @@ public function setTypes($types): self
...
+     *
+     * @throws InvariantViolation
+     *
+     * @api
+     */
+    public function setScalarOverrides(?array $scalarOverrides): self
+    {
+        if ($scalarOverrides === null) {
+            $this->scalarOverrides = null;
+
```
spawnia, line 296:
```suggestion
        if (! $scalarOverrides) {
```

Catch both `null` and `[]` - both are fine to just leave the default assignment of `null` and do no further work.

src/Type/SchemaConfig.php
```diff
@@ -252,6 +271,51 @@ public function setTypes($types): self
...
+            if (! $scalarOverride instanceof ScalarType) {
+                $scalarTypeClass = ScalarType::class;
+                $notScalarType = Utils::printSafe($scalarOverride);
+                throw new InvariantViolation("Expected instanceof {$scalarTypeClass}, got: {$notScalarType}.");
+            }
+
+            if (! Type::isBuiltInScalarName($scalarOverride->name)) {
+                $builtInScalarNames = implode(', ', Type::BUILT_IN_SCALAR_NAMES);
+                throw new InvariantViolation("Expected scalar override to be named after a built-in scalar ({$builtInScalarNames}), got: {$scalarOverride->name}.");
+            }
```
spawnia: Can we move those checks into schema validation and just do a couple `assert` statements here to satisfy PHPStan and catch basic errors at development time? That makes it more efficient at runtime. Schema validation would catch deeper errors, such as correct names.

src/Type/SchemaConfig.php
tmoitie: I don't think`null` and `[]` are interchangeable in this feature, they mean different things:

- `null` → discover overrides by scanning `types` (the legacy path, which resolves a lazy `types` callable)
- `[]` → "no overrides, and don't scan" — this is the whole point of the option, letting you keep lazy `types` unresolved

`Schema::getScalarOverrides()` branches on exactly this:

```php
$explicitScalarOverrides = $this->config->scalarOverrides;
if ($explicitScalarOverrides !== null) { // [] returns here, skips the scan
    return $this->scalarOverrides = $explicitScalarOverrides;
}
// null falls through to scanning types
```

Collapsing `[]` into `null` would force the scan back on for the empty case, defeating the lazy-loading fix and breaking `testExplicitScalarOverridesKeepLazyTypesUnresolved` (passes `'scalarOverrides' => []` and asserts the `types` callable is never called). The "empty array for none" behaviour is also documented on the property. So I'd like to keep the explicit `=== null` check here.

spawnia:
```suggestion
        if ($scalarOverrides === null) {
```

Good point. We can still shorten to this.

src/Type/SchemaConfig.php
tmoitie: Happy to do this, but one caveat worth flagging first: `validate()`/`assertValid()` is documented as "do not use in production", and `assert()` is stripped when `zend.assertions=-1`. So moving both checks out of the setter means a bad override (non-scalar, or a non-built-in name) would pass silently in production instead of failing fast at config time.

The setter runs once at config time, not on any hot path, so the runtime saving is marginal. Given that, do you still want the move? Two ways forward:

1. Keep the eager `InvariantViolation` throws in the setter as-is (fail-fast, current behaviour).
2. Convert the setter to `assert()` for the type check and add a proper check in `SchemaValidationContext` for the built-in-name rule (overrides land in the type map under the scalar name, so this needs a new explicit validation there — it isn't covered today).

src/Type/SchemaConfig.php
tmoitie: Is that true? It effectively locks the consumer of the code out from being able to set $this->scalarOverrides to null after they have set it to something. Happy to do this, but for completeness I feel as though it should still be settable back to null.

spawnia: ah, agreed.

src/Type/SchemaConfig.php
spawnia: Full schema validation is meant to run pre-production and do the expensive checks there.

spawnia (approved):

src/Type/SchemaConfig.php
```diff
@@ -252,6 +271,47 @@ public function setTypes($types): self
...
+        foreach ($scalarOverrides as $scalarOverride) {
+            // Deeper validation (correct scalar names, actually being a ScalarType) happens
+            // during schema validation, see SchemaValidationContext::validateScalarOverrides().
+            // This assertion just catches basic misuse at development time and narrows the type
+            // for the name-keyed map below.
+            // @phpstan-ignore-next-line not strictly enforceable unless PHP gets generics
+            assert($scalarOverride instanceof ScalarType, 'Expected instanceof ' . ScalarType::class . ', got: ' . Utils::printSafe($scalarOverride));
+
+            $this->scalarOverrides[$scalarOverride->name] = $scalarOverride;
+        }
```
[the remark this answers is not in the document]
spawnia: I think the comment explains it fine, dismissed.

spawnia:
Thanks @tmoitie! Released as [v15.35.0](https://github.com/webonyx/graphql-php/releases/tag/v15.35.0).

PR #1945, opened 2026-07-09
Fix client schema deprecated input values

simPod:
## Summary

Preserve `deprecationReason` when rebuilding input values from introspection in `BuildClientSchema`.

This keeps deprecated field arguments and input object fields available when a client schema is printed back to SDL.


spawnia (approved):

spawnia:
Thanks @simPod! Released as [v15.34.1](https://github.com/webonyx/graphql-php/releases/tag/v15.34.1).

PR #1948, opened 2026-07-16
Optimize Lexer::readString

OpaqueRock:
## Summary

Bulk-scan runs of ordinary bytes with `strcspn()` instead of decoding/validating one UTF-8 character at a time via `readChar()`. Falls back to per-character logic only for quotes, line terminators, control characters, and escape sequences. TAB (0x09) is excluded from the stop-byte set since it is a legal, unescaped SourceCharacter.

## Why

`readString()` is on the hot path for every request in a GraphQL server. The per-call overhead of character-by-character decoding compounds across large queries with many string literals.

## Benchmarks

Measured with an isolated benchmark harness (PHP 8.3.6, x86_64, Xdebug disabled, 500 iterations averaged over 3 runs):

| Metric | master | This PR | Speedup |
|--------|--------|---------|---------|
| `Parser::parse` (Lexer) | 10,907 μs/op | 4,026 μs/op | 2.71x |

## Testing

- Full test suite passes (1990+ tests)
- PHPStan passes
- New tests cover all single-char escape sequences, multi-byte UTF-8 in `readChar()`, and the `&` punctuator — achieving 100% line coverage on Lexer.php


src/Language/Visitor.php
```diff
@@ -416,17 +434,13 @@ public static function visitInParallel(array $visitors): array
...
-            'leave' => static function (Node $node) use ($visitors, $skipping, $visitorsCount) {
+            'leave' => static function (Node $node, $key, $parent, $path, $ancestors) use ($skipping, $visitorsCount, $leaveFns) {
                 for ($i = 0; $i < $visitorsCount; ++$i) {
                     if ($skipping[$i] === null) {
-                        $fn = self::extractVisitFn(
-                            $visitors[$i],
-                            $node->kind,
-                            true
-                        );
+                        $fn = $leaveFns[$i][$node->kind] ?? null;
```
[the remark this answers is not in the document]
OpaqueRock: Replicated and fixed the issue in 5605b5b1. New tests added to cover the custom kinds.

src/Language/Visitor.php
```diff
@@ -383,24 +383,42 @@ public static function visitInParallel(array $visitors): array
...
+        static $allKinds;
+        $allKinds ??= array_values(array_filter((new \ReflectionClass(NodeKind::class))->getConstants(), 'is_string'));
+
+        $enterFns = [];
+        $leaveFns = [];
+        foreach ($visitors as $i => $visitor) {
+            foreach ($allKinds as $kind) {
+                $enterFns[$i][$kind] = self::extractVisitFn($visitor, $kind, false);
+                $leaveFns[$i][$kind] = self::extractVisitFn($visitor, $kind, true);
+            }
```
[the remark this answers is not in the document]
OpaqueRock: Replicated and fixed the issue in 5605b5b1. New tests added to cover the custom kinds.

tests/Language/VisitorTest.php
```diff
@@ -1905,4 +1905,77 @@ public function testThrowsExceptionWhenAddingIntegerToNodeList(): void
...
+        self::assertSame(
+            [
+                ['enter', 'CustomKind'],
+                ['leave', 'CustomKind'],
+            ],
+            $visited
+        );
+    }
+
+    /** Same as above, but for a visitor keyed specifically by the custom kind. */
```
spawnia:
```suggestion
    /** Same as testVisitInParallelCallsGenericVisitorsForCustomKind, but for a visitor keyed specifically by the custom kind. */
```

tests/Language/LexerTest.php
```diff
@@ -296,6 +296,21 @@ public function testLexesStrings(): void
         );
     }
 
+    public function testLexesStringsWithUnescapedTabCharacter(): void
+    {
+        $source = '"before' . "\t" . 'after"';
```
spawnia: more readable with HEREDOC?

src/Language/Visitor.php
```diff
@@ -383,24 +385,42 @@ public static function visitInParallel(array $visitors): array
...
+        // Enter and leave are cached independently in their own dispatcher.
+        $enterFns = [];
+        $leaveFns = [];
+
         return [
-            'enter' => static function (Node $node) use ($visitors, $skipping, $visitorsCount) {
+            // Declares node, key, parent, path, and ancestors explicitly to
+            // match Visitor::visit()'s `$visitFn($node, $key, $parent,
+            // $path, $ancestors)` call.
+            'enter' => static function (Node $node, $key, $parent, $path, $ancestors) use ($visitors, $skipping, $visitorsCount, &$enterFns) {
```
spawnia: Types?

src/Language/Visitor.php
```diff
@@ -383,24 +385,42 @@ public static function visitInParallel(array $visitors): array
...
+        // (visitor, kind, isLeaving) triple never changes during one
+        // traversal. Cache it per visitor/kind on first encounter instead,
+        // using `false` as a "no callback" sentinel so a plain `??` lookup
+        // can tell "not cached yet" (missing key -> null) apart from
+        // "cached, nothing to call" (`false`). A Node's kind isn't required
+        // to be a NodeKind constant (custom Node subclasses are supported),
+        // so this works for arbitrary kinds, not just the built-in ones.
+        // Enter and leave are cached independently in their own dispatcher.
+        $enterFns = [];
+        $leaveFns = [];
```
spawnia: Add `@var` with types - what specifically is in those arrays?

src/Language/Lexer.php
```diff
@@ -391,18 +391,50 @@ private function readString(int $line, int $col, Token $prev): Token
...
+        // character at a time.
+        //
+        // The stop-byte set below includes ASCII control characters
+        // (0x00-0x1F), not just backslash/quote/line-terminators, because
+        // assertValidStringCharacterCode() must still reject most of them.
+        // TAB (0x09) is excluded: it is a legal, unescaped SourceCharacter
+        // per the GraphQL spec, so it is scanned through like any other
+        // ordinary character instead of being individually inspected.
+        static $stopBytes;
+        $stopBytes ??= '"\\' . implode('', array_map('chr', array_diff(range(0x00, 0x1F), [0x09])));
```
spawnia: Write this out as a const?

spawnia (commented):
Are those two independent changes? Please split if so.

OpaqueRock:
Thanks for the review! I've pushed two updates:

1. **Addressed the requested changes**: converted the tab-character Lexer test to a HEREDOC, added strict param typing (`array` for `$path`/`$ancestors`) plus `@phpstan-param` docblocks for the `enter`/`leave` closures in `visitInParallel()` (union types aren't available since this targets PHP 7.4), added `@var` docblocks for the `$enterFns`/`$leaveFns` caches, and replaced the runtime-computed `$stopBytes` string with a `STRING_STOP_BYTES` class constant.

2. **Split out the unrelated `visit()` fix**: per your suggestion, the `visit()` undefined-array-key fix for custom Node kinds (and its two regression tests) is no longer part of this PR — it's now a standalone PR: #1952.

This PR is now scoped solely to the `Lexer::readString()` / `Visitor::visitInParallel()` optimizations. Let me know if anything else needs adjusting.

spawnia:
I would like the Lexer and Visitor optimizations as separate PRs.

OpaqueRock:
The `Visitor::visitInParallel()` optimization has been now split out into its own PR, #1953

src/Language/Lexer.php
```diff
@@ -416,79 +453,77 @@ private function readString(int $line, int $col, Token $prev): Token
...
-                $chunk .= $char;
+                    }
+
+                    $this->assertValidStringCharacterCode($code, $position - 2);
+
+                    $value .= Utils::chr($code);
+                    break;
+                    // null means EOF, will delegate to general handling of unterminated strings
+                case null:
+                    break;
```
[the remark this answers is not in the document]
spawnia: Let's add a failing test case first to make the fix durable.

OpaqueRock:
I looked into the null value comparison and it's a real, reachable bug, though it pre-dates this PR (the `switch`/`case null` block existed identically on `master`; this PR just moved/reformatted it).

**The problem:** `switch` in PHP uses loose (`==`) comparison for each `case`. So `case null:` also matches integer `0`, not just true EOF. In `readString()`, after consuming a backslash, `$code` is `0` for a literal NUL byte (0x00) in the input, and `null` only on genuine EOF. Because of the loose comparison, a literal NUL byte after a backslash was being misidentified as EOF.

**Impact:** For input containing `\` followed by a raw NUL byte inside a string literal, the lexer silently dropped both bytes and kept parsing, instead of raising a syntax error. E.g. `"ab\<NUL>cd"` produced the token value `'abcd'` with no error at all.

Per the GraphQL spec, a NUL byte is not a valid raw string character and has no dedicated escape (`\0` isn't in the `EscapedCharacter` grammar) — the only spec-compliant way to include it is `\u0000`. So `\<NUL>` should always be rejected as an invalid escape sequence, the same as `\q` or any other unrecognized escape.

**Fix:** Added an explicit `if ($code === null) { break; }` strict-comparison check immediately before the `switch`, and removed the now-dead `case null:` branch inside the switch. Considered using PHP 8's `match` expression instead, but the project's floor is PHP 7.4, so that's not available; the strict pre-check is also simpler here since the `\u` case has multi-statement logic that doesn't map cleanly to `match` anyway.

Added a regression test (`reportsUsefulStringErrors` in `LexerTest.php`) covering `\` followed by a raw NUL byte, verified it failed before the fix and passes after.

tests/Language/LexerTest.php
```diff
@@ -296,6 +296,23 @@ public function testLexesStrings(): void
...
+        self::assertArraySubset(
+            [
+                'kind' => Token::STRING,
+                'start' => 0,
+                'end' => strlen($source),
+                'value' => substr($source, 1, -1), // strip the surrounding quotes
+            ],
+            (array) $this->lexOne($source)
+        );
+    }
```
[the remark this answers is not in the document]
spawnia: This is incorrect. Since PHP 7.3, flexible heredoc/nowdoc with an indented closing marker does not append a trailing newline. The string ends at the last content character. Verified: `$source` is exactly 14 bytes (`"before\tafter"`) with no trailing newline.

src/Language/Lexer.php
```diff
@@ -391,18 +400,46 @@ private function readString(int $line, int $col, Token $prev): Token
...
+
+            if ($runLength > 0) {
+                $chunk = substr($body, $byteStart, $runLength);
                 $value .= $chunk;
+                // $this->position counts decoded characters, not bytes: count the
+                // non-continuation bytes (i.e. NOT matching 10xxxxxx) in the chunk to
+                // keep it accurate for multi-byte UTF-8 content.
+                $continuationBytes = preg_match_all('/[\x80-\xBF]/', $chunk);
+                assert(is_int($continuationBytes), 'Pattern is valid, so preg_match_all() cannot fail');
+                $this->position += strlen($chunk) - $continuationBytes;
```
spawnia: The three-line `preg_match_all` + `assert` + arithmetic block can be replaced with a single `mb_strlen` call that returns the codepoint count directly. For valid UTF-8, `mb_strlen($chunk, 'UTF-8')` is semantically equivalent to `strlen($chunk) - count(continuation bytes)`, avoids PCRE overhead, and makes the intent ("count codepoints") explicit.

```suggestion
                $this->position += mb_strlen($chunk, 'UTF-8');
```

spawnia (commented):
Comment.

The PR is a sound performance optimization with no correctness bugs. The three surviving comments are minor improvements: a simplification opportunity (`mb_strlen` over `preg_match_all`), an open reviewer request for named constants, and a test hygiene suggestion. None rise to the level of requesting changes.

src/Language/Lexer.php
```diff
@@ -416,79 +453,79 @@ private function readString(int $line, int $col, Token $prev): Token
                 );
             }
 
-            $this->assertValidStringCharacterCode($code, $this->position);
-            $this->moveStringCursor(1, $bytes);
-
-            if ($code === 92) { // \
-                $value .= $chunk;
-                [, $code] = $this->readChar(true);
+            if ($code === 10 || $code === 13) { // LineTerminator
```
spawnia, line 453: Open question: could extracting `10` (LF) and `13` (CR) into named constants have a performance cost in such a hot path? Either way, whichever approach we go with should be consistent across the file — right now it's magic numbers everywhere.

tests/Language/LexerTest.php
```diff
@@ -296,6 +296,23 @@ public function testLexesStrings(): void
...
+        $source = <<<'GRAPHQL'
+        "before	after"
+        GRAPHQL;
+
+        self::assertArraySubset(
+            [
+                'kind' => Token::STRING,
+                'start' => 0,
+                'end' => strlen($source),
+                'value' => substr($source, 1, -1), // strip the surrounding quotes
```
spawnia: The expected `end` and `value` are derived from `$source` at assertion time (`strlen`, `substr`). If the test input were wrong, the assertion would be wrong in the same direction and the test would still pass. Since the source is a fixed 14-byte literal (`"before<TAB>after"`), both values can be hardcoded.

```suggestion
                'end' => 14,
                'value' => "before\tafter",
```

spawnia:
Reviewed at 031cb52 (standard, full)

🤖 Posted by Claude Code

src/Language/Lexer.php
OpaqueRock: It seems based on quick testing on my machine that `mb_strlen` performs much better (2-8x speed) on short strings and strings with Unicode characters. It falls behind (1.3x execution time) on very long ASCII strings like base64 encoded files. I will switch to `mb_strlen` based on these findings. It is not that much slower on long ASCII.

src/Language/Lexer.php
OpaqueRock: I would expect that because these are just number literals, there is no measurable performance difference compared to switching them to constants. If anything, constants might be slower than inlined values in some cases, as they have to be resolved at some point, where as inlined value is already resolved. My rule of thumb when optimizing is that usually "the messier it looks, the faster it runs". Then it becomes a balancing act of keeping code maintainable while trying to avoid as much of the overhead of a generic solution as possible.

OpaqueRock:
Replaced `preg_match_all` codepoint-counting with `mb_strlen` and hardcoded the tab-character test's expected string/length.

src/Language/Lexer.php
spawnia: fine, thanks!

src/Language/Lexer.php
spawnia: nice!

spawnia (approved):

spawnia:
Thanks @OpaqueRock! Released as [v15.36.0](https://github.com/webonyx/graphql-php/releases/tag/v15.36.0).

PR #1954, opened 2026-07-24
perf(amp): avoid fibers in future adapter

simPod:
## Context
`AmpFutureAdapter` created an AMPHP Fiber for every promise continuation and aggregate. Wide GraphQL result graphs therefore retained many Fiber states and event-loop callbacks even when the underlying Futures only represented batched work.

## Decision
Compose and aggregate Futures through `map()` and `catch()` callbacks, settling one `DeferredFuture` while retaining Promise fulfillment, rejection, and nested-Future flattening semantics.

## Consequences
- Reduces per-field allocation and event-loop scheduling for wide asynchronous result graphs.
- Preserves deferred completion for genuinely asynchronous resolvers.
- Adds coverage for nested Future fulfillment/recovery, fulfillment errors, aggregate key preservation, aggregate rejection, and resolution after rejection.

## Benchmark

On the same wide GraphQL result graph with 838 parent objects and three Future-backed fields:

| Adapter | CPU time | Peak memory |
|---|---:|---:|
| Synchronous adapter | 1.48s | 122MB |
| Existing AMPHP adapter | 2.11s | 274MB |
| This change | 1.67-1.70s | 149MB |

Relative to the existing AMPHP adapter, this reduces CPU time by about 19% and peak memory by about 46%. The remaining overhead compared with the synchronous adapter is about 13-15% CPU and 22% memory.

## Disclosure

This change was identified while migrating a production GraphQL application that is not yet fully migrated to asynchronous resolving. It is a step toward that migration, rather than a claim of end-to-end asynchronous production use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
+- Avoid creating an AMPHP Fiber for each `AmpFutureAdapter` continuation and aggregate https://github.com/webonyx/graphql-php/pull/1954
+
 ## v15.35.0
 
 ### Added
--- a/src/Executor/Promise/Adapter/AmpFutureAdapter.php
+++ b/src/Executor/Promise/Adapter/AmpFutureAdapter.php
@@ -8,9 +8,6 @@ use GraphQL\Error\InvariantViolation;
 use GraphQL\Executor\Promise\Promise;
 use GraphQL\Executor\Promise\PromiseAdapter;
 
-use function Amp\async;
-use function Amp\Future\await;
-
 /**
  * Allows integration with amphp/amp v3 (fiber-based futures).
  *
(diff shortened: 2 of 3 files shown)


src/Executor/Promise/Adapter/AmpFutureAdapter.php
```diff
@@ -46,25 +43,33 @@ public function then(Promise $promise, ?callable $onFulfilled = null, ?callable
...
-            } catch (\Throwable $reason) {
-                if ($onRejected === null) {
-                    throw $reason;
+        $deferred = new DeferredFuture();
+
+        static::observeFuture(
+            $future,
+            static function ($value) use ($deferred, $onFulfilled): void {
+                try {
+                    static::resolveDeferred($deferred, $onFulfilled === null ? $value : $onFulfilled($value));
```
spawnia:
```suggestion
                    static::resolveDeferred(
                        $deferred,
                        $onFulfilled === null 
                            ? $value
                            : $onFulfilled($value)
                    );
```

src/Executor/Promise/Adapter/AmpFutureAdapter.php
```diff
@@ -127,30 +132,48 @@ public function all(iterable $promisesOrValues): Promise
             ? $promisesOrValues
             : iterator_to_array($promisesOrValues);
 
-        /** @var array<Future<mixed>> $futures */
-        $futures = [];
+        $deferred = new DeferredFuture();
+        $resultFuture = $deferred->getFuture();
```
spawnia:
```suggestion
        $deferredFuture = new DeferredFuture();
        $future = $deferred->getFuture();
```

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ You can find and compare releases at the [GitHub release page](https://github.co
 
 ## Unreleased
 
+### Fixed
+
+- Avoid creating an AMPHP Fiber for each `AmpFutureAdapter` continuation and aggregate.
```
spawnia:
```suggestion
- Avoid creating an AMPHP Fiber for each `AmpFutureAdapter` continuation and aggregate https://github.com/webonyx/graphql-php/pull/1954
```

spawnia (approved):

spawnia:
Thanks @simPod! Released as [v15.36.1](https://github.com/webonyx/graphql-php/releases/tag/v15.36.1).

